PR #451, opened 2014-12-27
Add a MONGOOSE_NO_URL_DECODE flag

sebgod:
For certain applications, it is essential that
http://host.name:80/path/with/inline%2fslash/
and
http://host.name:80/path/with/inline/slash/
are different.
Currently, mongoose always decodes the URL from the HTTP header,
thus irreversibly loosing that distinction.
This flag gives the user a chance to change this behaviour.

mongoose.c:
    Do not decode the request uri path if MONGOOSE_NO_URL_DECODE
    is defined.

docs/Embed.md:
    Document the MONGOOSE_NO_URL_DECODE flag.

### DIFF
--- a/docs/Embed.md
+++ b/docs/Embed.md
@@ -146,6 +146,7 @@ all Net Skeleton functions will be available too.
     -DMONGOOSE_NO_CGI           Disable CGI support
     -DMONGOOSE_NO_DAV           Disable WebDAV support
                                 (PUT, DELETE, MKCOL, PROPFIND methods)
+    -DMONGOOSE_NO_URL_DECODE    Disables decoding of URL pathes
     -DMONGOOSE_NO_DIRECTORY_LISTING  Disable directory listing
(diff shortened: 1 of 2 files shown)


sebgod:
Hello, first sorry for the delay, had no internet access the last days.
I think I had a misconception about the Amazon AWS API, they only use embedded URIs in the query part, not in the URI path part.
A valid example of embedded URI's is path part is the [OCSP protocol over HTTP](http://tools.ietf.org/html/rfc2560#appendix-A)
```
An OCSP request using the GET method is constructed as follows:

GET {url}/{url-encoding of base-64 encoding of the DER encoding of
the OCSPRequest}
```
My basic use case is that I am building a wrapper for the mongoose library
for a functional/logic language called Mercury. For technical reasons,
it is much simpler to use my "split by / and decode all segments"
approach to provide users access the the
URI path. 
For I don't know what users of my API require, it would be really helpful
to allow for this of course optional flag as in my diff. Since for logic languages,
using split and lists is the most feasible way to implement matching of URI's and pathes etc.

PR #454, opened 2015-01-06
Enable building of example/Makefile on Win32

sebgod:
* Adding example/.gitignore to ignore *.exe
* Introduced variable CLIBS, which on Windows is initialised with -lws2_32
* using this variable in all example Makefiles


sebgod:
I did introduce that variable since I wanted to ensure that the library comes at the end of the command line (most linkers insist on that). I didn't see the CFLAGS_EXTRA which does essentially that.
So closing this pull request in favour of #456
Thanks for that comment.

PR #459, opened 2015-01-16
Update Makefile

joshcriz:
Fixing typo in 'clean' section


joshcriz:
Actually, it looks like this typo is in multiple Makefiles in the "examples" folder.

PR #463, opened 2015-01-27
Replace obsolete gethostbyname by getaddrinfo

CurlyMoo:
This also protect against the latest glibc bug

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -593,28 +593,28 @@ int ns_socketpair(sock_t sp[2]) {
 
 // TODO(lsm): use non-blocking resolver
 static int ns_resolve2(const char *host, struct in_addr *ina) {
-	int rv = 0;
-	struct addrinfo hints, *servinfo, *p;
-	struct sockaddr_in *h = NULL;
-	char *ip = malloc(17);
-	memset(ip, '\0', 17);
+  int rv = 0;
+  struct addrinfo hints, *servinfo, *p;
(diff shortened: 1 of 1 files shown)


cpq:
Thank you for the change!

`getaddrinfo` is not as portable as gethostbyname, for example MSVC6 doesn't have it.
Other OSes (especially embedded ones) might not have getaddrinfo as well.
Since this change doesn't provide a visible benefit over gethostbyname, and looses in portability, I tend to not to merge it.

CurlyMoo:
I think the benefits are clear:
1. gethostbyname isn't thread safe.
2. gethostbyname isn't reentrant.
3. gethostbyname doesn't support ipv6.
4. gethostbyname is considered obsolete because all of the above.

So instead of just not including it you should opt for the alternative and that is to check if getaddrinfo is available and then use it, if not fallback to gethostbyname. For example, that's how PolarSSL ensures portability.

cpq:
Makes sense. Please wrap your change into the `#ifdef NS_ENABLE_GETADDRINFO` conditional.

CurlyMoo:
Done

cpq:
Thank you. Please confirm that you're fine to release your change under the current mongoose license - then I'll merge it.

CurlyMoo:
As long as it `also` stays Open Source GPLv2 i'm fine with it.

cpq:
Merging, thank you.

PR #465, opened 2015-01-29
Allow for compilation in Visual Studio with NS_ENABLE_IPV6.

doug16rogers:
This adds #include \<Ws2tcpip.h\>, which is required for IPv6 support in VS9.
Also, VS9 does not allow c99 in-block declarations, so there's a change to move an assignment to be after some const declarations.

Tested with VS 2008 x86 and VS 2013 x64 (and x86) by running:

    mhttpd.exe -listening_port "[::1]:8080"

Tested with IE11 and Chrome 40.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -79,6 +79,7 @@
 #ifdef _MSC_VER
 #pragma comment(lib, "ws2_32.lib")    // Linking with winsock library
 #endif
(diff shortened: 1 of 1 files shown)


cpq:
Added winsock2.h include as well, to allow build on MSVC6.
Thanks!

PR #472, opened 2015-02-16
Fix compile on MINGW

mmicko:
Tested on Linux as well.


cpq:
Pushed https://github.com/cesanta/mongoose/commit/53a55da7bc13d715d1072b12f464393d2960fd13
Thanks!

PR #474, opened 2015-02-24
OS/2 patches by KO Myung-Hun

mmicko:
Upstream patch from MAME project


cpq:
Thanks Miodrag.
Do you agree to release this patch under the existing mongoose copyright?
If yes, I'll integrate it.

mmicko:
Sure, this is external contribution but it is fine, since all contributions
to MAME are considered public.

Regards,
Miodrag

cpq:
Thanks. Merging.
I am surprised to see that os2 is still out there.

PR #485, opened 2015-03-12
Fix remove_double_dots_and_double_slashes removing all the dots leading http server to serve wrong URI

dsrosario:
If the uri is something like '/js/...jquery.js', remove_double_dots_and_double_slashes will remove the 3 dots and the http server will serves the /js/jquery.js file.
remove_double_dots_and_double_slashes should check if a dot or double dots is followed by a slash (or backslash) and only remove this to avoid disclosure attack.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/485)

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -2428,7 +2428,8 @@ static void remove_double_dots_and_double_slashes(char *s) {
       // Skip all following slashes, backslashes and double-dots
       while (s[0] != '\0') {
         if (s[0] == '/' || s[0] == '\\') { s++; }
-        else if (s[0] == '.' && s[1] == '.') { s += 2; }
+        else if (s[0] == '.' && (s[1] == '/' || s[1] == '\\')) { s += 2; }
+        else if (s[0] == '.' && s[1] == '.' && (s[2] == '/' || s[2] == '\\')) { s += 3; }
         else { break; }
       }
     }
--- a/mongoose.c
+++ b/mongoose.c
@@ -2429,6 +2429,7 @@ static void remove_double_dots_and_double_slashes(char *s) {
       while (s[0] != '\0') {
         if (s[0] == '/' || s[0] == '\\') { s++; }
         else if (s[0] == '.' && (s[1] == '/' || s[1] == '\\')) { s += 2; }
+		else if (s[0] == '.' && s[1] == '.' && s[2] == '\0') { s += 2; }
         else if (s[0] == '.' && s[1] == '.' && (s[2] == '/' || s[2] == '\\')) { s += 3; }
         else { break; }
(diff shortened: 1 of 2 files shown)


cpq:
Hi Marko,
could you review this PR for the potential disclosure attack please?

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/mongoose/485)

Reviewed files:
- mongoose.c @ r1





---

<sup>**[mongoose.c, line 2432 \[r1\]](https://reviewable.io:443/reviews/cesanta/mongoose/485#-Jkbp8UW0lSEhgf2dPgr)** ([raw file](https://github.com/cesanta/mongoose/blob/3cd20a852792843d6b24878649a056ca8cdc6c9d/mongoose.c#L2432)):</sup>
Yep the old code is bad. However consider this case:

```c
char str[] = "foo/x..y/././y/../../..";
remove_double_dots_and_double_slashes(str);
printf("%s", str);
```

yields: `foo/x..y/././y/`

While with your patch it yields: `foo/x..y/y/..`.

I guess that what you want to do is to check if the second dot is followed by either a `'/'`, a `'\\'` or a a `'\0'`.

I like the removal of the spurious `.` components although AFAIK it has no impact on security and file lookup semantics, but it helps at canonicalizing the paths, which might be useful for log aggregation or whatnot.

Would you also like to provide a unit test so it's clear which patterns have been thought out?

---

dsrosario:
Hi,

I agreed that we should check the '..' is ended with \0 or not. But having a function that removes spurious '.', I don't think it is a good idea. That approach leads the webserver serving wrong or unexisting url/files.
The problem I'm fixing is the webserver serves '/js/....jquery.js' (with the content of jquery.js) when in reality it should denied the request at the first place.
Another problem that I can imagine is with this function, mongoose will never match any files that starts with a '.' (I don't known if this was intended or not, but if the user doesn't want that file to be served at the first place, it should remove it from the "virtual directory" or define url with a explicit handler to denied it).

dsrosario:
I have updated the unit_test to pass .... :\ and added 2 more tests: foo/x..y/././y/../../.. that should be translated to foo/x..y/y/ and foo/..x that should be unchanged.

mmikulicic:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/mongoose/485)

Reviewed files:
- mongoose.c @ r3
- test/unit_test.c @ r2

---

Thanks, LGTM. I'm going to squash the commits and merge

---

mmikulicic:
Merged manually

PR #486, opened 2015-03-13
Compile fix

mmicko:


mmicko:
Added fix for Visual Studio builds as well

cpq:
Firt commit looks good to me,
second commit defines ssize_t - but mongoose doesn't use it, could you elaborate please?

PR #498, opened 2015-03-28
an important file (to use mongoose as a submodule)

dynnamitt:
makes it painless to build inside git submodule

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/498)


cpq:
Thanks - could you elaborate why does it help?
I hesitate to litter a repo with the apparently useless housekeeping things

dynnamitt:
Ok 1st of all; i love this code. Without it I wouldn't see any real use for 
C anymore. It gives me great joy now to hack pointers again

I dont understand what you mean by litter in case of gitignore

It is a must in every git codebase. http://git-scm.com/docs/gitignore

You have Makefiles in example dirs that do produce litter. (From a git 
standpoint this is) Objectfiles must therefore be ignored.

At root; no makefile but surely programmers (users of mongoose as 
subdir/submodule) must build a mongoose object file here from our own Makefile

mongoose/mongoose.o is buildt whenever it is added as an dependecy

Sample here https://github.com/dynnamitt/awaitor/blob/master/Makefile

Git status will always report changes in  submodule without all objectfiles 
listed in .gitignore file

Its just anoyance that makes me fork a codebase immediately

Submodules in git makes the embedding/syncing of 3rd party C code a-lot 
better. The (old) practice of taking a copy/snapshot of a project and 
checking the whole into other projects is wasteful data dup and doesnt give 
powerful hyperlinks.
Submodule use is superior but it demands that we all agree on some small 
conventions maybe

Regards
D

Sent with AquaMail for Android
http://www.aqua-mail.com

cpq:
Thanks for the elaborate reply!
We'd like to keep our repos free from housekeeping things like .gitignore files, so I won't be merging this. Thanks for the PR though!

PR #503, opened 2015-04-02
Adding a check to avoid undefined behavior if socket is out of the bound...

himanshupsharma:
Adding a check to avoid undefined behavior if socket is out of the bounds of FD_SETSIZE.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/503)

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -1060,7 +1060,7 @@ static void ns_handle_udp(struct ns_connection *ls) {
 }
 
 static void ns_add_to_set(sock_t sock, fd_set *set, sock_t *max_fd) {
-  if (sock != INVALID_SOCKET) {
+  if ( (sock != INVALID_SOCKET) && (sock >= 0) && (sock < FD_SETSIZE) ) {
     FD_SET(sock, set);
     if (*max_fd == INVALID_SOCKET || sock > *max_fd) {
       *max_fd = sock;
@@ -1971,6 +1971,8 @@ struct threadparam {
 };
 
 static int wait_until_ready(sock_t sock, int for_read) {
+  if ( (sock < 0) || (sock >= FD_SETSIZE) )
+    return 0;
   fd_set set;
   FD_ZERO(&set);
   FD_SET(sock, &set);


himanshupsharma:
Done. Thanks.

mongoose.c
```diff
@@ -1060,7 +1060,7 @@ static void ns_handle_udp(struct ns_connection *ls) {
 }
 
 static void ns_add_to_set(sock_t sock, fd_set *set, sock_t *max_fd) {
-  if (sock != INVALID_SOCKET) {
+  if ( (sock != INVALID_SOCKET) && (sock >= 0) && (sock < FD_SETSIZE) ) {
```
pavel-pimenov: Visual C++ V547 Expression 'sock >= 0' is always true. Unsigned type value is always >= 0

mongoose.c
```diff
@@ -1971,6 +1971,8 @@ struct threadparam {
 };
 
 static int wait_until_ready(sock_t sock, int for_read) {
+  if ( (sock < 0) || (sock >= FD_SETSIZE) )
```
pavel-pimenov: 1. error build *.c file 
mongoose\mongoose.c(1989): error C2275: 'fd_set' : illegal use of this type as an expression

fix
static int wait_until_ready(sock_t sock, int for_read) {
  fd_set set;
  if ( (sock < 0) || (sock >= FD_SETSIZE) )
    return 0;
....

pavel-pimenov: Windows - V547 Expression 'sock < 0' is always false. Unsigned type value is never < 0

himanshupsharma:
Oh, so sock_t is unsigned int on Windows? On Posix, socket() might return -1 for error cases.
Then, "INVALID_SOCKET=-1" is going to wrap around on Windows. Is that intentional?

pavel-pimenov:
http://www.viva64.com/en/d/0137/print/

Another example:
....
SOCKET csd;
csd = accept(nsd, (struct sockaddr *) &sa_client, &clen);
if (csd < 0)
  ....
The accept function in Visual Studio header files returns a value that has the unsigned SOCKET type. That's why the check 'csd < 0' is invalid since its result is always false. The returned values must be explicitly compared to different constants, for instance, SOCKET_ERROR:

if (csd == SOCKET_ERROR)
......

cpq:
yeah on windows, it is

    #define INVALID_SOCKET  (SOCKET)(~0)
    #define SOCKET_ERROR            (-1)

So it INVALID_SOCKET is essentially -1, but I agree with Pavel, the check needs to be changed somehow.

One way to do it is to disable it on windows:

    if ( (sock != INVALID_SOCKET) &&
    #ifndef _WIN32
     (sock >= 0) &&
    #endif
     (sock < FD_SETSIZE) )

mmikulicic:
Beware nitpick below:

TL;DR: comparing to literal -1 works and is defined, unless when on a platform which has sizeof(socket_t) < sizeof(int) (which I don't know if it even exist). Checking using relational operators is always wrong.


AFAIK `INVALID_SOCKET` is the sentinel value for the socket type (`~0`). `SOCKET_ERROR` is defined as -1 in `winsock.h`, and should be used with calls that an error status, e.g. `connect`.

The example on http://www.viva64.com/en/d/0137/print/ works because according to the C99 standard ("6.3.1.3 Signed and unsigned integers") during a comparison between a signed int (`-1`) and an unsigned int (the socket), the signed integer is converted to the an unsigned value by adding `UINT_MAX + 1`. It means that the expression is equivalent to UINT_MAX. An invalid socket will be returned as `~0`, which is also UINT_MAX.

This works because the `-1` literal has the same rank (or smaller) than SOCKET (which is either `unsigned int` on w32 or `unsigned __int64` on w64).

For example this expression is always false on w32: `scd == -1L`, while it would work on w64.


---

References:
- https://msdn.microsoft.com/en-us/library/windows/desktop/ms740516%28v=vs.85%29.aspx
- https://msdn.microsoft.com/en-us/library/windows/desktop/ms737625%28v=vs.85%29.aspx

Grepping through MSVC6 include files:
```
typedef u_int           SOCKET;
#define INVALID_SOCKET  (SOCKET)(~0)
#define SOCKET_ERROR            (-1)
SOCKET PASCAL FAR accept (SOCKET s, struct sockaddr FAR *addr,
int PASCAL FAR connect (SOCKET s, const struct sockaddr FAR *name, int namelen);
```

pavel-pimenov:
Windows Sockets handles have no restrictions, other than that the value INVALID_SOCKET is not a valid socket. Socket handles may take any value in the range 0 to INVALID_SOCKET–1.

"sock < FD_SETSIZE"  -  error if _WIN32  (#define FD_SETSIZE      64)

if ( sock != INVALID_SOCKET 
#ifndef _WIN32
 && sock >= 0 && sock < FD_SETSIZE 
#endif
) 

?

mmikulicic:
@cpq:
> I believe a socket could be either -1, or >= 0.

A socket value on windows is always >= 0, even if it "is" -1.

@pavel-pimenov:
I'm not sure I understand the problem. 
`sock != INVALID_SOCKET && sock < FD_SETSIZE ` should be valid on any platform (provided that we define INVALID_SOCKET as `-1` on non windows, which we do in mongoose).

The inverse check `sock == INVALID_SOCKET || sock >= FD_SETSIZE` is also correct, both if sock is signed or unsigned.

@pavel-pimenov:
+1 for the reordering of the `fd_set` declaration before the conditional statement.

himanshupsharma:
Reverted (sock < 0) checks.

PR #518, opened 2015-04-26
Wrongly interpretation of vsnprintf return code

CurlyMoo:
Formatting a message after we received a -1 can cause segfaults. This fix properly implements the needed buffer size check and the additional allocation afterwards. The code will also stop gracefully whenever vsnprintf cannot parse the message.

It also fixed a potential memory leak because the buffer was overwritten by a new malloc instead of resized by a realloc.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/518)

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -451,30 +451,24 @@ static void ns_remove_conn(struct ns_connection *conn) {
 // and return allocated buffer.
 int ns_avprintf(char **buf, size_t size, const char *fmt, va_list ap) {
   va_list ap_copy;
+  char nul;
   int len;
 
   va_copy(ap_copy, ap);
-  len = vsnprintf(*buf, size, fmt, ap_copy);
+  len = vsnprintf(&nul, 1, fmt, ap_copy);
   va_end(ap_copy);
 
   if (len < 0) {
-    // eCos and Windows are not standard-compliant and return -1 when
-    // the buffer is too small. Keep allocating larger buffers until we
-    // succeed or out of memory.
-    *buf = NULL;
-    while (len < 0) {
-      if (*buf) NS_FREE(*buf);
-      size *= 2;
-      if ((*buf = (char *) NS_MALLOC(size)) == NULL) break;
-      va_copy(ap_copy, ap);
-      len = vsnprintf(*buf, size, fmt, ap_copy);
-      va_end(ap_copy);
+    DBG(("unproperly formatted message %s", fmt));
+    len = -1;
+  } else {
+    if (len >= (int) size) {
+      // Standard-compliant code path. Allocate a buffer that is large enough.
+      if ((*buf = (char *) NS_REALLOC(*buf, len + 2)) == NULL) {
+        len = -1;
+      }
     }
-  } else if (len > (int) size) {
-    // Standard-compliant code path. Allocate a buffer that is large enough.
-    if ((*buf = (char *) NS_MALLOC(len + 1)) == NULL) {
-      len = -1;
-    } else {
+    if(len > -1) {
       va_copy(ap_copy, ap);
       len = vsnprintf(*buf, len + 1, fmt, ap_copy);
       va_end(ap_copy);


CurlyMoo:
I encountered this issue before when working with vsprintf in my logging library. This is the segfault i occasionally see passing in my program caused by mongoose.
```
(gdb) backtrace
#0  0xb6dc1810 in wcslen () from /lib/arm-linux-gnueabihf/libc.so.6
#1  0xb6dc2a24 in wcsrtombs () from /lib/arm-linux-gnueabihf/libc.so.6
#2  0xb6d831e4 in vfprintf () from /lib/arm-linux-gnueabihf/libc.so.6
#3  0xb6da5b58 in vsnprintf () from /lib/arm-linux-gnueabihf/libc.so.6
#4  0x0002fc14 in ns_avprintf (buf=0xb047e1a8, size=8192,
    fmt=0x925dd0 "{\"devices\":{\"rpi\":{\"uuid\":\"0000-00-00-39-298c8b\",\"origin\":\"0000-00-00-39-298c8b\",\"timestamp\":1430040197,\"protocol\":[\"cpu_temp\"],\"id\":[{\"id\":1}],\"temperature\":50.843,\"poll-interval\":10,\"temperature-off"..., ap=...) at /home/pilight/daemon-dev/libs/pilight/core/mongoose.c:435
#5  0x000364e8 in mg_printf_data (c=0x1705dc,
    fmt=0x925dd0 "{\"devices\":{\"rpi\":{\"uuid\":\"0000-00-00-39-298c8b\",\"origin\":\"0000-00-00-39-298c8b\",\"timestamp\":1430040197,\"protocol\":[\"cpu_temp\"],\"id\":[{\"id\":1}],\"temperature\":50.843,\"poll-interval\":10,\"temperature-off"...) at /home/pilight/daemon-dev/libs/pilight/core/mongoose.c:2717
#6  0x0002b140 in webserver_request_handler (conn=0x1705dc) at /home/pilight/daemon-dev/libs/pilight/core/webserver.c:395
#7  0x0002df80 in webserver_handler (conn=0x1705dc, ev=MG_REQUEST) at /home/pilight/daemon-dev/libs/pilight/core/webserver.c:1000
#8  0x00033498 in call_user (conn=0x1705d8, ev=MG_REQUEST) at /home/pilight/daemon-dev/libs/pilight/core/mongoose.c:1833
#9  0x00039aec in call_request_handler (conn=0x1705d8) at /home/pilight/daemon-dev/libs/pilight/core/mongoose.c:3065
#10 0x0003aa54 in call_request_handler_if_data_is_buffered (conn=0x1705d8) at /home/pilight/daemon-dev/libs/pilight/core/mongoose.c:3327
#11 0x00040b74 in on_recv_data (conn=0x1705d8) at /home/pilight/daemon-dev/libs/pilight/core/mongoose.c:4732
#12 0x0004270c in mg_ev_handler (nc=0x81ae88, ev=3, p=0xb0481478) at /home/pilight/daemon-dev/libs/pilight/core/mongoose.c:5255
#13 0x00030124 in ns_call (nc=0x81ae88, ev=3, p=0xb0481478) at /home/pilight/daemon-dev/libs/pilight/core/mongoose.c:519
#14 0x0003143c in ns_read_from_socket (conn=0x81ae88) at /home/pilight/daemon-dev/libs/pilight/core/mongoose.c:968
#15 0x00031cb4 in ns_mgr_poll (mgr=0x17f8d8, milli=1000) at /home/pilight/daemon-dev/libs/pilight/core/mongoose.c:1113
#16 0x000416c8 in mg_poll_server (server=0x17f8d8, milliseconds=1000) at /home/pilight/daemon-dev/libs/pilight/core/mongoose.c:4931
#17 0x0002d430 in webserver_worker (param=0x0) at /home/pilight/daemon-dev/libs/pilight/core/webserver.c:839
#18 0xb6f2cc00 in start_thread () from /lib/arm-linux-gnueabihf/libpthread.so.0
#19 0xb6e0e728 in ?? () from /lib/arm-linux-gnueabihf/libc.so.6
#20 0xb6e0e728 in ?? () from /lib/arm-linux-gnueabihf/libc.so.6
```

I'm still trying to show you what is happening in a small unit test, but couldn't reproduce it properly. To show you what i mean with the return code, you can check this small snippet:
```
#include <stdarg.h>
#include <stdio.h>
#include <string.h>

void _printf(const char *format_str, ...) {
	va_list ap;
	char nul;
	
	va_start(ap, format_str);
	printf("%d\n", vsnprintf(&nul, 1, format_str, ap));
	va_end(ap);
}

int main() {
	_printf(NULL);
	return 0;
}
```

This code will segfault on my FreeBSD machine and return -1 on my ARM machine.

CurlyMoo:
I noticed the error occurs when using `mg_printf_data` without any additional arguments. E.g.:
`mg_printf_data(conn, "DATE_ADD(%H:%M:%S)")`
Although this should not segfault the vsnprintf function, i cannot find why it does. Replacing the `mg_printf_data` with `mg_send_data` has resolved the issue as well.

However, the `-1` interpretation is still wrong on at least Linux:
```
If an output error is encountered, a negative value is returned. 
```

An outputted string that caused issues in my logging library was (thus returning `-1` from `vsnprintf`):
`Memory usage is: %f%` instead of `Memory usage is: %f%%`.

cpq:
Thanks for the status!

CurlyMoo:
Will there be any actions taking to solve the wrongly parsed `-1` on linux?

cpq:
Before doing that, I'd like to understand the reason for the crash.
I can see that passing `Memory usage is: %f%` format string can make `vsnprintf()` to return -1.
I do not understand however what causes a crash:

```
```

CurlyMoo:
> Although this should not segfault the vsnprintf function, i cannot find why it does.

cpq:
If that is possible to do, I'd appreciate a bit more info on a backtrace you've shown above.
Could you switch to frame 4, and dump all arguments, and values of local vars please (e.g. len, size, buf) ?

CurlyMoo:
It's no use. I could even trigger the fault with this `%S\n`.

Issue #510, opened 2015-04-14
Directories not listed


monoped:
I get directory listings only if the client is in the same subnet as the server. If the server is contacted from a different subnet, the response code is 200 and I get an empty file. Is that intended? Or am I missing something? Everything else works fine. 
The server  (monoped.crabdance.com) is running on a BeagleBoneBlack and accessed via dynamic DNS. Mongoose is the latest version from Github, the configuration is:

    document_root /local/bernd/www
    run_as_user bernd
    listening_port 80
    access_log_file /local/bernd/access_log
    index_files index.lsp,index.html
    cgi_pattern **.lsp$

- Bernd

cpq:
That is unexpected.
Would you be able to get tcpdump captures on the server side for both good and bad cases, please?

monoped:
On 15.04.2015 08:08, Sergey Lyubka wrote:
> That is unexpected.
> Would you be able to get tcpdump captures on the server side for both
> good and bad cases, please?

Sure. Attached are 2 dumps created with

	tcpdump -vvA -i eth0 'tcp port 80'

If you need different options, please let me know.

- Bernd


13:37:05.645127 IP (tos 0x0, ttl 53, id 32855, offset 0, flags [DF], proto TCP (6), length 60)
    37.84.34.0.39483 > beagle.http: Flags [S], cksum 0xcfa9 (correct), seq 362560388, win 14600, options [mss 1452,sackOK,TS val 2372451283 ecr 0,nop,wscale 9], length 0
E..<.W@.5...%T"....p.;.P..;.......9............
.h.........	o.t.
13:37:05.645496 IP (tos 0x0, ttl 64, id 0, offset 0, flags [DF], proto TCP (6), length 60)
    beagle.http > 37.84.34.0.39483: Flags [S.], cksum 0x4722 (correct), seq 2152172542, ack 362560389, win 14480, options [mss 1460,sackOK,TS val 9468947 ecr 2372451283,nop,wscale 6], length 0
E..<..@.@.0P...p%T"..P.;.G....;...8.G".........
..|..h......
13:37:05.669110 IP (tos 0x0, ttl 53, id 32856, offset 0, flags [DF], proto TCP (6), length 52)
    37.84.34.0.39483 > beagle.http: Flags [.], cksum 0xae47 (correct), seq 1, ack 1, win 29, options [nop,nop,TS val 2372451308 ecr 9468947], length 0
E..4.X@.5...%T"....p.;.P..;..G.......G.....
[...]

PR #522, opened 2015-05-15
scan_directory should default to document root if passed an empty uri

emonti:
Otherwise... scan_directory returns an empty listing in the document root if a request was made for httx://server:port

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/522)

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -3506,6 +3506,11 @@ static int scan_directory(struct connection *conn, const char *dir,
   int arr_size = 0, arr_ind = 0, inc = 100;
   DIR *dirp;
 
+  // default to document root if passed an empty uri
+  if (dir && strcmp(dir, "") == 0) {
+    dir = conn->server->config_options[DOCUMENT_ROOT];
+  }
+
   *arr = NULL;
   if ((dirp = (opendir(dir))) == NULL) return 0;
 


cpq:
Thank you.
Is that related to #510 ?
How can I reproduce an empty listing?

emonti:
I saw #510 also but wasn't sure if this was related. It may be?

Now, also, I can't seem to reproduce this issue on my project. So feel free to ignore, sorry.

emonti:
Ok, update... I actually can reproduce this. It occurs if the document root is explicitly set to "/". I know seems like you wouldn't want this anyway. But in my case, I do for troubleshooting an embedded device.

You can reproduce with the following server:
```
#include "mongoose.h"

int main() {
    struct mg_server *server = mg_create_server(NULL, NULL);
    mg_set_option(server, "listening_port", "9099");
    mg_set_option(server, "document_root", "/");
    printf("Starting on port %s\n", mg_get_option(server, "listening_port"));
    for (;;) {
        mg_poll_server(server, 100);
    }
    mg_destroy_server(&server);
    return 0;
}

```

emonti:
... and I think i see where the issue is now, along with a better fix. See: #523

PR #523, opened 2015-05-15
prevent truncating '/' document root

emonti:
This was causing directory listings to fail on the root directory when a document root of "/" was specified.

See also #522 

(possibly related: #510)

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/523)

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -3319,9 +3319,9 @@ static int find_index_file(struct connection *conn, char *path,
     }
   }
 
-  // If no index file exists, restore directory path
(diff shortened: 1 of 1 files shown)


cpq:
Thank you.
Out of curiosity, what kind of embedded board are you working on?
We're also working on a set of embedded boards from different vendors.

emonti:
Thanks for fixing. I am not using a development board. I'm working with a 3rd party consumer product running linux on ARM, and building with a clang cross-compiler toolchain.

cpq:
Eric, thanks for the background!

PR #524, opened 2015-05-15
Implementing array varibles

NickCis:
Mongoose didn't implement a way for retrieving array variables like:
```
http://www.example.com?foo[]=1&foo[]=2&foo[]=3
```

If you try to get them through `mg_get_var` you only get the first one. This patch implements `mg_get_n_var` which gets the `n` aparition of the desired var. An exmaple of how to use it is in `examples/array_vars`.

The basic idea is to iterate using `mg_get_n_var` until the function returns `-1`, which means that no more array value was found. E.g.:
```
// Supposing this is called inside an ev_handler.
// conn is a valid mg_connection
char buffer[1024];
int ret,
  i;
for(i=0; (ret = mg_get_n_var(conn, "foo[]", buffer, 1024, i)) > 0; i++)
  mg_printf_data(conn, "\nfoo[%d] = %s", i, buffer);
```

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/524)

### DIFF
--- a/examples/array_vars/Makefile
+++ b/examples/array_vars/Makefile
@@ -1,0 +1,21 @@
+# Copyright (c) 2014 Cesanta Software
+# All rights reserved
+
+PROG = array_vars
+CFLAGS = -W -Wall -I../.. -pthread -g -O0 $(CFLAGS_EXTRA)
+SOURCES = $(PROG).c ../../mongoose.c
+
+all: $(PROG)
+
+run: $(PROG)
+	./$(PROG)
+
+$(PROG): $(SOURCES) Makefile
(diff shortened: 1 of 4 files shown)


NickCis:
I've corrected the mistakes you've highlighted.

cpq:
Nick, can I ask you to agree to http://cesanta.com/contributors_la.shtml please?

cpq:
FYI: your PR broke unit test, I had to fix them in https://github.com/cesanta/mongoose/commit/14d6f71f8260da894a7ab4c78a14e8f4b61bddc8

NickCis:
I've just signed the agreement.
I'm sorry about breaking the unit test, i didn't realize the proyect had tests, i forgot to check that :(

PR #540, opened 2015-06-05
multithread webserver with root path setting sample

songtzu:
this sample allow user to setting the root path, not exactly the same as /example/multi_threaded_server/

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/540)


cpq:
Introduced functionality (setting the root path) is something that is not worth extra example for.
Instead, I'd update multi_threaded_server example.

PR #542, opened 2015-06-11
Update web_server.c

mirbuf:
This patch adds to mongoose webserver the possibility to specify multiple times some of the options which require comma separated values.
url_rewrites is the most useful of these, and having comma separated url in a single row can be quite difficult to admin.
Prepending a value with '^' char will not append the value, but will set it as it would normally do.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/542)

### DIFF
--- a/examples/web_server/web_server.c
+++ b/examples/web_server/web_server.c
@@ -151,6 +151,18 @@ static char *sdup(const char *str) {
   return p;
 }
 
+static char *sdup_append(const char *str, const char * str2, const char sep) {
+  char *p;
+  char strsep[2] = { 0, 0 };
+  strsep[0] = sep;
+  if ((p = (char *) malloc(strlen(str) + strlen(strsep) + strlen(str2) + 1)) != NULL) {
+    strcpy(p, str);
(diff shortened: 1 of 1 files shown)


cpq:
Thank you. Could you show an example please?

mirbuf:
Sure:

# Set first URL rewrite
url_rewrites /dir/=r:/temp/
# append a second rewrite to the previous
url_rewrites /LoU/=r:/last-of-us/
# that is equivalent to have
# url_rewrites /dir/=r:/temp/,/LoU/=r:/last-of-us/

# Now set a new rewrite that will remove previous definitions
url_rewrites ^/new/=r:/newdir/

Same goes with options which require comma separated values, like

- access_control_list
- index_files
- extra_mime_types

It would be a good idea to raise MAX_OPTIONS define to an higher value (i.e. 1000)

cpq:
I think that is too obscure and difficult to understand.
The price to pay (extra code) is too big for the benefit it brings.
Not integrating this - but thanks for sending the patch, though!

PR #545, opened 2015-06-16
Update file_upload.c

meijia85:
the old version while of mg_parse_multipart is wrong, it may endless loop or lose data

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/545)


cpq:
Thank you.
Pushed https://github.com/cesanta/mongoose/commit/824e4dc0e457183dd5bd6606dd772f95fc715160

PR #553, opened 2015-07-09
web_server: allow space in configuration file

wimpunk:
This patch fixes an issue when using a configuration file with comments in it.  It stops reading after the second word and doesn't keep on reading until the comment sign or the end of line.

According to https://github.com/cesanta/mongoose/blob/master/docs/SSL.md you can add a comment after a configuration line and this fix makes the parsing work correctly.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/553)

### DIFF
--- a/examples/web_server/web_server.c
+++ b/examples/web_server/web_server.c
@@ -212,7 +212,7 @@ static void process_command_line_arguments(char *argv[], char **options) {
         continue;
       }
 
-      if (sscanf(line, "%s %[^\r\n#]", opt, val) != 2) {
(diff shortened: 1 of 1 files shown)


cpq:
Yes, that is the intended logic. Primarily, for files names with spaces inside.
Can you give an example when the current logic doesn't behave as expected?

wimpunk:
My example in your own examples.  Take a look at the example configuration in https://github.com/cesanta/mongoose/blob/master/docs/SSL.md .  If you use that configuration it doesn't work.  I think you should either change the example or change the code.

cpq:
I actually removed `examples/web_server` alltogether.

PR #574, opened 2015-10-04
Do not use mg_connection.user_data

sebgod:
In versions < 6.0, `user_data` was not touched by mongoose, which is a
very useful property and some software might require it to be unchanged.
This patch uses `proto_data` instead.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/574)


cpq:
That cannot be integrated, case for both cases, proto_data is being used and this patch overrides it.

Issue #579, opened 2015-10-16
Memory leak at function 'mg_destroy_conn()' when compiling with SSL (defining 'MG_ENABLE_SSL').


rantoniello:
Memory leak at function 'mg_destroy_conn()' when compiling with SSL (defining 'MG_ENABLE_SSL').

Proposed solution:

Add the following remarked lines to 'mg_destroy_conn()' (modifications are marked by tag '//RAL')

> static void mg_destroy_conn(struct mg_connection *conn) {
>   if (conn->sock != INVALID_SOCKET) {
>     closesocket(conn->sock);
>     /*
>      * avoid users accidentally double close a socket
>      * because it can lead to difficult to debug situations.
>      * It would happen only if reusing a destroyed mg_connection
>      * but it's not always possible to run the code through an
>      * address sanitizer.
>      */
>     conn->sock = INVALID_SOCKET;
>   }
>   mbuf_free(&conn->recv_mbuf);
>   mbuf_free(&conn->send_mbuf);
> #ifdef MG_ENABLE_SSL
>   if (conn->ssl != NULL) {
>     SSL_free(conn->ssl);
>   }
>   if (conn->ssl_ctx != NULL) {
>     SSL_CTX_free(conn->ssl_ctx);
>   }
>   // { //RAL: Add to avoid leak
>   //SSL_COMP_free_compression_methods();
>   sk_SSL_COMP_free(SSL_COMP_get_compression_methods());
>   EVP_cleanup();
>   CRYPTO_cleanup_all_ex_data();
>   printf("Check point cleanup.\n");
>   // }
> #endif
>   MG_FREE(conn);
> }

Verification setting (Tested on Ubuntu 14.04 LTS):

Compile 'simplest_web_server' example with the following little modifications to add SSL support (modifications are marked by tag '//RAL').
Execute with VALGRIND or similar tool.

Makefile modifications (set '\<my_ssl_include_dir\>' and '\<my_ssl_lib_dir\>' accordingly):

> PROG = simplest_web_server
> SOURCES = $(PROG).c ../../mongoose.c
> CFLAGS = -W -Wall -I../.. $(CFLAGS_EXTRA)
> CFLAGS+= -I'\<my_ssl_include_dir\>' -L'\<my_ssl_lib_dir\>' -lssl -lcrypto -DMG_ENABLE_SSL #//RAL
>
> all: $(PROG)
>
> $(PROG): $(SOURCES)
[...]

PR #580, opened 2015-10-16
Solve leak when releasing SSL resources.

rantoniello:
Solve memory leak at function 'mg_destroy_conn()' when compiling with SSL (defining 'MG_ENABLE_SSL').
Refer to issue #579.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/580)


rojer:
these functions are not part of the standard OpenSSL API and thus cannot be merged.
i don't think there is a problem here, we test for leaks in SSL mode and there aren't any currently.

Issue #564, opened 2015-08-25
Do not use memcmp when comparing strings


Naav:
There is a number of uses of `memcmp` to compare strings. For example (but not only), in the following extract from `mongoose.c`, lines 696-704:

    if (memcmp(str, "ssl://", 6) == 0) {
      str += 6;
      *use_ssl = 1;
    } else if (memcmp(str, "udp://", 6) == 0) {
      str += 6;
      *proto = SOCK_DGRAM;
    } else if (memcmp(str, "tcp://", 6) == 0) {
      str += 6;
    }

But in the case where `str` is shorter than the string it is being compared against, this results in an out-of-bounds access. A possible fix would be to use `strcmp` or `strncmp`.

mmikulicic:
Hi @Naav and @nitsel88

For the record,  `memcmp` as used in this case won't cause any out of bound memory access.

`memcmp` as defined by ISO/IEC 9899:1990 states:

```
The memcmp() function returns zero if the two strings are identical, otherwise returns the difference between the first two differing bytes
```

This means that if `str` is shorter than `"ssl://"` in the above example, `memcmp` will compare the NUL terminator of `str` with a char from `"ssl://"` and then return.

Indeed ASAN doesn't issue any violations in the following code:

```c
#include <stdio.h>
#include <string.h>

int main() {
  char b1[] = "abcd";
  char b2[] = "ab";
  if (memcmp(b1, b2, 2) == 0) {
    printf("equal\n");
  }
  return 0;
}
```

Implementations of memcmp might compare full words at a time, as long as it doesn't break the mandated behaviour. Out of bound reads are fine as long as the reads don't trigger side effects and as long as tooling such as valgrind and ASAN know about it. Usually any kind of memory access control or (side effects due to memory mapped IO) happen at least at a word boundary (usually a page boundary or equivalent).

---

That said, I don't see why we shouldn't use `strncmp` here, if for anything for readability, given that we already perform by access on that string (hence the only benefit of `memcmp` 32-bit only memory access on some platform is invalidated). This particular use has already been changed.

cpq:
Thanks for reporting - as Marko said, that's not a bug.
Closing this.

Naav:
Your snippet seems to be backwards with respect to the problem I described in my first post.. It should be:

```
#include <stdio.h>
#include <string.h>

int main() {
  char b1[] = "ab";
  char b2[] = "abcd";
  if (memcmp(b1, b2, 4) == 0) {
[...]

Issue #536, opened 2015-05-26
For gcc 4.6.3 warning remove


ywhwang:
Here's my patch.
///////////////in mg_url_decode
  return i >= src_len ? (int)j : -1;

cpq:
That's fiexed now- closing.

PR #583, opened 2015-10-30
commit for fixing issue #564

nitsel88:
The memcmp was replaced with strncmp as it would be safe to compare two strings

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/583)


mmikulicic:
please amend the commit message `5.3.6` -> `536`

Would you please stash the two commits in one ? We have some tooling that assumes one commit per PR.

Would you also please check in the contributor agreement at https://docs.cesanta.com/contributors_la.shtml ?

cpq:
Both #536 and #564 are not an issue, closing.
Thanks!

Issue #586, opened 2015-11-10
cgi zombies


monoped:
Any call of a cgi program leaves a zombie process in the system, which gets a bit annoying after a while (tested with haserl and bash on Linux). Maybe somewhere a wait() is missing.

PR #595, opened 2015-12-02
Inhibit zombie production

monoped:
This inhibts the zombie production caused by cgi execs (#586).

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/595)

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -5745,6 +5745,7 @@ static pid_t start_process(const char *interp, const char *cmd, const char *env,
   char buf[500];
   pid_t pid = fork();
   (void) env;
+  signal(SIGCHLD, SIG_IGN);
 
   if (pid == 0) {
     /*


cpq:
@mmikulicic please take a look at it

mmikulicic:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[mongoose.c, line 5748 \[r1\]](https://reviewable.io:443/reviews/cesanta/mongoose/595#-K4wqcCOMbOJ6ey81KMh:-K4wqcCOMbOJ6ey81KMi:-1902324569) ([raw file](https://github.com/cesanta/mongoose/blob/b74b2548b65733b77f1b33d0993596f3d0856954/mongoose.c#L5748)):*
hmm, when mongoose is used as a library, I don't think we should make decisions such as this that can affect how the rest of the application behaves.

Furthermore not all OSes (e.g. QNX) reap zombies if SIGCHLD is ignored. On QNX, for example, waitpid() fails if SIGCHLD is ignored, making system() non-functional.

I have the feeling this choice doesn't belong to mongoose: it's the responsibility of whoever writes `main`.

(I'm open to arguments though)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cesanta/mongoose/595)*

monoped:
On 07.12.2015 15:39, Marko Mikulicic wrote:
> Review status: 0 of 1 files reviewed at latest revision, 1 unresolved
> discussion.
>
> ------------------------------------------------------------------------
>
> /mongoose.c, line 5748 [r1]
> <https://reviewable.io:443/reviews/cesanta/mongoose/595#-K4wqcCOMbOJ6ey81KMh:-K4wqcCOMbOJ6ey81KMi:-1902324569>
> (raw file
> <https://github.com/cesanta/mongoose/blob/b74b2548b65733b77f1b33d0993596f3d0856954/mongoose.c#L5748>):/
> hmm, when mongoose is used as a library, I don't think we should make
> decisions such as this that can affect how the rest of the application
> behaves.
>
> Furthermore not all OSes (e.g. QNX) reap zombies if SIGCHLD is ignored.
> On QNX, for example, waitpid() fails if SIGCHLD is ignored, making
> system() non-functional.
>
> I have the feeling this choice doesn't belong to mongoose: it's the
> responsibility of whoever writes |main|.

Ah, your feeling is right. I've been confused by the output of 'ps ax' 
(I'm on Linux). Of course my main() becomes parent of the cgi processes, 
so SIGCHLD handling must be done there. Works fine now.

- Bernd

PR #603, opened 2015-12-16
Flb

jhindin:
There are few minor issue in the load balancer, apparently fixed by the proposed patch

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/603)


cpq:
Thank you Joseph

PR #606, opened 2015-12-17
Ssl proxy

jhindin:
The patch introduces the function 'mg_set_ssl_server', supposedly quite useful for implementing explicit SSL proxy.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/606)

PR #611, opened 2016-01-03
Setting SSL server state for already open socket

jhindin:
The pull request superseeds request #606 and takes into account rojer's ideas how to avoid code duplication in a new function mg_set_ssl_server. 
The function itself is necessary when socket starts as in a plain mode and changed into SSL later, like in the HTTP SSL proxy with bumping.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/611)


cpq:
Quick glance over it suggests that this PR is redundant - `mg_set_ssl` does the job.
Closing this.

PR #626, opened 2016-02-28
including winsock2.h if ipv6 enabled

f4bsch:
By default windows.h loads ```winsock.h```, not ```WinSock2.h```. The latter is required (plus ```Ws2ipdef.h```) for IPv6 support.

Also note that in the mongoose.h there is a pragma comment to link with Winsock2 lib, but mongoose actually uses the Winsock1 headers.

[Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/626)

### DIFF
--- a/mongoose.h
+++ b/mongoose.h
@@ -147,6 +147,10 @@
 #ifdef _MSC_VER
 #pragma comment(lib, "ws2_32.lib") /* Linking with winsock library */
 #endif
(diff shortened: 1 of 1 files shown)


rojer:
committed in 1d0a89879ceb6e27bd4933f1e2c804bdb7f43275, thanks

PR #630, opened 2016-03-14
fix response of http range request broken when range value is large, …

julee:
…exceeding 50 bytes length.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/cesanta/mongoose/630)


rojer:
committed as 15c333feb077c59f845af55fb8c36d10932c90e2
thank you!

PR #637, opened 2016-04-04
Fix epoll() accept bug.

marayl:
The `mg_ev_mgr_epoll_ctl()` function calls `abort()` after encountering the following error when a client TCP connection is accepted:

`epoll_ctl(5, EPOLL_CTL_MOD, 7, {EPOLLIN, {u32=20605408, u64=749695152056800}}) = -1 ENOENT (No such file or directory)`

N.B. `-DMG_MGR_EV_MGR=1` is used to enable `epoll()` on Linux.

The problem is that `mg_add_conn()` is called in `mg_if_accept_new_conn()` before `mg_sock_set()` has set the socket descriptor for the new connection:

`mg_ev_mgr_epoll_ctl  0x144c9e0 -1 1`
`mg_if_accept_new_conn 0x144fa68 0x144c9e0 -1 0, (nil) (nil)`
`mg_sock_set          0x144c9e0 7`

Caveat: the extent of my testing has been limited to plain TCP sockets, so similar problems may exist elsewhere (such as SSL).

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/637)

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -2616,7 +2616,6 @@ struct mg_connection *mg_if_accept_new_conn(struct mg_connection *lc) {
   nc->proto_handler = lc->proto_handler;
   nc->user_data = lc->user_data;
   nc->recv_mbuf_limit = lc->recv_mbuf_limit;
-  mg_add_conn(nc->mgr, nc);
   DBG(("%p %p %d %d, %p %p", lc, nc, nc->sock, (int) nc->flags, lc->ssl_ctx,
        nc->ssl));
   return nc;
@@ -3204,6 +3203,7 @@ static void mg_accept_conn(struct mg_connection *lc) {
     return;
   }
   mg_sock_set(nc, sock);
+  mg_add_conn(nc->mgr, nc);
 #ifdef MG_ENABLE_SSL
   if (lc->ssl_ctx != NULL) {
     nc->ssl = SSL_new(lc->ssl_ctx);


rojer:
the truth is, epoll support is broken, in more ways than one.
it also do not bring much benefit, as we still have to walk the list of connections to deliver polls.
do you see a big difference versus select()?

rojer:
the truth is, epoll support is broken, in more ways than one.
it also does not bring much benefit, as we still have to walk the list of connections to deliver polls.
do you see a big difference versus select()?

marayl:
One of my users raised this as an issue when trying to run with `epoll()`. (See https://github.com/swirlycloud/swirlyc/issues/55) Not a major concern for me at this stage; if the `epoll()` implementation is buggy, then I will simply document this as a known issue and disable support for it on my end.

I would suggest, however, that since `epoll()` is notionally supported in Mongoose, it's still worth taking this specific fix. At least it will 1) prevent duplication of effort, 2) may avoid scaring off new customers, and 3) may highlight other issues that also need fixing. A comment for the unwary close to the following would also be good:

`#define MG_MGR_EV_MGR 0 /* select() */`

What are your plans for `epoll()` support longer term?

rojer:
as mentioned above, it doesn't bring much in terms of improvement anyway. Mongoose is more about simplicity than scale, so epoll support is just extra complexity. i'm thinking of removing it altogether.

rojer:
https://github.com/cesanta/mongoose/commit/19dd02acfd443f12ce278e7d269db872393a7f4f removed epoll support

PR #638, opened 2016-04-05
Update mongoose.h

choday:
use mongoose.h without openssl

---

This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/638)


rojer:
this is not the right way of doing it, instead, code that uses these members should be ifdefed.

@alashkin, looks like compilation w/o SSL is broken. can you make sure that Mongoose compiles w/o SSL? add a compile test too.

alashkin:
@choday @rojer 
Just tried to compile a couple of mongoose examples with and w/out SSL - everything is OK [for me]

@choday 
What errors do you have?

alashkin:
Closed due inactivity

PR #639, opened 2016-04-05
mg_poll_conn - sends message to a single connection

JohnDearlight:
We wrote this simple mod so that we could send a message to a single connection without triggering a broadcast poll.

In our applications, we been passing the connection in the message.  The handler would then decide whether to accept or reject the message by comparing the 'nc' mg_connection argument to the connection passed in the message.  We wanted to avoid this broadcast kind of logic by targeting a specific connection.  By adding the connection to the ctl_msg, a very simple mod, the ctl_msg handler can easily determine dispatch to that connection immediately without polling every connection.

We figured that this mod is sufficiently non-intrusive to warrant asking that it be merged into the main release.

---

This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/639)

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -114,6 +114,7 @@ MG_INTERNAL int mg_is_not_modified(struct http_message *hm, cs_stat_t *st);
 #endif
 
 struct ctl_msg {
+  struct mg_connection* conn;
   mg_event_handler_t callback;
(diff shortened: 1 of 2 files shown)


cpq:
It is unsafe, because even holding a pointer to a connection in another thread is unsafe.

PR #641, opened 2016-04-10
Adds visualgdb working example

benjamind:
I finally got Mongoose working under VisualGDB.

Had to modify the normal esp8266.mak that VisualGDB provides to link in libcirom and libmirom.

Also found I had to use the compact eagle.v6.512.compact.ld file found in this commit https://github.com/cesanta/mongoose/commit/d2b9ad570fd51acce86741b73af33291ad39e644

---

This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/641)


rojer:
first, let's call it ESP8266_VisualGDB
there's a ton of duplication here... first, let's not check in mongoose.c and .h again.
then it looks like some of these files are autogenerated, can be regenerated and presumably don't need to be checked in.
please make sure only the files that are absolutely necessary are included and then i'll take a closer look

benjamind:
Yeah sure thing, just wanted to get this up so you could take a look. I'm still finding my way around VisualGDB so have to figure out what can and can't be autogenerated. I'll work this up a bit more then send you a ping to take another look.

cpq:
@benjamind ping - do you have an updated version ?

benjamind:
Afraid I haven't come back to this in quite some time. Sorry!

cpq:
All right - then I am closing this issue, please reopen it back if you have an update.
Thanks so much for sending a PR in the first place!

PR #644, opened 2016-04-13
Fix crash in mg_stat on MinGW-w64

spuschhof:
Fixing a crash in mg_stat when compiling with MinGW-w64 on 32bit Windows. MinGW-w64's definition of stat can not be converted to _stati64, but _stati64 is always defined, so _stati64 should be used on MinGW-w64.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/644)

### DIFF
--- a/mongoose.h
+++ b/mongoose.h
@@ -209,7 +209,7 @@ typedef uint32_t in_addr_t;
 #define INT64_FMT "I64d"
 #define INT64_X_FMT "I64x"
 #define SIZE_T_FMT "Iu"
-#ifdef __MINGW32__
+#if defined __MINGW32__ && !defined __MINGW64_VERSION_MAJOR
 typedef struct stat cs_stat_t;
 #else
 typedef struct _stati64 cs_stat_t;


rojer:
thank you.

can we use `__MINGW64__` here? i understand that for 64 bit version, both are defined.

spuschhof:
`__MINGW64__` is only defined when compiling in 64bit mode, but the crash occures in 32bit mode.
According to [this list](https://sourceforge.net/p/predef/wiki/Compilers/), the only macros defined for MinGW-w64 in both 32 and 64bit mode, but not for MinGW32, are `__MINGW64_VERSION_MAJOR` and `__MINGW64_VERSION_MINOR`.

rojer:
ok, sounds good. i need to apply this to our internal repo, so i won't merge this exact PR, but i'll mention you in the commit.
thanks again!

rojer:
published in 8a62005a2d840e4d91440c75ecc45de4c491368f

Issue #438, opened 2014-11-16
 GET http://127.0.0.1:8080/packed_build/ammo.asm.js net::ERR_CONTENT_LENGTH_MISMATCH


makc:
Chrome randomly does this, on Mac OS.

Response headers report correct length
```
HTTP/1.1 200 OK
Date: Sun, 16 Nov 2014 20:08:31 GMT
Last-Modified: Sun, 09 Nov 2014 03:57:13 GMT
Etag: "545ee619.3800928"
Content-Type: application/x-javascript
Content-Length: 3800928
Connection: close
Accept-Ranges: bytes
```
but for some reason it terminates after 2.3 MB:
![screen shot 2014-11-16 at 21 48 48](https://cloud.githubusercontent.com/assets/242577/5063152/007249c4-6ddb-11e4-9984-3e30dca2dc18.png)

I don't have any means to reproduce it, except random encounter. Posting just so you keep your eye on it.

cpq:
Can't reproduce this.
Is it still happening? What is the setup?

makc:
I don't use mongoose with that project any more, their node-based server now handles everything itself. And, in fact, 

> doesn't change during download

it might have been actually copied over by the script that was triggered by html request from node server.

f00bard:
I'm seeing this behavior running the "simplest_web_server" example under Cygwin.  Pulled the latest from master and compiled.  This is with Chrome on Windows 7 x64.

![snip1](https://cloud.githubusercontent.com/assets/382860/12493885/601491c0-c055-11e5-85f4-62e45915f307.PNG)

![snip2](https://cloud.githubusercontent.com/assets/382860/12493887/64ad8ad4-c055-11e5-9a4b-7d8a632f412f.PNG)

Adding some additional debug out put I can see that when send() returns -1 errno is set to EAGAIN.

cpq:
Rojer,
could you take a look at this please?

Premature connection drop on Windows, when send() returns -1 & errno set to
EAGAIN.
Is there any way to simulate that in unit test?

On 21 January 2016 at 20:43, Mack Hooper <notifications@github.com> wrote:

> I'm seeing this behavior running the "simplest_web_server" example under
> Cygwin. Pulled the latest from master and compiled. This is with Chrome on
> Windows 7 x64.
>
> [image: snip1]
> <https://cloud.githubusercontent.com/assets/382860/12493885/601491c0-c055-11e5-85f4-62e45915f307.PNG>
>
> [image: snip2]
[...]

PR #650, opened 2016-04-22
Check for return of EAGAIN/EWOULDBLOCK when sending

f00bard:
This issue is still present in Mongoose 6.4 on Cygwin.  To reproduce, build the simplest_web_server example and serve from a directory with an HTML file that references a few other files (in my case, serving a packed .js of ~750KB).  Issue occurs intermittently but with great frequency.

#438

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/650)


rojer:
i'm not going to merge this exact PR, but it do have one that does similar thing.
in particular, we should only be doing it for non-ssl connection , because SSL code does its own error/not an error check.

rojer:
closing in favor of https://github.com/cesanta/mongoose/commit/55398684ccfcbc13b42db114ac248d9a2b4ac762

thanks!

PR #652, opened 2016-04-26
Fixed compilation with MG_DISABLE_SOCKETPAIR

danielinux:
When MG_DISABLE_SOCKETPAIR is selected, not all calls to `mg_socketpair` are compiled out.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/652)

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -3785,6 +3785,12 @@ int mg_socketpair(sock_t sp[2], int sock_type) {
 
   return ret;
 }
+#else
+int mg_socketpair(sock_t sp[2], int sock_type) {
+    (void)sp;
+    (void)sock_type;
+    return -1;
+}
 #endif /* MG_DISABLE_SOCKETPAIR */
 
 #ifndef MG_SOCKET_SIMPLELINK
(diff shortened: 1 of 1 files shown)


rojer:
it's not quite the right thing to do.
the root of the problem is that CGI requires socketpair and CGI-handling code will not compile without it. however, your PR makes it compile and not work, which is worse.
the right thing to do is to automatically disable CGI if socketpair is disabled. i'll make a change that does that and will close this PR in favor of that.
thanks for reporting though, i'll mention you in the commit.

danielinux:
Thanks, this fixes our problem. Upstream mongoose now compiles with frosted-OS using `-DMG_DISABLE_CGI` 

https://github.com/insane-adding-machines/frosted-userland/commit/ce07f60c6d366a91b379e28d184bdf219657d547

rojer:
after https://github.com/cesanta/mongoose/commit/21665e367c2dc002092a019f269caf4016879337 you shouldn't need explicit `-DMG_DISABLE_CGI` anymore but i suggest that you keep it, for clarity.

Issue #516, opened 2015-04-22
Add server option to allow CORS


ListerTheTormentor:
Can't config the server to allow CORS, The server need config options to allow the header include Access-Control-Allow-Origin and Access-Control-Allow-Credentials.

cpq:
I think a generic option like "extra_headers" would do the job.

cpq:
Pushed  https://github.com/cesanta/mongoose/commit/0c95290f932c33dad3c778c24a9428134ae90b8a

PR #657, opened 2016-05-09
Send extra options with OPTIONS

magcius:
In order to support CORS, browsers make OPTIONS requests to "preflight"
certain requests. We need to make sure that we reply to such requests
with our custom header options.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -7463,9 +7463,11 @@ static void mg_http_send_digest_auth_request(struct mg_connection *c,
             domain, (unsigned long) time(NULL));
 }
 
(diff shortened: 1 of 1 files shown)


telesma:
Any progress with this pull request?

telesma:
#516 will do the job

magcius:
That commit (which was back in 2015) was then erased when 518a7f8a71e87537040efcec9d7cf026d968aaca happened. As far as I know, this is still relevant.

telesma:
@magcius I don‘t know why such a regression could happen. CORS handling is a must feature for a webserver!

PR #669, opened 2016-06-14
initialize char array

Jackarain:
Signed-off-by: jack jack.wgm@gmail.com

---

This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/669)


cpq:
There is no need to init those buffers.

PR #688, opened 2016-08-25
Copy remote peer's address to spawned thread's local connection

sunxiaoguang:
When threading is enabled, the sa field of connection created for spawned thread is not copied, therefore can't get real peer's address and port can be fixed by this patch.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/688)


sunxiaoguang:
It was added by in master branch already, so I'm closing it.

PR #689, opened 2016-08-25
Use socketpair directly on unix platform which all have this syscall.

sunxiaoguang:
Use socketpair directly on unix platform which all have this syscall on recent
versions. Current way of creating socketpair is more portable but may block
for nondeterministic time when calling connect. Also it reduce number of
syscalls have to make

---

This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/689)


cpq:
Integrated, https://github.com/cesanta/mongoose/commit/d1d8a77bf60fdce43ca3ee45e9d51161d12de1b6
Thank you.

Issue #711, opened 2016-10-08
mg_match_prefix fails to match pattern with single asterisk


jagoyette:
mg_match_prefix fails to match a glob pattern with a single asterisk (_). For example, with the given string "/api/user/12345/friends" and pattern "/api/user/_/friends", I would expect mg_match_prefix to return a positive value indicating a match. However, -1 is returned.

The following code can be used to reproduce this issue:

```
#include "mongoose.h"

int main()
{
    const char *str = "/api/user/192/friends";
    const char *pattern = "/api/user/*/friends";

    int numBytes = mg_match_prefix(pattern, strlen(pattern), str);
    if (numBytes < 0) {
        printf("No match found\n");
    } else {
        printf("Match found: %d bytes\n", numBytes);
    } 
}
```

PR #712, opened 2016-10-08
Fixed issue #711 - mg_match_prefix fails to match single asterisk

jagoyette:
The change in this branch fixes the problem where mg_match_prefix fails to properly match a glob pattern with a single asterisk. I tested it with the following code:

```
#include "mongoose.h"

int main()
{
    const char *str = "/api/user/192/friends";
    const char *pattern = "/api/user/*/friends";

    int numBytes = mg_match_prefix(pattern, strlen(pattern), str);
    if (numBytes < 0) {
        printf("No match found\n");
    } else {
        printf("Match found: %d bytes\n", numBytes);
    } 
}
```

---

This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/712)

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -7900,7 +7900,7 @@ int mg_match_prefix_n(const struct mg_str pattern, const struct mg_str str) {
         len = str.len - j;
       } else {
         len = 0;
-        while (j + len != str.len && str.p[len] != '/') {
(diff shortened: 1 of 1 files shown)


cpq:
@alashkin please verify, add unit test, and integrate.

alashkin:
Fixed in one of previous commits
Thank you

PR #716, opened 2016-10-12
Avoid snprintf linking errors on MinGW.

iSLC:
Don't create the `_snprintf` and `_vsnprintf` aliases for `snprintf` and `vsnprintf` as it could lead to linker issues on MinGW. Occurred on recent MinGW-w64 versions.

---

This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/716)

### DIFF
--- a/mongoose.h
+++ b/mongoose.h
@@ -198,9 +198,11 @@
 #define STR(x) STRX(x)
 #define __func__ __FILE__ ":" STR(__LINE__)
 #endif
+#if !defined(__MINGW32__) && !defined(__MINGW64__)
 #define snprintf _snprintf
-#define fileno _fileno
 #define vsnprintf _vsnprintf
+#endif
+#define fileno _fileno
 #define sleep(x) Sleep((x) *1000)
 #define to64(x) _atoi64(x)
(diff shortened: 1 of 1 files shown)


mongoose.c
```diff
@@ -2870,7 +2870,9 @@ static int mg_accept_conn(struct mg_connection *lc) {
   /* NOTE(lsm): on Windows, sock is always > FD_SETSIZE */
   sock_t sock = accept(lc->sock, &sa.sa, &sa_len);
   if (sock == INVALID_SOCKET) {
-    if (mg_is_error(-1)) DBG(("%p: failed to accept: %d", lc, errno));
+    if (mg_is_error(-1)) {
```
mmikulicic: while I personally like explicit braces, omitting them is not so brittle as you imply here.

```sh
$ gcc -DCS_NDEBUG mongoose.c
```

```c
...
  sock_t sock = accept(lc->sock, &sa.sa, &sa_len);
  if (sock == (-1)) {
    if (mg_is_error(-1)) ;
    return 0;
  }
...
```

As you can see the `;` is not part of the macro and thus creates an (empty) statement regardless of whether `DBG` expands to nothing.

Please remove this commit from this PR

PR #719, opened 2016-10-14
Fix misspelled words in source code and documentation

thdot:
This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/719)


cpq:
@alashkin please integrate

alashkin:
Most of mentioned stuff is generated, thus, it cannot be changed directly.
Root causes fixed in b5dd54a404dccd4f80811368a52d46996aed058d
Thanx!

PR #747, opened 2016-11-24
Add a method to modify connect default time in non-blocking mode

datonli:
Add a method to modify connect default time in non-blocking mode. Image in some general use occasions which the server ip given doesn't exist, due to setting non-blocking mode in mongoose default, it'll run just like right and we don't know the client is trying to connect nonexistent server until getting through a default time of connect, after that, select function would find the socketfd return read-write available both, actually connect is failed totally. The time is decided by system where you run, and it's probably between 75 seconds to a few minutes. It's wasting the time for our client at least 75 seconds especially for some occasions strict requirments for time. So the time need to be modified in some way.

Hope this would help community~~ LOL

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/747)


cpq:
@osDanielLee could you merge all commits into one please?

cpq:
Closing as there is no response.

PR #757, opened 2016-12-09
Prevent conflict with openssl

mat007:
When linking openssl and mongoose in the same binary the MD5_Init, MD5_Update and MD5_Final symbols are defined twice. Renaming the ones in mongoose works around this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/757)


cpq:
@alashkin please verify and integrate

alashkin:
@cpq @mat007 
I think -DDISABLE_MD5 does the same
Pls, reopen PR if I'm wrong

Issue #607, opened 2015-12-19
 cgi stress test, mongoose dead


sigusr1:
In function open_cgi_endpoint, if there is no sock left( such as on linux, you create 1024 fd and reach max limit), follow while will loop forever.

```
  do {
    ns_socketpair(fds);
  } while (fds[0] == INVALID_SOCKET);
```

cpq:
I guess we shall limit spin count ?

rojer:
we should not spin at all. no socket -> 500.

cpq:
I am pretty sure I've seen spurious errors, supposedly due to interrupted syscall during socketpair creation, hence the spin.

rojer:
i see the comment, yes. but we should handle `EINTR` instead of just spinning.
i'll do it.

cpq:
thx

jmgregory:
I'm having this issue as well.  I can get around it by setting `MG_ENABLE_BROADCAST=0`, but then I also have to turn off multi-threading.

Is it possible that too many sockets are being opened but not being closed?  My system only has 2-3 concurrent clients, max.  The requests come in very fast, but they're single-threaded on the client sides.  I set `MG_DISABLE_HTTP_KEEP_ALIVE`, but it didn't seem to help.

jmgregory:
Confirmed, sockets are not being closed by mongoose.  I used `lsof` to count open socket connections while my processes were running.  My clients have about 20 total, while my server has 1013.

My event handler matches (only) on `MG_EV_HTTP_REQUEST`, and *every* code path after a match goes through the following function:

    void make_response(struct mg_connection *conn, int status_code, const char *body, bool is_json)
    {
        if (body == NULL || strlen(body) == 0) {
            mg_send_head(conn, status_code, 0, NULL);
        }
        else {
            int len = strlen(body);
            mg_send_head(conn, status_code, len, is_json ? "Content-Type: application/json" : NULL);
            mg_send(conn, body, len);
        }
        conn->flags |= MG_F_SEND_AND_CLOSE;
    }

My event handler doesn't match any other events.  Is there something I'm missing?

jmgregory:
More info: the sockets that aren't closing appear to be some kind of control sockets made by mongoose, not the HTTP request sockets.  They're created by this chunk of code at `mongoose.c:3582`:

    void mg_socket_if_init(struct mg_iface *iface) {
      (void) iface;
      DBG(("%p using select()", iface->mgr));
[...]

PR #761, opened 2016-12-16
Close dummy mg_mgr sockets when MG_ENABLE_BROADCAST is set (Issue 607)

jmgregory:
When MG_ENABLE_BROADCAST is set, the dummy mg_mgr struct in `spawn_handling_thread()` creates its own socket pair, which never gets closed.  This pull requests closes the pair once we're done with `dummy`.

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/761)

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -4731,6 +4731,12 @@ static void spawn_handling_thread(struct mg_connection *nc) {
   c[0] = mg_add_sock(&dummy, sp[0], forwarder_ev_handler);
   c[1] = mg_add_sock(&dummy, sp[1], nc->listener->priv_1.f);
 
+#if MG_ENABLE_BROADCAST
+  if (dummy.ctl[0] != INVALID_SOCKET) closesocket(dummy.ctl[0]);
+  if (dummy.ctl[1] != INVALID_SOCKET) closesocket(dummy.ctl[1]);
+  dummy.ctl[0] = dummy.ctl[1] = INVALID_SOCKET;
+#endif
+
   /* link_conns replaces priv_2, storing its value */
   poll_timeout = (intptr_t) nc->priv_2;
 


cpq:
@alashkin please verify this

alashkin:
This was a part of multithreading API, which was deleted.
Closing this.

cpq:
I don't think this is related to multithreading.
It looks like it's about closing socketpairs - see issue #607.
@alashkin could you double-check please?

alashkin:
@cpq 
`spawn_handling_thread` function was called from `multithreaded_ev_handler` _only_
Current master doesn't have this function at all

alashkin:
If this is about #607 - proposed in this PR fix is wrong anyway

jmgregory:
Yes, on the surface, my issue seemed the same as #607, but once I had dug into it a bit and found the real cause, it might have been better to open a new issue.  Not that it matters anyways, since multithreading has been removed entirely.

While we're on that subject, removing multithreading seems like a pretty dramatic change to me.  Is there an announcements board or some place I could have found out you were about to do that?  Our app as built depends on multithreading; now we're stuck with the older version of Mongoose.  Our client paid for a Mongoose license.  If we'd have known how unstable the codebase was going to be, we might have looked elsewhere...

cpq:
@jmgregory Justin thanks for the background.

It was particularly a MT support that was unstable and constantly giving us problems - you can search issues to see the history. The whole mechanism was not robust.

We'd be happy to collaborate to make a better solution for your client - are you open to do that? If yes, please describe a minimal case which would simulate the real life, and we can come up with an example that could be used as a skeleton.
Would that work for you?

jmgregory:
Thanks, Sergey, I appreciate the offer!  When I get some time, I'll go back and see if the newer version of the single-threaded event API can meet our needs.  If not, I'll get in touch with you.  At this point, part of me is thinking, "if it ain't broke, don't fix it".

alashkin:
@cpq 
Reassigning to you
Fill free to return this issue back to me if there is subject to fix/implement/whatever

cpq:
Ok closing this, @jmgregory please let us know if you need an assistance from our side!

Issue #740, opened 2016-11-18
websocket connection is interrupted and closed with code 1006 in binary messages with more than 1.8MB only in Chrome


lmiguelmh:
I needed to send some pdf files to a local websocket server and found that when sending a file with more than approx. 1.8MB the websocket "CLOSED" with code=1006 CLOSE_ABNORMAL (https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent). This issue only happens with Chrome (54.0.2840) and not Firefox (47.0.1) and always in the same position (curiously enough this position varies depending on the file). The message never arrives to the server and the connection is closed.

I've added some log messages (I hope it is clear). I am using Chrome here:

	001F18C0 after fd=128 nc_flags=256 rmbl=1834831 smbl=0
	mg_if_poll MG_EV_POLL event
	mg_http_handler ??? event
	001F18C0 fd=128 fd_flags=1 nc_flags=256 rmbl=1834831 smbl=0
	001F18C0 8192 bytes (PLAIN) <- 128
	001F18C0 8192 1834831   <-----   here is the last position 1,834,831 bytes 
	mg_recv_common MG_EV_RECV event
	001F18C0 >BEF proto ev=3 ev_data=0045F070 flags=256 rmbl=1843023 smbl=0
	mg_ws_handler ??? event
	001F18C0 >BEF user ev=3 ev_data=0045F070 flags=256 rmbl=1843023 smbl=0
	MG_EV_RECV: 8192
	001F18C0 >AFT user flags=256 rmbl=1843023 smbl=0
	001F18C0 >AFT proto flags=1280 rmbl=1843009 smbl=0   <---   the flag changes?!

	001F18C0 after fd=128 nc_flags=1280 rmbl=1843009 smbl=0
	mg_if_poll MG_EV_POLL event
	mg_http_handler ??? event
	mg_close_conn MG_EV_CLOSE event   <-----   don't understand why the CLOSE event is generated
	001F18C0 1280 128
	001F18C0 >BEF proto ev=5 ev_data=00000000 flags=1280 rmbl=1843009 smbl=0
	mg_ws_handler ??? event
	001F18C0 >BEF user ev=5 ev_data=00000000 flags=1280 rmbl=1843009 smbl=0
	MG_EV_CLOSE
	127.0.0.1:3448 -- left

The same file doesn't have any problems in Firefox.

[...]

PR #767, opened 2016-12-28
reassembled buffer should not be tested for WEBSOCKET_OP_CLOSE

lmiguelmh:
if reass==1 then frames are appended to the reassembled buffer thus buf[0] will have whatever data wsm.data[0] had (see memmove in line #8882) so is possible that the 8-LSB will be WEBSOCKET_OP_CLOSE and the connection will be closed. In fact this was the reason for the issue https://github.com/cesanta/mongoose/issues/740

---
This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/767)

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -8896,9 +8896,11 @@ static int mg_deliver_websocket_data(struct mg_connection *nc) {
       mbuf_remove(&nc->recv_mbuf, (size_t) frame_len); /* Cleanup frame */
     }
 
-    /* If client closes, close too */
-    if ((buf[0] & 0x0f) == WEBSOCKET_OP_CLOSE) {
-      nc->flags |= MG_F_SEND_AND_CLOSE;
+    if(!reass) {
+      /* If client closes, close too. Only valid when the frame is not reassembled */
+      if ((buf[0] & 0x0f) == WEBSOCKET_OP_CLOSE) {
+        nc->flags |= MG_F_SEND_AND_CLOSE;
+      }
     }
   }
 


cpq:
@alashkin please verify and integrate

alashkin:
Fixed in aed6d05d8c52345840379a0ca1a303da4eee5647
Thanks

lmiguelmh:
Thanks for not accepting my push detecting and fixing this issue took me like 3 days. :(
BTW the comment "if client closes close too" is wrong: the client doesn't want to close. The problem is that this test is only valid when the frame is not a reassembled one. A reassembled frame could have any data and the first byte of that data could be WEBSOCKET_OP_CLOSE.

vhmth:
^ Why would you close @lmiguelmh's PR and push the same code with no explanation as to why that was done?

rojer:
@vhmth this is due to the way our repositories are set up: there is a private repository that we work on and then public repositories, including cesanta/mongoose, are published from it. https://github.com/cesanta/mongoose/commit/aed6d05d8c52345840379a0ca1a303da4eee5647 is such a commit, made to the internal repo and then published from it.

vhmth:
I see @rojer - understood. Should that be added when opening a new PR as a note? That way people won't be upset/confused like @lmiguelmh seemed to be here. Not trying to make a mountain out of a mole hill, but I think mongoose will see a lot of drop-off from potential repeat-contributors if they get their code merged without an explanation.

PR #785, opened 2017-02-17
add simplest basel BUILD file

makdharma:
This change is [Reviewable](https://reviewable.io/reviews/cesanta/mongoose/785)


cpq:
Thanks, but we're not adding a parallel build system at this point.

Issue #826, opened 2017-06-24
OpenSSL build error


hendrikw01:
I get a build error when using `-DMG_ENABLE_SSL`

```
../src/mongoose.c: In function 'mg_ssl_if_ossl_psk_cb':
../src/mongoose.c:4273:33: error: dereferencing pointer to incomplete type 'SSL {aka struct ssl_st}'
     (struct mg_ssl_if_ctx *) ssl->ctx->msg_callback_arg;
                                 ^~
../src/mongoose.c: In function 'mg_ssl_if_ossl_set_psk':
../src/mongoose.c:4323:17: error: dereferencing pointer to incomplete type 'SSL_CTX {aka struct ssl_ctx_st}'
     ctx->ssl_ctx->msg_callback_arg = ctx;
                 ^~
```

```
pkg-config --modversion openssl
1.1.0f
```

PR #831, opened 2017-07-03
Fix for recent openSSL

hendrikw01:
Replaced msg_callback_arg with SSL_CTX_set_app_data / SSL_CTX_get_app_data
This PR fixes: #826


cpq:
@rojer please verify and integrate if valid

rojer:
yes, using `SSL_CTX_get/set_app_data` is the right thing to do here.
there's no need for ifdef check, as these functions are present even in ancient versions of openssl. i simply didn't find them at the time when i was writing this code.
i'm going to create a separate PR to apply this to our internal repo.
thank you!

Issue #832, opened 2017-07-05
Error resolving local hostname on Windows


CfirTsabari:
On windows 7 a specific configuration on the registry could lead up to missing the DNS configured on the computer and therefore returning the default DNS address(8.8.8.8) which mean that resolving of a local host could not be achieved.

What happen is that the code at [this lines](https://github.com/CfirTsabari/mongoose/blob/master/mongoose.c#L10638) basically check to see if the 'NameServer' key exist and if not it check if the 'DhcpNameServer' key is exist. after this it check if the value is empty.
This mean that in case of  a valid 'DhcpNameServer' key and an empty 'NameServer' key the correct value won't be taken from 'DhcpNameServer'.

here an example of a registry bracnh that lead to this error (taken from my computer and trimmed unnecessary data).
[example.txt](https://github.com/cesanta/mongoose/files/1125754/example.txt)

also here is a question regarding this issue at [stackoverflow](https://stackoverflow.com/questions/42451190/client-mongoose-c-app-to-query-lan-dns
)

cpq:
Please verify and integrate

PR #833, opened 2017-07-05
fixing ignoring DhcpNameServer in case that the NameServer key is empty.

CfirTsabari:
fixing ignoring DhcpNameServer in case that the NameServer key is empty.

https://github.com/cesanta/mongoose/issues/832


cpq:
Ruslan, please verify and integrate

PR #876, opened 2017-11-18
Modify the method comments that are inconsistent with the parameters.

tinylcy:
Modify the method comments that are inconsistent with the parameters.


cpq:
Closing this, as the docstrings are gone, and code refactored.
Thank you!

PR #878, opened 2017-11-22
Update api_server.c

pinkra:
Fix bad comparison.


cpq:
Thank you.
Closing this, as  with 7.0, the api_server example is gone.

PR #891, opened 2018-01-29
Fix strict-prototypes, missing-prototypes and shadow warning

herbrechtsmeier:


cpq:
Thank you.
Current source afaik adhere to those rules.

PR #892, opened 2018-01-30
bug: response line and headers are not removed from recv_mbuf after MG_EV_HTTP_REPLY

pukaifei:
bug: When using HTTP1.1 keep-alive, the response line and headers are still in nc->recv_mbuf after MG_EV_HTTP_REPLY.

fix: mongoose.c mg_http_handler mbuf_remove(io, hm->message.len) -> mbuf_remove(io, nc->recv_mbuf.len)


cpq:
Thank you.

That's wrong. This change totally destroys everything in the recv buffer, which might contain the next request.
The unit test has, in fact. `test_http_pipeline()` function.

PR #926, opened 2018-06-03
add line number for log prefix

sigusr1:
I think add line number  for log will be help for debug.


cpq:
Thanks!
Already there!

PR #927, opened 2018-06-05
Non-SSL TCP read now gets all available data

gegzy:
Instead of just reading one data block per mg_poll, this patch now reads all available data and calls the callback per block (as per SSL functionality).

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -3547,14 +3547,14 @@ static void mg_handle_tcp_read(struct mg_connection *conn) {
   } else
 #endif
   {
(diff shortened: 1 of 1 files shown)


rojer:
this part has changed in the most recent version, can you check if it still applies to the [dev branch](https://github.com/cesanta/mongoose/tree/dev) and rebase?
open a new pull request if it does.
thanks!

gegzy:
The dev branch changes work fine, therefore no requirement for my patch. Thank you.

PR #939, opened 2018-06-26
fix bug in mg_mgr_init_opt method

YankunLi:
opts.num_ifaces should be used intead of mg_num_ifaces, if
opts.num_ifaces is not equal to zero, there will be an error here.


YankunLi:
@Sergey Lyubka

rojer:
looks good, integrating the change. this PR will be closed but don't worry, you'll see a commit here soon enough.

thanks @YankunLi !

PR #940, opened 2018-06-26
Fix pipelining issue with multipart upload

dg-dboehi:
We found a issue with the multipart upload handling, when the streaming functionality is enabled.

When a client uses pipelining, and sends two post requests with multipart data directly after each other,
the Mongoose server was unable to process the second request. 

We found that the issue is that the server clears the receive buffer after the first request, and with that deletes part of the second request, which is already in the buffer. 

I've attached both a new test case, as well as a fix for this issue.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -6725,7 +6725,6 @@ static void mg_http_multipart_continue(struct mg_connection *c) {
         break;
       }
       case MPS_FINISHED: {
-        mbuf_remove(&c->recv_mbuf, c->recv_mbuf.len);
         return;
       }
     }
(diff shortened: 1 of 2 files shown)


rojer:
thank you! this looks good, especially like that you included a test.
before your contribution can be accepted, please sign the CLA as documented [here](https://github.com/cesanta/mongoose#contributions).

dg-dboehi:
@rojer My company, duagon AG, has already signed the CLA.

rojer:
ok, integrating. this PR will be closed and a commit by a commit to our internal repo which will later be published to the public repo. you will see it attached to this PR.
thank you for the fix!

PR #944, opened 2018-07-02
Don't manipulate the send buffer when sending MQTT messages

### LINKED ISSUE #943: Publishing multiple MQTT messages eventually fails
Hey,

I'm doing a bit of development over at [OVMS](https://github.com/openvehicles/Open-Vehicle-Monitoring-System-3). I've run into a problem with the mongoose MQTT implementation: When the OVMS boots its publishes to multiple MQTT topics (ca. 100 topics). However at some point it crashes.

The assert at [mongoose.c:10528](https://github.com/cesanta/mongoose/blob/master/mongoose.c#L10528) fails. The only way i can explain this is that the data from the `send_mbuf` is flushed whilst inside `mg_mqtt_publish`, clearing out `send_mbuf` and resetting its `len` to 0. When `nc->send_mbuf.len - old_len` is calculated in [mongoose.c:10635](https://github.com/cesanta/mongoose/blob/master/mongoose.c#L10634-L10635) it overflows and gives a very large integer.

Using gdb i can inspect the `len` parameter of `mg_mqtt_prepend_header`:
```C
#3  0x401001ab in mg_mqtt_prepend_header (nc=0x3f81dd04, cmd=<optimized out>, flags=<optimized out>, len=4294965548)
```

The full backtrace is [here](https://gist.github.com/M4GNV5/6e5bdf64f2a9d55c02bba09d05d77672)

I am unsure on how to fix this. One could either rewrite the MQTT implementation to not rely on modifying `send_mbuf` or one could add a flag to sockets temporarily preventing them from being flushed.

M4GNV5:
Calculating the total packet length first allows to send the packet header before sending the rest of the packet.

This fixes #943

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -10516,18 +10516,13 @@ void mg_set_protocol_mqtt(struct mg_connection *nc) {
   nc->proto_data_destructor = mg_mqtt_proto_data_destructor;
 }
 
-static void mg_mqtt_prepend_header(struct mg_connection *nc, uint8_t cmd,
-                                   uint8_t flags, size_t len) {
+static void mg_send_mqtt_header(struct mg_connection *nc, uint8_t cmd,
+                                uint8_t flags, size_t len) {
   struct mg_mqtt_proto_data *pd = (struct mg_mqtt_proto_data *) nc->proto_data;
-  size_t off = nc->send_mbuf.len - len;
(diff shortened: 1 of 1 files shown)


rojer:
i am integrating this PR into our internal repo with minor changes. it will be closed and a bit later a commit will be published referencing it.

PR #945, opened 2018-07-11
Fix compiler warnings when CS_ENABLE_STDIO is not defined.

jelledevleeschouwer:
Fix -Werror=empty-body when CS_ENABLE_STDIO is not defined


rojer:
thank you! i'm incorporating this into my own related change.

PR #955, opened 2018-08-06
Add support for OpenSSL client SNI

Mellnik:
Adds SNI extension to Client Hello which can be used by setting ``mg_connect_opts.ssl_server_name``.


rojer:
looks good, integrating. thank you!

PR #966, opened 2018-08-29
Add mg_file_upload_handler_string

dooglz:
Add mg_file_upload_handler_string. Keeps uploaded file in memory in a char* string. Returns this string on MG_EV_HTTP_PART_END.


cpq:
Thank you.

The file upload functionality has been revamped.

PR #969, opened 2018-09-09
Add a very basic Meson build file to build Mongoose as shared library.

tsieprawski:
I have my [private pet-project]( https://bitbucket.org/tsieprawski/taskman) using Mongoose. I'm using [Meson](http://mesonbuild.com/) as build system and I've created a very basic Meson build file to include Mongoose, so it is [automatically downloaded and built alongside my code](https://mesonbuild.com/Wrap-dependency-system-manual.html). I think it would be cool to include it upstream. It's very basic, it just builds Mongoose as shared library.


cpq:
Thanks, but that does not belong to the Mongoose repo.

PR #987, opened 2018-11-21
Make MG_CTL_MSG_MESSAGE_SIZE overridable.

awong-dev:
The default buffer size of 8192 for control messages is too large on some
platforms such as the esp32. Allowing the size to be adjustable via
compile time flags allows for mongoose to be used on these platforms
without blowing up the stack.


cpq:
Thank you Albert.
We'll integrate this. Could you please sign a CLA in the mean time?
https://cesanta.com/cla.html

PR #1011, opened 2019-03-20
Modify the flag of NON-BLOCKING for fusion stack

xin-f:
MG uses fcntl(sock, F_SETFL, flags | O_NONBLOCK) to set nbio mode but fusion says: 
#define O_NONBLOCK      0x2000  /* Non-blocking (not supported in Fusion) */

Actually, #define MSG_NONBLOCKING 0x4000  /* override any blocking state */
is used for non-blocking mode.


cpq:
Thank you.

https://github.com/cesanta/mongoose/commit/c5b1c08d05aee3c00d4d5598612c7143203dbc58

PR #1012, opened 2019-03-22
Fix IPv6 connect in mg_socket_if_connect_tcp

zuckschwerdt:
IPv6 addresses in `mg_connect()` currently don't work (tested on unix / MacOS) as `mg_socket_if_connect_tcp()` uses the fixed family `AF_INET` in `socket()` and fixed `sin`-length in `connect()`. It should use `sa_family` from `socket_address` and the `sin6` length if applicable.

### DIFF
--- a/src/mg_net_if_socket.c
+++ b/src/mg_net_if_socket.c
@@ -33,7 +33,7 @@ static int mg_is_error(void) {
 void mg_socket_if_connect_tcp(struct mg_connection *nc,
                               const union socket_address *sa) {
   int rc, proto = 0;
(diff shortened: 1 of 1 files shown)


zuckschwerdt:
This still applies and is needed for IPv6 support in `mg_socket_if_connect_tcp()`.

cpq:
IPv6 listening and client sockets are now tested by the unit test, `make test`.
Verified on both MacOS and Linux.

zuckschwerdt:
Yes, the rewritten code now has working IPv6 in [mg_connect_resolved()](https://github.com/cesanta/mongoose/blob/master/src/sock.c#L306), thanks!

PR #1014, opened 2019-03-23
Set a cache for session id reusage

xin-f:


cpq:
Thank you.
Could you elaborate a bit more on the benefits of this change, please?

cpq:
Thanks.
It looks like that cache is not enabled by default, so won't integrate this.

PR #1049, opened 2019-09-15
Added a script that get mongoose version.

CfirTsabari:
hi,
I have created a script that get mongoose version (for my own usage),  and thought you might want it.
It was inspired from zmq [version.sh](https://github.com/zeromq/libzmq/blob/master/version.sh).

I tried to use the same conventions you used in the other python scripts except for using `sys.stdout.write` since we need a print without spaces when printing version number.


cpq:
```
perl -nle 'print $1 if /^#define\s+MG_VERSION\s+"(.+)"/' mongoose.h
```

Issue #1042, opened 2019-07-19
cookis_auth have bug


ZhangPengchao:
The function login_handler will be called with event MG_EV_CLOSE,and *p is NULL,then it will cause segment fault.It should judge the event like this:
    if(ev != MG_EV_HTTP_REQUEST){
        return;
    }

cpq:
Could you be more specific, where exactly the issue is, please?

atchutameka:
@cpq I believe the issue is with login_handler and logout_handler. They get called with a NULL argument for **void *p** and without a NULL check, it will crash.

I believe @ZhangPengchao  is saying we need a check like this:
```c
static void login_handler(struct mg_connection *nc, int ev, void *p) {
  struct http_message *hm = (struct http_message *) p;
  if (hm!=NULL && mg_vcmp(&hm->method, "POST") != 0) {
```

```c
static void logout_handler(struct mg_connection *nc, int ev, void *p) {
  struct http_message *hm = (struct http_message *) p;
  char shead[100];
  snprintf(shead, sizeof(shead), "Set-Cookie: %s=", SESSION_COOKIE_NAME);
  mg_http_send_redirect(nc, 302, mg_mk_str("/"), mg_mk_str(shead));
  if(hm!=NULL){
      struct session *s = get_session(hm);
 
```

cpq:
Please see https://github.com/cesanta/mongoose/pull/1060

PR #1060, opened 2019-10-08
Fix https://github.com/cesanta/mongoose/issues/1042

nliviu:


cpq:
Thanks @nliviu !
Closing this, as auth function has been refactored into the `mg_http_creds()` API, which fetches auth info from various sources including cookies.

There is a usage example also at https://github.com/cesanta/mongoose/tree/master/examples/multiple-logins

Issue #1065, opened 2019-10-18
Segmentation fault in mg_http_reverse_proxy()


dmundt:
MG_ENABLE_HTTP_URL_REWRITES is set and comma-separated list of `uri_pattern=url_file_or_directory_path` rewrites is used. If `uri_pattern` matches requested URI segmentation fault occurs. It seems `proto_data` field points to 0x0.

Behavior was introduced in Mongoose 6.15 and still exists in 6.16. Possibly related to #1019.

```
Program received signal SIGSEGV, Segmentation fault.
mg_http_reverse_proxy (nc=0x3312c8, hm=0x7efff2f0, mount=..., upstream=...)
    at D:/[...]/source/mongoose.c:8124
8124    D:/[...]/source/mongoose.c: No such file or directory.
(gdb) bt
#0  mg_http_reverse_proxy (nc=0x3312c8, hm=0x7efff2f0, mount=..., upstream=...)
    at D:/[...]/source/mongoose.c:8124
#1  0x00033d84 in mg_http_handle_forwarding (opts=<optimized out>, hm=<optimized out>, nc=<optimized out>)
    at D:/[...]/source/mongoose.c:8173
#2  mg_serve_http (nc=0x3312c8, hm=0x7efff2f0, opts=...)
    at D:/[...]/source/mongoose.c:8549
#3  0x00039494 in Service::OnRequest (nc=0x3312c8, ev=<optimized out>, ev_data=0x7efff2f0)
    at D:/[...]/source/service.cpp:653
#4  0x0002bf00 in mg_call (nc=0x3312c8,
    ev_handler=0x39198 <Service::OnRequest(mg_connection*, int, void*)>, user_data=<optimized out>, ev=0,
    ev_data=0x7efff2f0)
    at D:/[...]/source/mongoose.c:2462
#5  0x0002d47c in mg_http_call_endpoint_handler (nc=0x3312c8, ev=100, hm=0x7efff2f0)
    at D:/[...]/source/mongoose.c:8947
#6  0x0002e1fc in mg_http_handler (nc=0x3312c8, ev=<optimized out>, ev_data=<optimized out>)
    at D:/[...]/source/mongoose.c:6753
#7  0x0002bf00 in mg_call (nc=0x3312c8, ev_handler=0x2def0 <mg_http_handler>, user_data=<optimized out>,
    ev=0, ev_data=0x7efff6e8)
    at D:/[...]/source/mongoose.c:2462
[...]

PR #1066, opened 2019-10-18
Fix segmentation fault in mg_http_reverse_proxy #1065

dmundt:
Fix segmentation fault in mg_http_reverse_proxy() #1065

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -6034,6 +6034,9 @@ MG_INTERNAL struct mg_http_proto_data *mg_http_create_proto_data(
 
 static struct mg_http_proto_data *mg_http_get_proto_data(
     struct mg_connection *c) {
+  if (c->proto_data == NULL) {
+    return mg_http_create_proto_data(c);
+  }
   return (struct mg_http_proto_data *) c->proto_data;
 }
 


anatolyborodin:
[mongoose.c](https://github.com/cesanta/mongoose/blob/master/mongoose.c) is a so called "amalgamated" library. I'm not really sure, if it's being updated automatically or manually during a release. Anyway, [src/mg_http.c](https://github.com/cesanta/mongoose/blob/master/src/mg_http.c) is the original source, and should be updated as well. See e.g. 3fc61a3edc6b45f1218b61a0a560bf0332d5532c

@cpq what is the right way?

anatolyborodin:
The logic regarding `mg_http_get_proto_dat()/mg_http_create_proto_data()` was changed in e5910da300613f0b37fb73f66e5f8c7b96cbd6b9. `mg_http_get_proto_dat()` allocated the missing object before. I'm not sure if it's a right thing to do again (should a getter function allocate anything?)

https://github.com/cesanta/mongoose/commit/e5910da300613f0b37fb73f66e5f8c7b96cbd6b9?diff=unified#diff-e2958a0adcf1d38b24ce49505c902074R670-R672 allows (?) the pointer to be `NULL`.

https://github.com/cesanta/mongoose/commit/e5910da300613f0b37fb73f66e5f8c7b96cbd6b9?diff=unified#diff-e2958a0adcf1d38b24ce49505c902074R3071-R3072 here is an explicit `mg_http_create_proto_data()` call.

mwm29:
It looks like the seg fault occurs on line 8124 of mongoose.c: `mg_http_get_proto_data(be)->reverse_proxy_data.linked_conn = nc;` `be`'s proto_data is `NULL`.

I agree that it doesn't make sense for a getter to allocate data. You could add a check before this line instead of in the getter function. However with the fix in this pr or the fix I proposed, the reverse_proxy example still doesn't work. It no longer seg faults but the request to the backend server just hangs indefinitely. Is anyone else seeing this behavior?

cpq:
Please debug it - run with the hexdump turned on, compile with the debugging on, etc.

mwm29:
I ran the reverse_proxy example and connected with http_client. Attached are the hexdumps from reverse_proxy (hexdump.txt) and from http_client (clienthex.txt).


[hexdump.txt](https://github.com/cesanta/mongoose/files/3997135/hexdump.txt)
[clienthex.txt](https://github.com/cesanta/mongoose/files/3997136/clienthex.txt)

mwm29:
So it looks like reverse_proxy sees the GET on 8000, sends it on 8000, receives it on 8001, 8001 generates and sends a 200, the 200 response is received on 8000, but isn't forwarded back to the client.

TianQingshuai:
You can find the fixes code [https://github.com/cesanta/mongoose/issues/1087](here)
It does work fine.

cpq:
The reverse proxy example was rewritten, adopted for the latest release.
See https://github.com/cesanta/mongoose/tree/master/examples/http-reverse-proxy
Closing this.

PR #1074, opened 2019-11-09
Fix typedef bool for MSC

zuckschwerdt:
When compiling for the `CS_P_WINDOWS` platform with Mingw-w64 (or MinGW) the `stdbool` header is available. The `typedef int bool` alternative should be limited to MSC.


cpq:
https://github.com/cesanta/mongoose/commit/4e89eac8dd979b7da44fd4c8a4a469f384dd5ff5

Thank you.

Issue #1121, opened 2020-06-17
GPL2 licence not included


ullenius:
The full licence text is not included.

The GPL2 states:

> "You may copy and distribute verbatim copies of the Program's source code as you receive it, in any medium, provided that you... **and give any other recipients** of the Program **a copy of this License along with the Program**."

Hence it is usually included in repositories.

cpq:
This software is under a dual license, not solely under GPLv2.

PR #1122, opened 2020-06-17
Add GPL2 full licence

ullenius:
Add full GPL2 licence (https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt)

Solves issue #1121


cpq:
This software is under a dual license, not solely under GPLv2.

PR #1131, opened 2020-07-20
<bug>: avoid realloc(xx, 0)

zuzy:
Call resize(buf, 0) by chance and get double freed....

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -1431,6 +1431,11 @@ void mbuf_free(struct mbuf *mbuf) {
 
 void mbuf_resize(struct mbuf *a, size_t new_size) WEAK;
 void mbuf_resize(struct mbuf *a, size_t new_size) {
+  /*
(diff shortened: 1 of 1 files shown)


rojer:
as long as `realloc(xx, 0)` returns NULL in that case, it should be fine.
what is the compiler we are talking about?

zuzy:
Ubuntu16.04, x86, gcc-5 for now.
I treat mongoose-mqtt as a way to transport files. Publish & on-message callback are called heavily.
Sometimes 'Double free' error occurs by accient, and I found that 'realloc 0' is somehow unclear in standard. 
As a result, 'double free' disappaerd after I did that.

rojer:
ok, we'll need something more substantial. this code is well exercised in many scenarios and behaves correctly.
you can use ASAN to track down your double-free.

cpq:
@zuzy please re-test with the current master.

cpq:
Closing this until confirmed.

Issue #1135, opened 2020-08-12
A buffer overflow error in mg_get_http_header


BushraAloraini:
A buffer overflow error in mg_get_http_header function in mongoose/src/mg_http.c in Mongoose 6.18, where header_names and header_values have a bound of (MG_MAX_HTTP_HEADERS); however, there is no check to ensure that the loop does not exceed the upper bound. A Maliciously crafted  http header can trigger this bug. To fix it, modify the loop condition into:
   
 for (i = 0;i < MG_MAX_HTTP_HEADERS && hm->header_names[i].len > 0  ; i++)

rojer:
fix sent, thank you!

rojer:
actually, this will not happen in practice because mg_http_parse_headers will not populate the last slot:
https://github.com/cesanta/mongoose/blob/86f391c70158a0ee553746e326bc996a3a39b4c0/src/mg_http.c#L408

but it's good to have a range limit too.

BushraAloraini:
Thank you, rojer, for your response.  This bug has been detected using a static analysis tool that is part of a research study. Hence, your input is very important. Actually, the static analysis tool was run with an older version of mongoose, which was used with an Android app (AdAway); at that version, mg_get_http_header is defined in mongoose.c. I just realized that mg_get_http_header is still defined in mongoose.c with the newest version (I thought the definition was moved to a new file). I see that mg_get_http_header that is defined in mongoose.c has been fixed, which is great. Also, I see that mg_get_http_header has been called 22 times in mongoose.c file.  Hence, there are many execution paths, so some might trigger the error.

cpq:
@BushraAloraini so you ran an analysis tool that spotted a false positive, and logged a https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-25756 ?

PR #1138, opened 2020-08-19
Add missing range check in mg_get_http_header()

rojer:
h/t @BushraAloraini

Fixes https://github.com/cesanta/mongoose/issues/1135

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -6160,7 +6160,7 @@ int mg_parse_http(const char *s, int n, struct http_message *hm, int is_req) {
 struct mg_str *mg_get_http_header(struct http_message *hm, const char *name) {
   size_t i, len = strlen(name);
 
-  for (i = 0; hm->header_names[i].len > 0; i++) {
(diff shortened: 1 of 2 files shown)


lnksz:
Isn't https://github.com/cesanta/mongoose/blob/86f391c70158a0ee553746e326bc996a3a39b4c0/src/mg_http_cgi.c#L359 also affected?

mtimperley:
> Isn't
> https://github.com/cesanta/mongoose/blob/86f391c70158a0ee553746e326bc996a3a39b4c0/src/mg_http_cgi.c#L359
> also affected @cpq ?

@lnksz, did you get an answer to your question? After reviewing this commit, I have the same question.

cpq:
@mtimperley current HEAD has CGI support removed, and HTTP codebase heavily refactored.
So, not affected.

mtimperley:
Thanks @cpq that is good to know. I'm currently supporting v6.18 as the plan to move to v7.1 is still some months away. I'm thinking this check will do no harm, so I might as well put it in with the others. Do you agree?

cpq:
That makes sense, @mtimperley , thank you.

PR #1161, opened 2020-11-11
Fix len check in mg_match_prefix_n

zuckschwerdt:
This fixed the `len-- > 0` check in `mg_match_prefix_n()` to `--len > 0` as the postfix operator is applied after the comparison and we already have `len != 0` to guard that. I take it that `len-- ; len > 0` (pseudo code) is meant.


cpq:
Closing this, as `mg_match_prefix()` has been renamed to `mg_globmatch()` and heavily refactored.
See https://github.com/cesanta/mongoose/blob/master/src/util.c#L60
Please feel free too reopen if the issue is still there - in that case, drop a couple of lines to the unit test to showcase it.

PR #1185, opened 2021-01-04
Suppress error: cast from ‘void*’ to ‘int’ loses precision [-fpermissive] in MG_ENABLE_OPENSSL build.

hakabahitoyo:
This PR suppresses the error `cast from ‘void*’ to ‘int’ loses precision [-fpermissive]` in the MG_ENABLE_OPENSSL build.

The cast from `void *` to `int` is illegal in some modern compilers (GCC, Clang). Casting to `ssize_t` is still legal.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -3506,7 +3506,7 @@ fail:
 int mg_tls_handshake(struct mg_connection *c) {
   struct mg_tls *tls = (struct mg_tls *) c->tls;
   int rc;
-  SSL_set_fd(tls->ssl, (int) c->fd);
+  SSL_set_fd(tls->ssl, (ssize_t) c->fd);
   rc = c->is_client ? SSL_connect(tls->ssl) : SSL_accept(tls->ssl);
   if (rc == 1) {
     LOG(LL_DEBUG, ("%lu success", c->id));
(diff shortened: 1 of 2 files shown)


cpq:
Looks good, please sign an CLA and we'll integrate it.
The CLA is at https://cesanta.com/cla.html

cpq:
BTW, which OS/compiler was that?
Current unit tests have openssl target and they pass. Maybe we need to add some extra openssl builds.

hakabahitoyo:
> BTW, which OS/compiler was that?

* g++ (Debian 8.3.0-6) 8.3.0 (with -std=c++17)
* gcc (Debian 8.3.0-6) 8.3.0
* clang version 7.0.1-8+deb10u2 (tags/RELEASE_701/final)
* Debian GNU/Linux 10.7

hakabahitoyo:
> please sign an CLA

OK. I have signed.

> make sure to synchronise an amalgamation with your changes by building mongoose.c and mongoose.h files

OK. This additional commit do that: https://github.com/cesanta/mongoose/pull/1185/commits/1689a0d56b368816cca6c6c6d82601414d6c36a3

PR #1232, opened 2021-03-08
Add mg_mqtt_next_unsub

nopnop2002:
There is mg_mqtt_next_sub.
But There isn't mg_mqtt_next_unsub.
So I added.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -2010,6 +2010,19 @@ int mg_mqtt_next_sub(struct mg_mqtt_message *msg, struct mg_str *topic,
   return new_pos;
 }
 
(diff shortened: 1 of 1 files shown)


cpq:
Shall we just rename the existing `mg_mqtt_next_sub` to something like `mg_mqtt_next_topic` and avoid having two duplicate functions?

nopnop2002:
I respect your opinion.

nopnop2002:
If you avoid having two duplicate functions, I have only one request.

My request is to have the version number raised.

This makes it clear that mg_mqtt_next_sub has been renamed to something like mg_mqtt_next_topic.

cpq:
Refactored and pushed https://github.com/cesanta/mongoose/commit/829827d46fc484fc74876265d23ae4187f058559

Thanks!

PR #1259, opened 2021-04-06
fixed msvc errors around VLAs when using TLS

mcottrell1:
MSVC doesn't like VLAs (e.g. https://gcc.godbolt.org/z/Kn6n4TzaP), which means mongoose won't compile with MSVC if using mbedtls or openssl. This resolves it by using a dynamic allocation.

Original errors from mongoose compile with VS2019:
```
mongoose.c(3671,35): error C2057: expected constant expression
mongoose.c(3671,35): error C2466: cannot allocate an array of constant size 0
mongoose.c(3671,36): error C2133: 'buf': unknown size
```

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -117,9 +117,10 @@ int mg_tls_init(struct mg_connection *c, struct mg_tls_opts *opts) {
     mbedtls_ssl_conf_ca_chain(&tls->conf, &tls->ca, NULL);
 #endif
     if (opts->srvname.len > 0) {
-      char buf[opts->srvname.len + 1];
-      sprintf(buf, "%.*s", (int) opts->srvname.len, opts->srvname.ptr);
+      char mem[128], *buf = mem;
+      mg_asprintf(&buf, sizeof(mem), "%.*s", (int) opts->srvname.len, opts->srvname.ptr);
       mbedtls_ssl_set_hostname(&tls->ssl, buf);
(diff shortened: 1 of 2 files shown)


src/tls.c
```diff
@@ -117,9 +117,12 @@ int mg_tls_init(struct mg_connection *c, struct mg_tls_opts *opts) {
     mbedtls_ssl_conf_ca_chain(&tls->conf, &tls->ca, NULL);
 #endif
     if (opts->srvname.len > 0) {
-      char buf[opts->srvname.len + 1];
-      sprintf(buf, "%.*s", (int) opts->srvname.len, opts->srvname.ptr);
-      mbedtls_ssl_set_hostname(&tls->ssl, buf);
+      char *buf = (char *) calloc(opts->srvname.len + 1, sizeof(char));
```
cpq, line 120: Please use  `mg_vasprintf`:

```c
char mem[128], *buf = mem;
mg_vasprintf(&buf, sizeof(mem), ...);
...
if (buf != mem) free(buf);
```

Also, please sign a CLA: https://cesanta.com/cla.html
Make sure to squash all commits into one, in order to have only one commit per PR.

Thanks!

mcottrell1: updated as per your recommendation (had to use mg_asprintf though). I think I already signed a CLA.

PR #1262, opened 2021-04-13
Chen

Chrischenny:
Change : in mg_socketpair, change the operation macro from AF_INET to AF_UNIX, and socket type from SOCK_DGRAM to SOCK_STREAM

Reason:  socketpair() can only use in AF_UNIX domain, and we usually use text msg (which use to communicate with other modules in our code) rather than streaming media. so, we should use SOCK_STREAM to make sure the msg is reachable

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "files.associations": {
+        "cerrno": "c"
+    }
+}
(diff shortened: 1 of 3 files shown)


cpq:
There is no AF_UNIX on other platforms, like Windows.

PR #1265, opened 2021-04-16
Added secure clearing of iobuf memory

mcottrell1:
This allows mongoose users to enable securing erasing of data in iobufs by setting MG_SECURE_MEMORY. Useful for ensuring any sensitive data that might have been in a http request/response is erased when the connection is cleaned up as per CERT guidelines MSC18-C and MSC06-C.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -1363,12 +1363,27 @@ struct mg_connection *mg_http_listen(struct mg_mgr *mgr, const char *url,
 #endif
 
 
+#ifdef MG_SECURE_MEMORY
+static void *memset_secure(void *v, int c, size_t n)
+{
+  if (v != NULL) {
+    volatile unsigned char *p = (volatile unsigned char *)(v);
+    while (n--) {
+      *p++ = c;
+    }
+  }
+  return v;
+}
+#endif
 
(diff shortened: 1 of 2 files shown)


mongoose.c
```diff
@@ -1363,12 +1363,27 @@ struct mg_connection *mg_http_listen(struct mg_mgr *mgr, const char *url,
 #endif
 
 
+#ifdef MG_SECURE_MEMORY
+static void *memset_secure(void *v, int c, size_t n)
```
cpq, line 1367: How is that different from a standard `memset()` ?

lnksz: See [PVS warning](https://www.viva64.com/en/w/v597/).
But why not reuse memset_s() and co if available?

mcottrell1: it would definitely be preferable to use memset_s where available, but I'm not sure how widespread compiler support for C11 is and don't know how much complexity it would add in terms of checking the C standard etc., hence going for the simplest possible approach

cpq:
Integrated https://github.com/cesanta/mongoose/commit/7061b7202f5757ee632049567d559b9d16187de2

Thank you gents.

PR #1273, opened 2021-04-27
get_chunk_length in http.c has problem

ksanadf:
chunk data analysis fun has some problem.


cpq:
This change does not have any comments, no associated unit test.
Could you at least give an example of data which has issues, please?

ksanadf:
I didn't check the original code. when I see "i != '\r'",  I thought "using i not buf[i]" (maybe sth err here).  Then I make some changes. sorry, It's my fault, I'll closed the request immediately.

Issue #326, opened 2014-02-11
Add top level Makefile and targets for DLL


anatol:
I see that mongoose 5.2 removed the Makefile. I believe that the idea that everyone should download sources and compile as embedded.

Although in some situation it would be great to distribute it as a shared library. Some distros (like Linux Arch) are in strong favor of this way. It would be great if mongoose could be distributed as a shared library.

So it would be nice to have Makefile that:
 - compiles *.so *.a with some predefined set of compile flag (enable all features by default is probably the best idea)
 - installs compiled binary files + header files
 - has 'check' targtet
 - *.pc file with compilation flags that are in sync with shared libraries

cpq:
It has been a while since that issue was filed.
I assume there is no interest anymore to have Makefile target for DLLs.
Closing.

Please reopen and send the PR if still an issue.

anatol:
We still interested in it. For now we just ignore the mongoose Makefile file and call gcc directly from our build script https://aur.archlinux.org/packages/mo/mongoose/PKGBUILD. And IMHO the project makefile should provide such basic functionality rather forcing users to copy-paste it.

jpleau:
Hi, I'm currently looking at the possibility of packaging mongoose for Debian.

The ability to produce both a shared library as well as a standalone webserver in different targets would really be useful for me. Subscribing to this bug :)

Another project I'm considering packaging makes use of mongoose, which is why I would also need to package mongoose (we try to avoid including embedded copies of libraries when packaging software in Debian)

Cheers

CurlyMoo:
> Hi, I'm currently looking at the possibility of packaging mongoose for Debian.

If you need help creating these packages you can always ask. Having lots of experience in apt servers and debian packaging.

renatosilva:
I have recently contributed a libmongoose package for MSYS2 containing both a static library and a DLL. The [packaging recipe](https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-libmongoose-git/PKGBUILD) is currently calling gcc directly though.

jpleau:
Hi

Here are the Makefile and .pc files I made for my Debian package.

https://gist.github.com/jpleau/327b2653633e97d955fe

Any feedback appreciated !

silvioprog:
Hello @anatol ,

Could you compile the DLL? I have interest to compile mongoose as .dll/.so libraries, but I can't understand C/C++ to do it.

Thank you!

PR #1295, opened 2021-06-07
Add static and dynamic lib targets to Makefile

Jasper-Ben:
I know that the topic has come up a few times already (e.g. https://github.com/cesanta/mongoose/issues/326) and I understand you, as the maintainers, understand mongoose source code to be simply added to an existing project.

However, I believe that packaging mongoose as a separate lib does have its advantages, e.g. when packaging it for a Linux distro, e.g. https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=mongoose or https://github.com/iris-GmbH/meta-iris-thirdparty/blob/feature/jaor/mongoose_recipe/recipes-core/mongoose/mongoose_7.2.bb. For the latter I extended the existing Makefile in what I believe to be a pretty unintrusive way.

I'd be happy if you would reconsider adding support for building mongoose as a library. If not, so be it. No hard feelings :slightly_smiling_face:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,8 @@ IPV6 ?= 1
 ASAN_OPTIONS ?=
 EXAMPLES := $(wildcard examples/*)
 EXAMPLE_TARGET ?= example
+PREFIX ?= /usr/local
+SOVERSION = 7.2
 .PHONY: ex test
 
 ifeq "$(SSL)" "MBEDTLS"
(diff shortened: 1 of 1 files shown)


Jasper-Ben:
> Please make this change 10 lines maximum.
> There is no need to declare variables that are used only once.
> 3 lines for the build libs target, 3 lines for the install target, 3 lines for uninstall target. Not more.

:wave: @cpq, I added a second commit which reduces the LOC and only kept those variables I deemed absolutely necessary. If these changes are OK for you now, I will squash the commit and sign the CLA, so you can merge

cpq:
That looks good, please go ahead, squash and sign the CLA!

Jasper-Ben:
Sqashed and CLA done. Feel free to merge, if happy with the change introduced in https://github.com/cesanta/mongoose/commit/8a5190bdf3d8c26e01f588d095d0b759b6165b1e :slightly_smiling_face:

PR #1310, opened 2021-07-09
_CRT_SECURE_NO_WARNINGS redefinition fix

star-ismael-serrano:
_CRT_SECURE_NO_WARNINGS should only be defined if it's not defined to avoid macro redefinition warnings if the project already has it defined.

### DIFF
--- a/mongoose.h
+++ b/mongoose.h
@@ -45,7 +45,9 @@
 #endif  // !defined(MG_ARCH)
 
 #define WIN32_LEAN_AND_MEAN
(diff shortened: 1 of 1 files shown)


cpq:
Again, mongoose.h is an amalgamation - next build swipes away this change.
A source file must be changed instead.

star-ismael-serrano:
> 
> 
> Again, mongoose.h is an amalgamation - next build swipes away this change.
> A source file must be changed instead.

Didn't notice this. Will have it in mind in case I add another pull  request.

Issue #1305, opened 2021-07-04
Unable to compile using Visual Studio


ajtruckle:
I created a empty c++ project and added your mongoose header / source files.

I used your sample:

#include "mongoose.h"

```
static const char* s_web_root_dir = ".";
static const char* s_listening_address = "http://localhost:1000";

static void cb(struct mg_connection* c, int ev, void* ev_data, void* fn_data) {
	struct mg_http_serve_opts opts;
	opts.root_dir = s_web_root_dir;

	if (ev == MG_EV_HTTP_MSG) mg_http_serve_dir(c, (struct mg_http_message *) ev_data, &opts);
}

int main(int argc, char* argv[]) {
	struct mg_mgr mgr;
	mg_mgr_init(&mgr);

	mg_http_listen(&mgr, s_listening_address, cb, &mgr);

	for (;;) mg_mgr_poll(&mgr, 1000);

	mg_mgr_free(&mgr);
	return 0;
}
```

But when I try to compile it:

```
1>main.obj : error LNK2019: unresolved external symbol "void __cdecl mg_mgr_poll(struct mg_mgr *,int)" (?mg_mgr_poll@@YAXPEAUmg_mgr@@H@Z) referenced in function main
1>  Hint on symbols that are defined and could potentially match:
1>    mg_mgr_poll
1>main.obj : error LNK2019: unresolved external symbol "void __cdecl mg_mgr_init(struct mg_mgr *)" (?mg_mgr_init@@YAXPEAUmg_mgr@@@Z) referenced in function main
1>  Hint on symbols that are defined and could potentially match:
1>    mg_mgr_init
1>main.obj : error LNK2019: unresolved external symbol "void __cdecl mg_mgr_free(struct mg_mgr *)" (?mg_mgr_free@@YAXPEAUmg_mgr@@@Z) referenced in function main
1>  Hint on symbols that are defined and could potentially match:
1>    mg_mgr_free
1>main.obj : error LNK2019: unresolved external symbol "struct mg_connection * __cdecl mg_http_listen(struct mg_mgr *,char const *,void (__cdecl*)(struct mg_connection *,int,void *,void *),void *)" (?mg_http_listen@@YAPEAUmg_connection@@PEAUmg_mgr@@PEBDP6AXPEAU1@HPEAX3@Z3@Z) referenced in function main
1>  Hint on symbols that are defined and could potentially match:
1>    mg_http_listen
[...]

PR #1311, opened 2021-07-12
#1305 Solve linking problems in c++ projects

star-ismael-serrano:
This change solves issue #1305 linking problems for users using mongoose in c++ projects.


cpq:
mongoose.h is an amalgamation. Net rebuild will swipe away this change.

cpq:
Pushed https://github.com/cesanta/mongoose/commit/015586950fd3df2ed104d00e7fd922f67d2b1cfe
Thank you.

Issue #1313, opened 2021-07-16
Provide means to determine local port when listening on port 0


Bklyn:
- My goal is: To run mongoose on a dynamic (OS-allocated port)
- My actions were: Listen with an address string like "http://0.0.0.0:0"
- My expectation was:  Some API to access the address the OS gave us (notably the port)
- The result I saw: Mongoose does not call getsockname
- My question is: 

In 6.x mongoose would call `getsockname` after `listen` and update the `sa` argument to `mg_open_listening_socket`.  There was even a comment to the effect that this was supported: /* In case port was set to 0, get the real port number */

The new `mg_open_listener` function takes only a url string and returns only a SOCKET, so it can't do this w/o changing the interface.  Perhaps this could be added to `mg_listen`, calling `getsockname` and populating the returned connection's `c->peer` with the results.

My purpose here is to use mongoose to expose an HTTP endpoint in any number of applications which may be on the same host.  The actual endpoint will be registered with a naming service once the server is up and running.

# Environment

- mongoose version: 7.2-ish (7.2-126-g89757388)
- custom changes?: No
- build environment:
  - OS: Linux
  - Compiler/IDE: gcc
- relevant example (see https://github.com/cesanta/mongoose/tree/master/examples) is: _?_
- relevant documentation topic (see https://cesanta.com/docs/) is: _?_

cpq:
Fixed, thank you.

PR #1314, opened 2021-07-17
#1313 expose listening port via conn->peer

Bklyn:
Pass an `mg_addr` pointer to `mg_open_listener` and use this to populate `conn->peer` for the caller
Addresses #1313

### DIFF
--- a/src/sock.c
+++ b/src/sock.c
@@ -130,22 +130,21 @@ static void mg_set_non_blocking_mode(SOCKET fd) {
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


cpq:
@Bklyn thank you.
Before it is integrated, could you confirm you've signed a CLA, please?

https://cesanta.com/cla.html

Bklyn:
Yes I have

cpq:
@Bklyn integrated, thank you.
Had to fix few minor issues and add a notice in the docs.

Issue #1315, opened 2021-07-18
Not all C compiler does supports `#pragma once`


edubart:
I am using a C compiler that does not support `#pragma once`, then when including `mongoose.h` twice I get compile errors due to double declarations. Please add ifdef guards in `mongoose.h`, that is:

```c
#ifndef MONGOOSE_H
#define MONGOOSE_H

/* header code */

#endif
```

cpq:
@edubart thanks. Could you elaborate what is the toolchain you're using, the target platform and environment, please?

edubart:
> @edubart thanks. Could you elaborate what is the toolchain you're using, the target platform and environment, please?

I'm using https://github.com/vnmakarov/mir for jiting C code.

edubart:
I've just made the PR #1317 with a possible fix for this.

PR #1317, opened 2021-07-18
Add ifdef guards, fixes issue #1315

edubart:
See https://stackoverflow.com/questions/1143936/pragma-once-vs-include-guards for other people rationale on this.
I followed a tip that an user suggested there, to use ifdef guards and pragma once this way:

```c
#ifndef BLAH_H
#define BLAH_H
#pragma once

// ...

#endif
```

### DIFF
--- a/mongoose.h
+++ b/mongoose.h
@@ -14,7 +14,10 @@
 //
 // Alternatively, you can license this software under a commercial
 // license, as set out in <https://www.cesanta.com/license>.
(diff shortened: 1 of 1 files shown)


cpq:
@edubart thank you, integrated
Since `mongoose.h` is generated (amaglamated), a slight modification to the PR made.

Could you also sign the CLA, please? https://cesanta.com/cla.html

PR #1319, opened 2021-07-20
The sanitiser pointed out this problem

ignacionr:
And I believe it's right. Other-side-close should be done only for the 'A' (proxy-client-side) connection.


cpq:
Thank you @ignacionr 
Please sign the CLA before this gets integrated:
https://cesanta.com/cla.html

ignacionr:
> Thank you @ignacionr
> Please sign the CLA before this gets integrated:
> https://cesanta.com/cla.html

This is done. Hope to be of more help in the future, since I truly love this project.

PR #1324, opened 2021-07-30
allow for in-url colon, when port is not specified

### LINKED ISSUE #1323: Mongoose as http client should entertain uris with a colon
- My goal is: To connect as an HTTP client to arbitrary URLs (including those where the host doesn't have a port indication, and the URI contains a colon)
- My actions were: mg_http_connect
- My expectation was: to get a connection to the corresponding host
- The result I saw: ends up failing at DNS (because it parses the url incorrectly)
- My question is: we should cut a host name parsing at either a colon or a slash

# Environment

- mongoose version: trunk
- OS: Ubuntu 21.04
- Compiler/IDE: Clang-12

ignacionr:
fixes #1323

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -4049,7 +4049,7 @@ static struct url urlparse(const char *url) {
       u.port = 0;
     } else if (url[i] == ']') {
       u.port = 0;  // IPv6 URLs, like http://[::1]/bar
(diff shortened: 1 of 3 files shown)


cpq:
Merged with thanks.

PR #1336, opened 2021-08-12
Update FREERTOS + LWIP support

zudwalex:
Fix small problems related to compilation on STM32/FREERTOS /LWIP platform


zudwalex:
@cpq VC98 fixed

PR #1343, opened 2021-08-27
add conn_data to mg_connection

mzihlmann:
conn_data is a hook for the user to attach arbitrary data on the connection itself. this allows the user to manage arbitrary state on the connection without relying on an external database for a mapping id -> void*.

### DIFF
--- a/mongoose.h
+++ b/mongoose.h
@@ -740,6 +740,8 @@ struct mg_connection {
   struct mg_iobuf send;        // Outgoing data
   mg_event_handler_t fn;       // User-specified event handler function
   void *fn_data;               // User-specified function parameter
+  void *conn_data;             // pointer to arbitrary User-speficied 
+                               // data attached to connection
   mg_event_handler_t pfn;      // Protocol-specific handler function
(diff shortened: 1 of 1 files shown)


cpq:
Use `fn_data`

Closing this.

mzihlmann:
`fn_data` is a reference to an object on the listening connection. Assume i only have one listening socket, so for me `fn_data` is like a global value, all incoming connections reference the same copy. I cannot modify this **per-connection**, it would affect the state of all connections currently handled. I could replace it on a connection basis but then i also lose reference to it and abuse mongoose, of course it works, but its more simple and straightforward if mongoose would offer a possibility to have a unique userdata attached to each unique connection id. If i get it wrong please explain it to me, would be greatly appreciated.

cpq:
You could modify it per connection, just set `c->fn_data` to a new value.

Let us explain that better in the docs, because it is not clear, and your confusion in genuine.

cpq:
Updated https://github.com/cesanta/mongoose/tree/master/docs#event-handler-function

Thanks for your patience and for getting in touch! Let us know if it is still unclear.

Issue #1375, opened 2021-10-11
MG_EV_READ does not receive the last response chunk for "mg_http_connect" connection


Akshaykumar2002:
- My goal is: _Write a proxy with support for both HTTP and HTTPS requests with some custom changes_
- My actions were:
I used the HTTP-reverse-proxy example to modify it to work for my specific URL. I made the following changes to test (patch provided below):
```
diff --git a/examples/http-reverse-proxy/main.c b/examples/http-reverse-proxy/main.c
index c6d7d32..61fd575 100644
--- a/examples/http-reverse-proxy/main.c
+++ b/examples/http-reverse-proxy/main.c
@@ -8,7 +8,7 @@
 // To enable SSL/TLS, build it like this:
 //    make MBEDTLS_DIR=/path/to/your/mbedtls/installation
 
-static const char *s_backend_url = "https://cesanta.com";
+static const char *s_backend_url = "http://www.google.com";
 static const char *s_listen_url = "http://localhost:8000";
 
 #include "mongoose.h"
@@ -41,6 +41,7 @@ static void forward_fn(struct mg_connection *c, int ev, void *ev_data, void *fn_
       mg_iobuf_del(&c->recv, 0, c->recv.len);
     }
   } else if (ev == MG_EV_CONNECT) {
+    c->is_hexdumping = 1;
     if (mg_url_is_ssl(s_backend_url)) {
       struct mg_tls_opts opts = {.ca = "ca.pem"};
       mg_tls_init(c, &opts);
@@ -55,7 +56,8 @@ static void fn(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
     struct mg_http_message *hm = (struct mg_http_message *) ev_data;
     // Client request, create backend connection Note that we're passing
     // client connection `c` as fn_data for the created backend connection.
-    c2 = mg_connect(c->mgr, s_backend_url, forward_fn, c);
+    c2 = mg_http_connect(c->mgr, s_backend_url, forward_fn, c);
+    c->is_hexdumping = 1;
     c->fn_data = c2;
 	
     if (c2 != NULL) {
@@ -77,9 +79,10 @@ static void fn(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
 int main(void) {
   struct mg_mgr mgr;
 
-  mg_log_set("3");                               // Set log level
+  mg_log_set("4");                               // Set log level
   mg_mgr_init(&mgr);                             // Initialise event manager
[...]

PR #1388, opened 2021-10-23
Don't delete recv buffer until all callbacks are processed.

jameshilliard:
Potential fix for #1375.


jameshilliard:
I think the issue I was hitting may have a different cause, I'm going to close this for now.

PR #1393, opened 2021-10-31
Sort mime types and remove duplicates

jcorporation:
As the title says. CLA is already signed by me.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -477,7 +477,6 @@ static struct mg_str guess_content_type(struct mg_str path, const char *extra) {
       MIME_ENTRY("zip", "application/zip"),
       MIME_ENTRY("3gp", "video/3gpp"),
       MIME_ENTRY("7z", "application/x-7z-compressed"),
-      MIME_ENTRY("7z", "application/x-7z-compressed"),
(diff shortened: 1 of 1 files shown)


cpq:
The list is intentionally unsorted: the most "popular" entries are at the top.
Please keep the original sorting, and retain only "duplicates" change.
Alternatively, a list could be sorted and a binary search can be employed instead of simple loop.

jcorporation:
Ok, the commit now removes only a duplicate entry.

In my opinion the list is relatively arbitrary. My preferred list would be:
- css
- js
- webp
- png
- jpg
- jpeg 
- svg
- html
- htm
- shtml
- ico
- woff
- woff2
- pdf

This is a list of the most frequently used files extensions.

cpq:
Using a shorter list sounds good.. How did you compile that short list?

jcorporation:
In my experience this are the most used file types for serving a webpage. But looking at my webserver statistics I would move css and js after html. CSS and JS files are often combined to reduce the number of assets to serve. Eventually we can also remove woff, woff2 is widely established.

PR #1423, opened 2021-12-06
Expose tls config

ptsolakos:
Currently there is  no way for the user application to use a custom security configuration.
Only the standard configuration  hard-coded in `mg_tls_init` can be used.

This pull request adds a new mg_tls API function allowing the user to set a callback, with which he can modify the tls security configuration at will.


ptsolakos:
Closing this MR: Config callback should be part of struct mg_tls and not be declared static.

PR #1425, opened 2021-12-07
Add tls configuration callback

ptsolakos:
Currently there is no way for the user application to use a custom security configuration.
Only the standard configuration hard-coded in `mg_tls_init `can be used.

This pull request allows the user to define a callback, with which he can modify the tls security configuration at will.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1838,6 +1838,7 @@ struct mg_tls_opts {
   const char *certkey;   // Certificate key
   const char *ciphers;   // Cipher list
   struct mg_str srvname; // If not empty, enables server name verification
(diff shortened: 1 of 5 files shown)


ptsolakos:
The proposed change will create issues with client code not properly initializing struct mg_tls_opts.

```c
struct mg_tls_opts tls_opts;
```
instead of
```c
struct mg_tls_opts tls_opts = {0};
```

I will make an other proposal adding a new api function which will have the callback as parameter:

```c
int mg_tls_init(struct mg_connection *c, struct mg_tls_opts *opts, void fn(void *))
```

PR #1426, opened 2021-12-08
Add mg_tls_init_modify_conf()

ptsolakos:
Adds possibility to modify the default configuration of the TLS library.

Currently there is no way for the user application to use a custom security configuration.
Only the standard configuration hard-coded in `mg_tls_init()` can be used.

This pull request adds a TLS API function similar to `mg_tls_init()`  with an extra callback as parameter. With this callback the user application can modify the TLS security configuration at will.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1886,6 +1886,27 @@ struct mg_tls_opts opts = {.cert = "ca.pem"};
 mg_tls_init(c, &opts);
 ```
 
+### mg\_tls\_init\_modify\_conf()
+
+```c
+void mg_tls_init_modify_conf(struct mg_connection *c, struct mg_tls_opts *opts, void (*fn)(void *));
+```
+
+Same as `mg_tls_init()` but with the possibility to modify the default configuration of the TLS library.
+
+Parameters:
+- `c` - Connection, for which TLS should be initialized
+- `opts` - TLS initialization parameters
+- `fn` - Function to modify the default configuration of the TLS library. 
+  If NULL, the default configuration of the TLS library will be used as in `mg_tls_init()`.
+  If set, `fn` will be invoked after the default configuration has been applied
+  and before the TLS handshake. In case of mbedTLS, the argument of `fn`
+  will be a pointer to the coresponding `mbedtls_ssl_config` configuration.
+  In case of OpenSSL the argument of `fn` will be a pointer to the coresponding
+  `SSL` object.
+
+Return value: None
+
 ## Timer
 
 ### struct mg\_timer
--- a/mongoose.c
+++ b/mongoose.c
@@ -3773,7 +3773,8 @@ static int rng_get(void *p_rng, unsigned char *buf, size_t len) {
 }
 #endif
 
-void mg_tls_init(struct mg_connection *c, struct mg_tls_opts *opts) {
+static void mg_tls_init_core(struct mg_connection *c, struct mg_tls_opts *opts,
+                             void (*fn)(void *)) {
   struct mg_tls *tls = (struct mg_tls *) calloc(1, sizeof(*tls));
   int rc = 0;
   const char *ca = opts->ca == NULL     ? "-"
@@ -3880,6 +3881,9 @@ void mg_tls_init(struct mg_connection *c, struct mg_tls_opts *opts) {
       goto fail;
     }
   }
+  if (fn != NULL) {
+    fn(&tls->conf);
+  }
   if ((rc = mbedtls_ssl_setup(&tls->ssl, &tls->conf)) != 0) {
     mg_error(c, "setup err %#x", -rc);
     goto fail;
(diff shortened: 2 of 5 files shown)


cpq:
Just pushed an alternative approach, see https://github.com/cesanta/mongoose/commit/f43a5a005c42ac8b5af245c3cce3945ab9e024e3

Please feedback - would it work for you?

ptsolakos:
Thank you for following up on this topic.

In the beginning I was thinking as well to expand `struct mg_tls_opts` with a function callback (see #1425). Then I abandoned the idea because I thought  it may cause backward compatibility issues with client code initializing the struct without a member initialization list. Do you think this is an issue?

Question to your approach:
I would expect that the custom callback is called after default configuration before this line:
https://github.com/cesanta/mongoose/blob/3f5fd89fd7b1716950ef1616aea33a7df41bdea0/src/tls_mbed.c#L154

Can you post a short snippet of a custom callback function you have in mind?

cpq:
By default, the init function supposed to be NULL and thus not called - therefore, there should not be any issue with the backward compatibility.

W.r.t. example.. Could you first show an expected usage of your interface please?

ptsolakos:
With the approach proposed in the beginning, I just modify only the default configuration only on the setting that I am interested in and leave the rest untouched. The callback function assumes that the initialization and default configuration is done by `mg_tls_init()` and the callback is called after initialization and default setting but before the handshake.

For example let's say I want to use the default mbettls configuration and change the minimum accepted RSA key length to 4096 at runtime depending on some condition (default is 2048 when mbedtls is shipped).
I would define a  callback like this:

```c
const mbedtls_x509_crt_profile custom_profile = {
    MBEDTLS_X509_ID_FLAG( MBEDTLS_MD_SHA224 ) |
    MBEDTLS_X509_ID_FLAG( MBEDTLS_MD_SHA256 ) |
    MBEDTLS_X509_ID_FLAG( MBEDTLS_MD_SHA384 ) |
    MBEDTLS_X509_ID_FLAG( MBEDTLS_MD_SHA512 ),/* MDs for signatures */
    0xFFFFFFF, /* Any PK alg    */
    0xFFFFFFF, /* Any curve     */
    4096,/* Minimum size for RSA keys  */
};

static void custom_cb(void* config) {
    mbedtls_ssl_config *conf = (mbedtls_ssl_config *)config;
    conf->cert_profile = &custom_profile;
}
```
and then the application would call the  dedicated interface function like this:

```c
mg_tls_init_custom_conf(conn, tls_opts, custom_cb);
````

Is this confusing from your point of view?

cpq:
Yeah it is very confusing.
Especially TLS initialisation may need more than some post-init tweaks.
The `custom_init()` approach is also quite limiting.
Maybe the right way is just to use `-DMG_ENABLE_CUSTOM_TLS=1`

cpq:
Actually, the recent change exposed TLS structures to the application.
Meaning, `c->tls`  pointer, which is `void *`, can be cast to the actual TLS data, e.g. `https://github.com/cesanta/mongoose/blob/master/src/tls_mbed.h#L21`

That means that after `mg_tls_init()` call an application can to any necessary post-configuration using TLS library API.

In my opinion, that renders the `custom_init()` approach, as well as an approach suggested by you, not necessary. Since the TLS data structures are exposed, there is no need for the extra API.

Could you confirm that you could do the necessary configuration with the latest codebase, please?

ptsolakos:
I understand that the interface proposed in this pull request is not so straightforward. It assumes knowledge to the `mg_init()` internals. It is not easy to provide a clean interface with the ability to have full customization.

Maybe as you mentioned, the right way is to use `-DMG_ENABLE_CUSTOM_TLS=1`. This was not available so far.

Regarding your last proposal to modify the exposed TLS data structure after `mg_tls_init()`, it is not bringing the expected result.

The problem is, that in order for the changes to have an effect, the `mbedtls_ssl_setup()` has to be called only once and before initiating the handshake. This is done at the end of `mg_tls_init()`, so it can not be done again by the application. Please see the documentation of the function [here](https://github.com/ARMmbed/mbedtls/blob/development/include/mbedtls/ssl.h#L1716).

cpq:
Thank you @ptsolakos. I don't see a reason why your code can modify the conf once, and an application cannot.

Anyhow, I read that the solution is to use `DMG_ENABLE_CUSTOM_TLS` - thus I consider this issue closed.
Thanks again for reporting.

ptsolakos:
@cpq thank you as well.

Just for your clarification:

My change was modifying the configuration inside `mg_tls_init()` before calling `mbedtls_ssl_setup()`. Only this way the modification has an effect in the TLS handshake. With the current API, even if the application modifies the configuration afterwards and calls `mbedtls_ssl_setup()` this will have no effect. This is a limitation of Mbed TLS as mentioned in the documentation of the function below. I have not tested it with OpenSSL.

```c
/**
 * \brief          Set up an SSL context for use
 *
 * \note           No copy of the configuration context is made, it can be
 *                 shared by many mbedtls_ssl_context structures.
 *
 * \warning        The conf structure will be accessed during the session.
 *                 It must not be modified or freed as long as the session
 *                 is active.
 *
 * \warning        This function must be called exactly once per context.
 *                 Calling mbedtls_ssl_setup again is not supported, even
 *                 if no session is active.
 *
 * \param ssl      SSL context
 * \param conf     SSL configuration to use
 *
 * \return         0 if successful, or MBEDTLS_ERR_SSL_ALLOC_FAILED if
 *                 memory allocation failed
 */
int mbedtls_ssl_setup( mbedtls_ssl_context *ssl,
                       const mbedtls_ssl_config *conf );
````

PR #1476, opened 2022-02-20
Allow controlling behavior when MG_MAX_RECV_SIZE is reached

jameshilliard:
When `MG_MAX_RECV_SIZE` is reached we may want to stop reading rather than throwing an error as a full buffer may just mean we need to wait for the consumer to clear the buffer on a future poll iteration.

Each poll iteration will call `MG_EV_POLL` which can consume from the buffer to allow reading to continue, we should not assume that the buffer will always be consumed below the recv max after each poll iteration as the consumer may need to apply backpressure to the recv buffer if the consumer is not able to clear the recv buffer faster than read_conn fills it.

Introduce a `MG_EV_BUFFER_FULL` event which allows the user to choose what to do when the buffer is reached, with the existing erroring behavior remaining the default.

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -179,6 +179,7 @@ struct mg_connection *mg_connect(struct mg_mgr *mgr, const char *url,
     c->is_client = true;
     c->fd = (void *) (size_t) -1;  // Set to invalid socket
     c->fn_data = fn_data;
+    c->recv_max = MG_MAX_RECV_SIZE;
     MG_DEBUG(("%lu -1 %s", c->id, url));
     mg_call(c, MG_EV_OPEN, NULL);
     mg_resolve(c, url);
@@ -201,6 +202,7 @@ struct mg_connection *mg_listen(struct mg_mgr *mgr, const char *url,
     LIST_ADD_HEAD(struct mg_connection, &mgr->conns, c);
     c->fn = fn;
     c->fn_data = fn_data;
+    c->recv_max = MG_MAX_RECV_SIZE;
     mg_call(c, MG_EV_OPEN, NULL);
     MG_DEBUG(("%lu %p %s", c->id, c->fd, url));
   }
@@ -214,6 +216,7 @@ struct mg_connection *mg_wrapfd(struct mg_mgr *mgr, int fd,
     c->fd = (void *) (size_t) fd;
     c->fn = fn;
     c->fn_data = fn_data;
+    c->recv_max = MG_MAX_RECV_SIZE;
     mg_call(c, MG_EV_OPEN, NULL);
     LIST_ADD_HEAD(struct mg_connection, &mgr->conns, c);
   }
--- a/src/sock.c
+++ b/src/sock.c
@@ -278,12 +278,15 @@ static long mg_sock_recv(struct mg_connection *c, void *buf, size_t len) {
 // (e.g. FreeRTOS stack) return 0 instead of -1/EWOULDBLOCK when no data
 static void read_conn(struct mg_connection *c) {
   long n = -1;
-  if (c->recv.len >= MG_MAX_RECV_SIZE) {
-    mg_error(c, "max_recv_buf_size reached");
+  if (c->recv.len >= c->recv_max) {
+    mg_call(c, MG_EV_BUFFER_FULL, &n);
+    if (n)
+      mg_error(c, "max_recv_buf_size reached");
   } else if (c->recv.size - c->recv.len < MG_IO_SIZE &&
+             c->recv.size + MG_IO_SIZE <= c->recv_max &&
              !mg_iobuf_resize(&c->recv, c->recv.size + MG_IO_SIZE)) {
     mg_error(c, "oom");
-  } else {
+  } else if (c->recv.size > c->recv.len) {
     char *buf = (char *) &c->recv.buf[c->recv.len];
     size_t len = c->recv.size - c->recv.len;
     n = c->is_tls ? mg_tls_recv(c, buf, len) : mg_sock_recv(c, buf, len);
@@ -406,6 +409,7 @@ static void accept_conn(struct mg_mgr *mgr, struct mg_connection *lsn) {
     c->pfn_data = lsn->pfn_data;
     c->fn = lsn->fn;
     c->fn_data = lsn->fn_data;
+    c->recv_max = MG_MAX_RECV_SIZE;
     mg_call(c, MG_EV_OPEN, NULL);
     mg_call(c, MG_EV_ACCEPT, NULL);
   }
--- a/test/unit_test.c
+++ b/test/unit_test.c
@@ -498,26 +498,55 @@ static void eh1(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
 }
 
 struct fetch_data {
-  char *buf;
-  int code, closed;
+  struct mg_iobuf *buf;
+  int code, closed, done;
 };
 
 static void fcb(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
   struct fetch_data *fd = (struct fetch_data *) fn_data;
   if (ev == MG_EV_HTTP_MSG) {
     struct mg_http_message *hm = (struct mg_http_message *) ev_data;
-    snprintf(fd->buf, FETCH_BUF_SIZE, "%.*s", (int) hm->message.len,
-             hm->message.ptr);
+    if (fd->buf->len == 0) {
+      size_t n;
+      n = fd->buf->size - fd->buf->len;
+      if (hm->message.len <= n)
+        n = hm->message.len;
+      mg_iobuf_add(fd->buf, fd->buf->len, hm->message.ptr, n, MG_IO_SIZE);
+    }
     fd->code = atoi(hm->uri.ptr);
     fd->closed = 1;
+    fd->done = 1;
     c->is_closing = 1;
+    fd->buf->buf[fd->buf->len] = '\0';
     (void) c;
+  } else if (ev == MG_EV_HTTP_CHUNK) {
+    struct mg_http_message *hm = (struct mg_http_message *) ev_data;
+    size_t n;
+    if (fd->buf->len == 0) {
+      fd->code = atoi(hm->uri.ptr);
+      n = fd->buf->size - fd->buf->len;
+      if (hm->head.len <= n)
+        n = hm->head.len;
+      mg_iobuf_add(fd->buf, fd->buf->len, hm->head.ptr, n, MG_IO_SIZE);
+    }
+    n = fd->buf->size - fd->buf->len;
+    if (hm->chunk.len <= n)
+      n = hm->chunk.len;
+    mg_iobuf_add(fd->buf, fd->buf->len, hm->chunk.ptr, n, MG_IO_SIZE);
+    mg_iobuf_del(&c->recv, c->recv.len - n, c->recv.len);
+    if (fd->buf->len >= hm->message.len) {
+      fd->closed = 1;
+      fd->done = 1;
+      c->is_closing = 1;
+      fd->buf->buf[fd->buf->len] = '\0';
+    }
   }
 }
 
 static int fetch(struct mg_mgr *mgr, char *buf, const char *url,
                  const char *fmt, ...) {
-  struct fetch_data fd = {buf, 0, 0};
+  struct mg_iobuf io_buf = {(unsigned char *)buf, FETCH_BUF_SIZE, 0};
+  struct fetch_data fd = {&io_buf, 0, 0, 0};
   int i;
   struct mg_connection *c = mg_http_connect(mgr, url, fcb, &fd);
   va_list ap;
@@ -542,7 +571,7 @@ static int fetch(struct mg_mgr *mgr, char *buf, const char *url,
   mg_vprintf(c, fmt, ap);
   va_end(ap);
   buf[0] = '\0';
-  for (i = 0; i < 250 && buf[0] == '\0'; i++) mg_mgr_poll(mgr, 1);
+  for (i = 0; i < 250 && fd.done == 0; i++) mg_mgr_poll(mgr, 1);
   if (!fd.closed) c->is_closing = 1;
   mg_mgr_poll(mgr, 1);
   return fd.code;
@@ -922,6 +951,9 @@ static void f3(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
     // ASSERT(mg_vcmp(&hm->method, "HTTP/1.1") == 0);
     // ASSERT(mg_vcmp(&hm->uri, "301") == 0);
     *ok = mg_http_status(hm);
+  } else if (ev == MG_EV_HTTP_CHUNK) {
+    struct mg_http_message *hm = (struct mg_http_message *) ev_data;
+    mg_http_delete_chunk(c, hm);
   } else if (ev == MG_EV_CLOSE) {
     if (*ok == 0) *ok = 888;
   } else if (ev == MG_EV_ERROR) {
@@ -1796,6 +1828,97 @@ static void test_invalid_listen_addr(void) {
   ASSERT(mgr.conns == NULL);
 }
 
+struct stream_status {
+  uint32_t polls;
+  size_t sent;
+  size_t received;
+  uint32_t send_crc;
+  uint32_t recv_crc;
+};
+
+// Consume recv buffer after letting it reach MG_MAX_RECV_SIZE
+static void eh8(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
+  struct stream_status *status = (struct stream_status *) fn_data;
+  if (c->is_listening)
+    return;
+
+  ASSERT(c->recv.len <= MG_MAX_RECV_SIZE);
+
+  if (ev == MG_EV_ACCEPT) {
+    // Optimize recv buffer size near max to speed up test
+    mg_iobuf_resize(&c->recv, MG_MAX_RECV_SIZE - MG_IO_SIZE);
+    status->received = 0;
+    status->recv_crc = 0;
+  }
+
+  if (ev == MG_EV_CLOSE) {
+    ASSERT(status->received == status->sent);
+  }
+
+  // Let buffer fill up and start consuming after 10 full buffer poll events
+  if (status->polls >= 10 && ev == MG_EV_POLL) {
+    // consume at most a third of MG_MAX_RECV_SIZE on each poll
+    size_t consume;
+    if (MG_MAX_RECV_SIZE / 3 >= c->recv.len)
+      consume = c->recv.len;
+    else
+      consume = MG_MAX_RECV_SIZE / 3;
+    status->received += consume;
+    status->recv_crc = mg_crc32(status->recv_crc, (const char *) c->recv.buf, consume);
+    mg_iobuf_del(&c->recv, 0, consume);
+  }
+
+  // count polls with full buffer
+  if (ev == MG_EV_BUFFER_FULL) {
+    long *error = (long *) ev_data;
+    ASSERT(*error == -1);
+    *error = 0;
+    status->polls += 1;
+  }
+  (void) ev_data;
+}
+
+// Send buffer larger than MG_MAX_RECV_SIZE to server
+static void eh10(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
+  struct stream_status *status = (struct stream_status *) fn_data;
+  if (ev == MG_EV_CONNECT) {
+    size_t len = MG_MAX_RECV_SIZE * 2;
+    struct mg_iobuf buf = {NULL, 0, 0};
+    mg_iobuf_init(&buf, len);
+    mg_random(buf.buf, buf.size);
+    buf.len = buf.size;
+    mg_send(c, buf.buf, buf.len);
+    status->sent = buf.len;
+    status->send_crc = mg_crc32(0, (const char *) buf.buf, buf.len);
+    mg_iobuf_free(&buf);
+  }
+  (void) ev_data;
+  (void) fn_data;
+}
+
+static void test_http_stream_buffer(void) {
+  struct mg_mgr mgr;
+  const char *url = "tcp://127.0.0.1:12344";
+  uint32_t i;
+  struct stream_status status;
+  mg_mgr_init(&mgr);
+  mg_listen(&mgr, url, eh8, &status);
+
+  status.polls = 0;
+  mg_connect(&mgr, url, eh10, &status);
+  for (i = 0; i < (MG_MAX_RECV_SIZE / MG_IO_SIZE) * 50; i++) {
+    mg_mgr_poll(&mgr, 1);
+    if (status.polls >= 10 && status.sent == status.received) break;
+  }
+  ASSERT(status.sent == status.received);
+  ASSERT(status.send_crc == status.recv_crc);
+
+  mg_mgr_free(&mgr);
+  ASSERT(mgr.conns == NULL);
+}
+
+#undef LONG_CHUNK
+
 static void test_multipart(void) {
   struct mg_http_part part;
   size_t ofs;
@@ -2072,6 +2195,7 @@ int main(void) {
   test_invalid_listen_addr();
   test_http_chunked();
   test_http_upload();
+  test_http_stream_buffer();
   test_http_parse();
   test_util();
   test_sntp();
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,11 +9,19 @@ jobs:
         cc: [gcc, clang]
         target: [test, test++, valgrind]
         ssl: [MBEDTLS, OPENSSL]
-    name: linux ${{ matrix.target }} CC=${{ matrix.cc }} SSL=${{ matrix.ssl }}
+        include:
+          - cc: gcc
+            target: test
+            ssl: MBEDTLS
+            tflags: -DMG_MAX_RECV_BUF_SIZE=4096
+    name: >
+      linux ${{ matrix.target }} CC=${{ matrix.cc }} SSL=${{ matrix.ssl }}
+      ${{ matrix.tflags && format('TFLAGS={0}', matrix.tflags) || '' }}
     env:
       IPV6: 0
       CC: ${{ matrix.cc }}
       SSL: ${{ matrix.ssl }}
+      TFLAGS: ${{ matrix.tflags }}
     steps:
     - uses: actions/checkout@v3
     - name: Install packages
--- a/mongoose.c
+++ b/mongoose.c
@@ -2758,8 +2758,11 @@ static void rx_udp(struct mip_if *ifp, struct pkt *pkt) {
   } else if (c != NULL) {
     c->rem.port = pkt->udp->sport;
     c->rem.ip = pkt->ip->src;
-    if (c->recv.len >= MG_MAX_RECV_SIZE) {
-      mg_error(c, "max_recv_buf_size reached");
+    if (c->recv.len >= c->recv_max) {
+      long n = -1;
+      mg_call(c, MG_EV_BUFFER_FULL, &n);
+      if (n)
+        mg_error(c, "max_recv_buf_size reached");
     } else if (c->recv.size - c->recv.len < pkt->pay.len &&
                !mg_iobuf_resize(&c->recv, c->recv.len + pkt->pay.len)) {
       mg_error(c, "oom");
@@ -3563,6 +3566,7 @@ struct mg_connection *mg_connect(struct mg_mgr *mgr, const char *url,
     c->is_client = true;
     c->fd = (void *) (size_t) -1;  // Set to invalid socket
     c->fn_data = fn_data;
+    c->recv_max = MG_MAX_RECV_SIZE;
     MG_DEBUG(("%lu -1 %s", c->id, url));
     mg_call(c, MG_EV_OPEN, NULL);
     mg_resolve(c, url);
@@ -3585,6 +3589,7 @@ struct mg_connection *mg_listen(struct mg_mgr *mgr, const char *url,
     LIST_ADD_HEAD(struct mg_connection, &mgr->conns, c);
     c->fn = fn;
     c->fn_data = fn_data;
+    c->recv_max = MG_MAX_RECV_SIZE;
     mg_call(c, MG_EV_OPEN, NULL);
     MG_DEBUG(("%lu %p %s", c->id, c->fd, url));
   }
@@ -3598,6 +3603,7 @@ struct mg_connection *mg_wrapfd(struct mg_mgr *mgr, int fd,
     c->fd = (void *) (size_t) fd;
     c->fn = fn;
     c->fn_data = fn_data;
+    c->recv_max = MG_MAX_RECV_SIZE;
     mg_call(c, MG_EV_OPEN, NULL);
     LIST_ADD_HEAD(struct mg_connection, &mgr->conns, c);
   }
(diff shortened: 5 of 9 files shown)


cpq:
Thanks @jameshilliard 
This change, however, defeats the purpose of MG_MAX_RECV_BUF_SIZE
What's the point of having the limit if it is ignored?
The point of this compile-time limit is to prevent a situation when an external client can DoS the server by pushing an infinite data, and let the user to set an automatic guard against the overflow.

If your code is aware about the potential overflow, and manages the incoming data with care, just set the `MG_MAX_RECV_BUF_SIZE` to some very high value. What's the point of disabling the guard?

jameshilliard:
> This change, however, defeats the purpose of MG_MAX_RECV_BUF_SIZE
> What's the point of having the limit if it is ignored?

This shouldn't allow the buffer itself to grow any larger than before. It just changes the behavior upon reaching the limit to halt reading more data into the buffer instead of hard erroring.

> The point of this compile-time limit is to prevent a situation when an external client can DoS the server by pushing an infinite data, and let the user to set an automatic guard against the overflow.

I'm not seeing how this change causes an issue there, the max recv buffer size is still respected, just the behavior upon reaching it is different. A buffer size limit is fundamentally different from a max upload size limit.

> If your code is aware about the potential overflow, and manages the incoming data with care, just set the `MG_MAX_RECV_BUF_SIZE` to some very high value. What's the point of disabling the guard?

But we don't actually want the buffer to grow beyond a particular size(`MG_MAX_RECV_BUF_SIZE` is deliberately far below our max upload size), we want the `MG_MAX_RECV_BUF_SIZE` to be respected so that large uploads don't use too much memory. But upon reaching `MG_MAX_RECV_BUF_SIZE`(which happens if we can't empty the buffer faster than it is being filled) we want to slow down the upload by halting the socket reading until we are able to sufficiently drain it.

Essentially we are forwarding/streaming the data in the buffer to another socket(internal unix socket), however that socket consumer is in some cases significantly slower than the mongoose tcp recv socket, and we can't buffer too much data in memory either(memory limited embedded systems) so we need a way to slow down the reads from the socket(and thus slow down the upload client via TCP back-pressure) to keep the upload from using up all our memory.

cpq:
I see the point, thank you.
I believe the semantics of this behavior is different from what is anticipated from the existing `MG_MAX_RECV_BUF_SIZE`. Basically, you're saying  that the `MG_MAX_RECV_BUF_SIZE` should be a rate limit window, akin to TCP window size.

That makes sense. However, such change requires a little bit more. Your code simply stops reading , but makes select wake up on every loop, which turns event loop into a busy loop. I believe, the rate-limited socket must also be excluded from the `select()` poll in `iotest()`, if the limit is reached.

src/sock.c
```diff
@@ -260,11 +260,12 @@ static long mg_sock_recv(struct mg_connection *c, void *buf, size_t len) {
...
   long n = -1;
   if (c->recv.len >= MG_MAX_RECV_BUF_SIZE) {
-    mg_error(c, "max_recv_buf_size reached");
+    return n;
   } else if (c->recv.size - c->recv.len < MG_IO_SIZE &&
+             c->recv.size + MG_IO_SIZE <= MG_MAX_RECV_BUF_SIZE &&
              !mg_iobuf_resize(&c->recv, c->recv.size + MG_IO_SIZE)) {
     mg_error(c, "oom");
-  } else {
+  } else if (c->recv.size > c->recv.len) {
```
jameshilliard, line 268: I added this which appears to be sufficient to prevent reads when the buffer is full.

jameshilliard:
> Basically, you're saying that the `MG_MAX_RECV_BUF_SIZE` should be a rate limit window, akin to TCP window size.

I mean the name `MG_MAX_RECV_BUF_SIZE` implies that the value is simply the max size `c->recv` is allowed to reach. I guess it's conceptually similar to TCP window sizing except it's capping the application buffer size.

> However, such change requires a little bit more.

Hmm, I do see an issue with the poll loop trying to do zero length reads when the buffer reaches max size which should now be fixed.

> Your code simply stops reading , but makes select wake up on every loop, which turns event loop into a busy loop. I believe, the rate-limited socket must also be excluded from the `select()` poll in `iotest()`, if the limit is reached.

Well the loop calls `mg_call(c, MG_EV_POLL, &now);` here:
https://github.com/cesanta/mongoose/blob/42ba1e4aed62947d401ac56b1d4ab0771f92294c/src/sock.c#L546

The idea is the `MG_EV_POLL` event in turn would then be used to consume the buffer using `mg_iobuf_del` and re-allow reads to occur.

> Also, consider a situation with HTTP upload. Say, you set a limit of 2k. But a client sends a request with lots of headers, where request length itself exceeds 2k (for example, there is a long URL with many query parameters).
What you'll get is a connection silently stalled waiting for the full request to buffer, and that'll never happen.

I think you could drain the buffer with `mg_iobuf_del` in this case in the `MG_EV_POLL` events right?

> What is the rate limit size you plan to apply in practice? What is your environment - MCU, approximate free RAM, connectivity type, RTOS, TLS lib?

Probably the default size limit is fine, we don't need an unusually small buffer, the issue is that we need to be able to handle unusually large uploads(potentially multiple gigabytes that far exceed available system RAM) without increasing the buffer size, target systems are embedded Linux based systems such as raspberry pi's or x86_64 industrial embedded linux systems. Available RAM is likely in most cases to be relatively large, for the systems I work with it typically ranges from 256mb to multiple gigabytes.

In our use case the streamed request from mongoose is essentially forwarded to a unix socket service thread which then parses and extracts an update package(which is basically a cpio archive containing a manifest and potentially various images/archives/binaries) on the fly while being streamed.

In cases where the update package/images are significantly larger than available system ram the update package images can be directly written to non-volatile storage without requiring the full image to be received over the network(it also can cache the image in RAM however this still requires that the mongoose buffer be limited to avoid doubling the RAM requirements as we still have to unpack the update package on the fly into RAM in that case).

The unix socket thread receiving/extracting and writing the update package archive may be effectively limited by multiple factors such as decompression overhead in the case of compressed archives/images and in the case of direct streamed flash writes the limit may effectively be the max IO speed of the non-volatile storage the update package is being written to.

Since connectivity may be wifi or even full gigabit the max upload speeds to mongoose may easily far exceed the speeds the extractor can process the upload.

Usually TLS is not used here as the update archives have their own code signing mechanisms.

jameshilliard:
Added a test to reproduce error and validate buffer behavior when `MG_MAX_RECV_BUF_SIZE` is reached.

jameshilliard:
FYI the test failure being hit here appears to be present in master and unrelated to this set of changes.

cpq:
@jameshilliard thank you.

I am still unsure about feasibility of this change.
I understand that your intention is to use `MG_MAX_RECV_BUF_SIZE` as a traffic shaping constant, which stalls receiving until the receiving side drains `c->recv` buffer.

However, practically, that should not happen. In case of HTTP upload, the event handler should handle `MG_EVHTTP_CHUNK` and drain the buffer as new data arrives - see examples/file-upload.

I don't see any other practical examples when traffic shaping would be required.
Can you elaborate please?

jameshilliard:
> However, practically, that should not happen. In case of HTTP upload, the event handler should handle `MG_EV_HTTP_CHUNK` and drain the buffer as new data arrives - see examples/file-upload.

Well what I'm doing is similar to what I'm simulating with the test I wrote here to reproduce the issue, I'm consuming from the buffer and then calling `mg_iobuf_del` to partially consume from the buffer by catching `MG_EV_READ` and `MG_EV_POLL` events since it's difficult to use the `MG_EV_HTTP_CHUNK` events instead as they expect a corresponding `mg_http_delete_chunk` call which fully removes each chunk from the input buffer(which is not what I want to do), which makes it effectively impossible to partially consume the chunk by operating on `MG_EV_HTTP_CHUNK` events.

Note that if my consumer is unable to accept data from the recv buffer I still want to fill the recv buffer up to the `MG_MAX_RECV_BUF_SIZE` size for performance reasons rather than blocking the thread entirely within the `MG_EV_HTTP_CHUNK` event handler.

The issue with `MG_EV_HTTP_CHUNK` events is that they don't give sufficient granular control in regards to rate limiting the recv buffer as you effectively have to fully consume each chunk whenever there is a `MG_EV_HTTP_CHUNK` event, which may not be possible when the consumer is rate limited. I think in order to use `MG_EV_HTTP_CHUNK` events I would need a way to both defer chunk events entirely(until the next poll cycle once my consumer is filled) as well as a way to partially consume/delete a chunk(the amount deleted would be determined by the amount of data the consumer is capable of accepting).

cpq:
@jameshilliard thanks.
Your change fails unit test if MG_MAX_RECV_BUF_SIZE set to some sane value, e.g. 4096.

jameshilliard:
> Your change fails unit test if MG_MAX_RECV_BUF_SIZE set to some sane value, e.g. 4096.

Hmm, not sure that's a regression, unit tests seem to fail for me without this change when `MG_MAX_RECV_BUF_SIZE` is `4096`. Does `MG_MAX_RECV_BUF_SIZE` set to `4096` pass for you on master?

jameshilliard:
I think this is an improvement for handling smaller `MG_MAX_RECV_BUF_SIZE` sizes, I tested setting `MG_MAX_RECV_BUF_SIZE` to `32768`, which fails on master but works with these changes.

jameshilliard:
> Your change fails unit test if MG_MAX_RECV_BUF_SIZE set to some sane value, e.g. 4096.

Should be fixed now.

jameshilliard:
> Your change fails unit test if MG_MAX_RECV_BUF_SIZE set to some sane value, e.g. 4096.

The test in #1543 reproduces the `MG_MAX_RECV_BUF_SIZE=4096` failure I saw in master, I added the same test to this pull request as well to validate that this fixes the test failure.

test/unit_test.c
```diff
@@ -498,26 +498,55 @@ static void eh1(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
...
+        n = hm->message.len;
+      mg_iobuf_add(fd->buf, fd->buf->len, hm->message.ptr, n, MG_IO_SIZE);
+    }
     fd->code = atoi(hm->uri.ptr);
     fd->closed = 1;
+    fd->done = 1;
     c->is_closing = 1;
+    fd->buf->buf[fd->buf->len] = '\0';
     (void) c;
+  } else if (ev == MG_EV_HTTP_CHUNK) {
```
jameshilliard, line 522: @cpq I modified the fetch callback to support `MG_EV_HTTP_CHUNK` so that it's able to process downloads larger than `MG_MAX_RECV_BUF_SIZE`, this fixes the failure in #1543.

src/sock.c
```diff
@@ -279,11 +279,16 @@ static long mg_sock_recv(struct mg_connection *c, void *buf, size_t len) {
 static long read_conn(struct mg_connection *c) {
   long n = -1;
   if (c->recv.len >= MG_MAX_RECV_BUF_SIZE) {
+#if MG_ENABLE_RECV_RETRY
+    return n;
+#else
     mg_error(c, "max_recv_buf_size reached");
+#endif
```
jameshilliard, line 286: Behavior when reaching `MG_MAX_RECV_BUF_SIZE` is now compile time configurable.

cpq:
So depending on the MG_ENABLE_RECV_RETRY, the behavior and meaning of the MG_MAX_RECV_BUF_SIZE is different.

I suggest to use a different constant for shaping:
- MG_MAX_RECV_BUF_SIZE - will continue to have its existing semantics and logic
- MG_STREAM_RECV_SIZE - a new constant used specifically for traffic shaping, with the semantics you're asking for

Does that make sense?

cpq:
Let's introduce MG_STREAM_RECV_SIZE, set it by default to MG_MAX_RECV_SIZE.

Also, if we stop receiving when MG_MAX_RECV_SIZE is reached, the `mg_mgr_poll()` might fall into a busy loop, because we're still select-ing/poll-ing on that connection, although not reading.. 

Can you please make that change? Your PR will be integrated then.

jameshilliard:
> So depending on the MG_ENABLE_RECV_RETRY, the behavior and meaning of the MG_MAX_RECV_BUF_SIZE is different.

Well it's really more just changing the behavior upon reaching the max buf size, the buf size limit is respected in both cases.

> I suggest to use a different constant for shaping:
> 
>     * MG_MAX_RECV_BUF_SIZE - will continue to have its existing semantics and logic
> 
>     * MG_STREAM_RECV_SIZE - a new constant used specifically for traffic shaping, with the semantics you're asking for
> 
> 
> Does that make sense?

It's somewhat unclear how the precedence would work here. Does `MG_STREAM_RECV_SIZE` take precedence over `MG_MAX_RECV_SIZE` if set?

> Let's introduce MG_STREAM_RECV_SIZE, set it by default to MG_MAX_RECV_SIZE.

If `MG_STREAM_RECV_SIZE` has a default of `MG_MAX_RECV_SIZE` that would mean the new behavior is the default? I'm a little confused here.

> Also, if we stop receiving when MG_MAX_RECV_SIZE is reached, the mg_mgr_poll() might fall into a busy loop, because we're still select-ing/poll-ing on that connection, although not reading..

Well the use case is for situations where there's buffer consumer attached to the poll events, we also use a timer to kill stalled uploads so that we don't get stuck in a busy loop forever.

cpq:
MG_MAX_RECV_SIZE is supposed to be some high value, like 5mb. That is always the highest priority. If a recv buffer gets this big and user handler not yet acted upon it, just kill the connection.

MG_MAX_STREAM_SIZE  by default is set to MG_MAX_RECV_SIZE. Application that want to cap the receving traffic,  set MG_MAX_STREAM_SIZE to a much lower value, e.g. 4kb, and activate the "streaming" logic. In this scenario, recv buffer will never grow up to MG_MAX_RECV_SIZE.

jameshilliard:
> MG_MAX_RECV_SIZE is supposed to be some high value, like 5mb. That is always the highest priority. If a recv buffer gets this big and user handler not yet acted upon it, a connection gets forcefully closed.

This seems a bit confusing, I think I have a better solution here that gives runtime behavior control to application logic when needed.

I reworked things so that a `MG_EV_BUFFER_FULL` event is emitted upon the buffer reaching `MG_MAX_RECV_SIZE` which allows the event handler to inhibit the error while still retaining default behavior, this is useful for cases where it may be desirable to dynamically disable the `max_recv_buf_size reached` error depending upon user `fn_data` state.

This is handy since it allow us to have application checks within a `MG_EV_BUFFER_FULL` handler to confirm that we are still waiting on our buffer consumer and that we're fine to continue looping.

> By the way, MG_MAX_RECV_SIZE was introduced in the first place to avoid making timers that kill stalled connections.

Oh, I thought it was to prevent memory from usage from getting too high. It doesn't really look like it will reliably kill stalled connections in all cases to me, only stalled connections that keep pushing data filling up the buffer from what I can tell.

> MG_STREAM_RECV_SIZE by default is set to MG_MAX_RECV_SIZE. Application that want to cap the recv traffic, set MG_STREAM_RECV_SIZE to a much lower value, e.g. 4kb, and activate the "streaming" logic. In this scenario, recv buffer never grows beyond MG_STREAM_RECV_SIZE and obviously never reaches MG_MAX_RECV_SIZE.

Well the thing we need is the ability to put back-pressure on the sender once the buffer hits the size limit as we proxy to a unix socket that may be quite slow at receiving traffic(may be streaming a very large update image to very slow flash memory that far exceeds ram size for example).

The lower e.g. 4kb size in the tests is just to ensure the chunking handler code works correctly with a `MG_MAX_RECV_SIZE` smaller than the size of the data being received.

In practice the default `MG_MAX_RECV_SIZE` doesn't appear to need to be different for streaming cases, it's the behavior once reaching `MG_MAX_RECV_SIZE` that needs to be controllable.

cpq:
Why a new event, MG_EV_BUFFER_FULL ?
An app can catch MG_EV_READ and compare c->recv.len with whatever threshold it want.
If an app wants to stall reads, we could introduce a user-settable flag, e.g. c->is_full, in which case sock.c won't read more data.

cpq:
So an event handler is something like:

```c
  if (ev == MV_EV_READ) {
     if (c->recv.len > MAX) c->is_full = true;
  }
  ...
  if (c->recv.len < MAX) c->is_full = false;
```


Try it out:

```diff
diff --git a/src/net.h b/src/net.h
index 044c7350..f6c17083 100644
--- a/src/net.h
+++ b/src/net.h
@@ -63,6 +63,7 @@ struct mg_connection {
   unsigned is_hexdumping : 1;  // Hexdump in/out traffic
   unsigned is_draining : 1;    // Send remaining data, then close and free
   unsigned is_closing : 1;     // Close and free the connection immediately
+  unsigned is_full : 1;        // Stop reads, until cleared
   unsigned is_readable : 1;    // Connection is ready to read
   unsigned is_writable : 1;    // Connection is ready to write
 };
diff --git a/src/sock.c b/src/sock.c
index 26a37edb..4ea9e6e1 100644
--- a/src/sock.c
+++ b/src/sock.c
@@ -469,14 +469,27 @@ int mg_mkpipe(struct mg_mgr *mgr, mg_event_handler_t fn, void *fn_data,
   return (int) sp[0];
 }
 
+static bool can_read(const struct mg_connection *c) {
+  return c->is_full == false;
+}
+
+static bool can_write(const struct mg_connection *c) {
+  return c->is_connecting || (c->send.len > 0 && c->is_tls_hs == 0);
+}
+
+static bool skip_iotest(const struct mg_connection *c) {
+  return (c->is_closing || c->is_resolving || FD(c) == INVALID_SOCKET) ||
+         (can_read(c) == false && can_write(c) == false);
+}
+
 static void mg_iotest(struct mg_mgr *mgr, int ms) {
 #if MG_ARCH == MG_ARCH_FREERTOS_TCP
   struct mg_connection *c;
   for (c = mgr->conns; c != NULL; c = c->next) {
-    if (c->is_closing || c->is_resolving || FD(c) == INVALID_SOCKET) continue;
-    FreeRTOS_FD_SET(c->fd, mgr->ss, eSELECT_READ | eSELECT_EXCEPT);
-    if (c->is_connecting || (c->send.len > 0 && c->is_tls_hs == 0))
-      FreeRTOS_FD_SET(c->fd, mgr->ss, eSELECT_WRITE);
+    if (skip_iotest(c)) continue;
+    if (can_read(c))
+      FreeRTOS_FD_SET(c->fd, mgr->ss, eSELECT_READ | eSELECT_EXCEPT);
+    if (can_write(c)) FreeRTOS_FD_SET(c->fd, mgr->ss, eSELECT_WRITE);
   }
   FreeRTOS_select(mgr->ss, pdMS_TO_TICKS(ms));
   for (c = mgr->conns; c != NULL; c = c->next) {
@@ -494,14 +507,12 @@ static void mg_iotest(struct mg_mgr *mgr, int ms) {
   memset(fds, 0, sizeof(fds));
   n = 0;
   for (struct mg_connection *c = mgr->conns; c != NULL; c = c->next) {
-    if (c->is_closing || c->is_resolving || FD(c) == INVALID_SOCKET) {
+    if (skip_iotest(c)) {
       // Socket not valid, ignore
     } else {
       fds[n].fd = FD(c);
-      fds[n].events = POLLIN;
-      if (c->is_connecting || (c->send.len > 0 && c->is_tls_hs == 0)) {
-        fds[n].events |= POLLOUT;
-      }
+      if (can_read(c)) fds[n].events = POLLIN;
+      if (can_write(c)) fds[n].events |= POLLOUT;
       n++;
       if (mg_tls_pending(c) > 0) ms = 0;  // Don't wait if TLS is ready
     }
@@ -512,7 +523,7 @@ static void mg_iotest(struct mg_mgr *mgr, int ms) {
   } else {
     n = 0;
     for (struct mg_connection *c = mgr->conns; c != NULL; c = c->next) {
-      if (c->is_closing || c->is_resolving || FD(c) == INVALID_SOCKET) {
+      if (skip_iotest(c)) {
         // Socket not valid, ignore
       } else {
         c->is_readable = (unsigned) (fds[n].revents & POLLIN ? 1 : 0);
@@ -534,12 +545,11 @@ static void mg_iotest(struct mg_mgr *mgr, int ms) {
   FD_ZERO(&wset);
 
   for (c = mgr->conns; c != NULL; c = c->next) {
-    if (c->is_closing || c->is_resolving || FD(c) == INVALID_SOCKET) continue;
-    FD_SET(FD(c), &rset);
-    if (FD(c) > maxfd) maxfd = FD(c);
-    if (c->is_connecting || (c->send.len > 0 && c->is_tls_hs == 0))
-      FD_SET(FD(c), &wset);
+    if (skip_iotest(c)) continue;
+    if (can_read(c)) FD_SET(FD(c), &rset);
+    if (can_write(c)) FD_SET(FD(c), &wset);
     if (mg_tls_pending(c) > 0) tv = tv_zero;
+    if (FD(c) > maxfd) maxfd = FD(c);
   }
 
   if ((rc = select((int) maxfd + 1, &rset, &wset, NULL, &tv)) < 0) {
```

jameshilliard:
> Why a new event, MG_EV_BUFFER_FULL ?

Seems to be the easiest way to control what happens when the buffer is full.

For example say the app is partially draining/processing the buffer on each poll loop, the app may want to catch this event and use it as an opportunity to check if whatever is draining/processing the buffer has stalled for whatever reason.

> An app can catch MG_EV_READ and compare c->recv.len with whatever threshold it wants.

Not sure how that would help, `MG_EV_READ` is only called when data is read from the socket AFAIU and doesn't appear to allow controlling what happens when the buffer is full.

> If an app wants to stall reads, we could introduce a user-settable flag, e.g. c->is_full, in which case sock.c won't read more data.

Well sock.c already has the `c->recv.len >= MG_MAX_RECV_SIZE` condition check which is equivalent of `c->is_full` condition, the important part is the ability to control what happens when the buffer is full.

cpq:
BTW, we could stall writes the same way. Then user-settable flags are better called something like  `c->is_stalled_reads` and `c->is_stalled_writes`

cpq:
Currently there is no such thing as "buffer is full". We read as far as we can malloc. And we malloc in MG_IO_SIZE chunks. That's all we have.

Catching MG_EV_READ is a reliable way to tap into that process

jameshilliard:
> So an event handler is something like:
> 
> ```c
>   if (ev == MV_EV_READ) {
>      if (c->recv.len > MAX) c->is_full = true;
>   }
>   ...
>   if (c->recv.len < MAX) c->is_full = false;
> ```

I think rather than making it a boolean it might make sense to have a `c->max_recv` instead of the static `MG_MAX_RECV_SIZE` check in `read_conn`(with `c->max_recv` defaulting to `MG_MAX_RECV_SIZE`) in addition to the `MG_EV_BUFFER_FULL` callback, that way the recv limit can be dynamically adjusted and the buffer full condition can be precisely controlled via the event handler(with current effective defaults preserved).

> BTW, we could stall writes the same way. Then user-settable flags are better called something like c->is_stalled_reads and c->is_stalled_writes

I think writes we don't need anything special here since we can already see how many bytes are in the outgoing buffer and can just pause writing to the send buffer if we need to slow things down there.

> Currently there is no such thing as "buffer is full". We read as far as we can malloc. And we malloc in MG_IO_SIZE chunks. That's all we have.

I mean, from looking at the code in `read_conn` it looks like the malloc/buffer size upper limit works out to be `MG_MAX_RECV_SIZE + (MG_MAX_RECV_SIZE % MG_IO_SIZE)` due to always incrementing in `MG_IO_SIZE` chunks, which makes `MG_MAX_RECV_SIZE` effectively a buffer full check under default configuration since the default `(MG_MAX_RECV_SIZE % MG_IO_SIZE) == 0`, unless I'm reading things wrong here.

> Catching MG_EV_READ is a reliable way to tap into that process

I guess it could probably work, but the API semantics seem a little tricky reason about for me as you're then dealing with two buffer limits and it's not obvious what happens if you're trying to make them the same.

Say I do something like this:
 ```c
   if (ev == MV_EV_READ) {
      if (c->recv.len >= MG_MAX_RECV_SIZE ) c->is_full = true;
   }
   ...
   if (c->recv.len < MG_MAX_RECV_SIZE ) c->is_full = false;
 ```

This appears at first glance like it would always inhibit a `max_recv_buf_size reached` error, however if `(MG_MAX_RECV_SIZE % MG_IO_SIZE) != 0` then it's not so clear what happens to me.

> Please, don't try to change MG_MAX_RECV_SIZE semantics. It is clear: as soon we hit that limit, we close, full stop.

I'm mostly a bit concerned with the interaction between `MG_MAX_RECV_SIZE` and `c->is_full` as it seems like it could be prone to subtle bugs with that approach due to effectively trying to do the same check in two different places.

cpq:
Please just forget about MG_MAX_RECV_SIZE.
It is not not "buffer full" constant. It is "close immediately" constant.

W.r.t. making a variable rather than a bool flag - well, it is exactly the same as MG_STREAM_RECV_SIZE suggestion.

jameshilliard:
> Please just forget about MG_MAX_RECV_SIZE.
> It is not not a "buffer full" constant. It is a "close immediately" constant.

So I think part of the issue is that we may actually need entirely different "buffer full" and "close immediately" behavior semantics that is dependent on the URI match.

Consider the [following example](https://github.com/jameshilliard/swupdate/blob/mongoose-v8/mongoose/mongoose_interface.c#L676-L705):
```c
static void ev_handler(struct mg_connection *nc, int ev, void *ev_data, void *fn_data)
{
	if (ev == MG_EV_HTTP_MSG) {
		if (!mg_http_is_authorized(hm, global_auth_domain, global_auth_file))
			mg_http_send_digest_auth_request(nc, global_auth_domain);
		else if (mg_http_get_header(hm, "Sec-WebSocket-Key") != NULL)
			websocket_handler(nc, ev_data);
		else if (mg_http_match_uri(hm, "/restart"))
			restart_handler(nc, ev_data);
		else
			mg_http_serve_dir(nc, ev_data, &s_http_server_opts);
	} else if (nc->label[0] != 'M' && ev == MG_EV_HTTP_CHUNK) {
		if (!mg_http_is_authorized(hm, global_auth_domain, global_auth_file))
			mg_http_send_digest_auth_request(nc, global_auth_domain);
		else if (mg_http_match_uri(hm, "/upload"))
			multipart_upload_handler(nc, ev, ev_data, upload_handler, NULL);
	} else if (nc->label[0] == 'M' && (ev == MG_EV_READ || ev == MG_EV_POLL || ev == MG_EV_CLOSE)) {
		multipart_upload_handler(nc, ev, ev_data, upload_handler, fn_data);
#if MG_ENABLE_SSL
	} else if (ev == MG_EV_ACCEPT && ssl) {
		mg_tls_init(nc, &tls_opts);
#endif
	} else if (ev == MG_EV_ERROR) {
		ERROR("%p %s", nc->fd, (char *) ev_data);
	} else if (ev == MG_EV_WS_MSG) {
		mg_iobuf_del(&nc->recv, 0, nc->recv.len);
	}
}
```

1. We want to "close immediately" upon hitting `MG_MAX_RECV_SIZE` for `mg_http_match_uri(hm, "/restart")` as it is a normal http event handler without chunking/streaming support.
2. Likewise for `mg_http_serve_dir(nc, ev_data, &s_http_server_opts)` we probably want to "close immediately" upon hitting `MG_MAX_RECV_SIZE`.
3. For the `websocket_handler(nc, ev_data)` we also probably want to "close immediately" if hitting `MG_MAX_RECV_SIZE` to prevent excess memory usage.
4. On the other hand we never want to hit a "close immediately" at all and instead want "buffer until full and loop" upon hitting a `MG_MAX_RECV_SIZE` equivalent for `mg_http_match_uri(hm, "/upload")` as it supports streaming uploads that may need to apply back-pressure.

Now in some cases we may want to vary the specific "close immediately" and "buffer until and then loop" limits based on the uri match handler requirements(not super important at the moment here but I could see it being helpful).

Using `c->max_recv` with a `MG_EV_BUFFER_FULL` callback seems to be the cleanest way to provide this flexibility cases as it lets us both decide what our `c->recv` max is and what happens when that max gets hit on a granular per uri/handler basis.

So for a handler like this we could theoretically want up to 4 different limits with 2 types of behavior(close immediately or continue looping) upon reaching the limits.

> W.r.t. making a variable rather than a bool flag - well, it is exactly the same as MG_STREAM_RECV_SIZE suggestion. I agree, having MG_STREAM_RECV_SIZE is more precise in terms of controlling the size. The boolean flag gives more freedom to the app to decide when to stall.

I think the semantics are cleaner with a dynamic `c->max_recv`, at least for more complex multi-handler scenarios.

> This way we always fill up the existing MG_IO_SIZE chunk, and only when it is full, we allocate the next. With this logic, the
c->recv.len < SOME_VALUE check in the event handler will be fairly precise. It will guarantee that the c->recv.size is not larger than the nearest MG_IO_SIZE boundary.

To me something like this in my PR seems clearer as it guarantees `c->recv.size` would not exceed `MG_MAX_RECV_SIZE`:
```c
else if (c->recv.size - c->recv.len < MG_IO_SIZE &&
             c->recv.size + MG_IO_SIZE <= MG_MAX_RECV_SIZE &&
             !mg_iobuf_resize(&c->recv, c->recv.size + MG_IO_SIZE)) {
```

> An advantage of this approach is that it correctly deals with the mg_iotest and does not fall into a busy loop when user stalls reads. If it works for your case, I suggest to go ahead with that.

The main worry with this approach is that it's not clear how one would fully inhibit(on a application logic conditional basis) hitting the `max_recv_buf_size reached` error condition while still allowing the buffer to reach a size of `MG_MAX_RECV_SIZE`.

test/unit_test.c
```diff
@@ -1796,6 +1828,97 @@ static void test_invalid_listen_addr(void) {
...
+    status->received += consume;
+    status->recv_crc = mg_crc32(status->recv_crc, (const char *) c->recv.buf, consume);
+    mg_iobuf_del(&c->recv, 0, consume);
+  }
+
+  // count polls with full buffer
+  if (ev == MG_EV_BUFFER_FULL) {
+    long *error = (long *) ev_data;
+    ASSERT(*error == -1);
+    *error = 0;
```
jameshilliard, line 1875: Making the buffer limit reached behavior controllable like this seems much easier to reason about since it gives clear XOR semantics for the decision tree which can be done in a subhandler rather than having to guess on which limit takes precedence in a particular code path if we have two separate limits. It also makes it trivial to retain default close on limit reached behavior by simply not catching `MG_EV_BUFFER_FULL` events for any particular subhandler.

src/net.c
```diff
@@ -179,6 +179,7 @@ struct mg_connection *mg_connect(struct mg_mgr *mgr, const char *url,
     c->is_client = true;
     c->fd = (void *) (size_t) -1;  // Set to invalid socket
     c->fn_data = fn_data;
+    c->recv_max = MG_MAX_RECV_SIZE;
```
jameshilliard, line 182: This seems to work well since then each subhandler can dynamically modify it as needed.

cpq:
Sorry, but I strongly against the MG_EV_BUFFER_FULL message and using the MG_MAX_RECV_SIZE for anything other than it is used now. It is for putting a cap for RAM usage and killing accidentally stalled connections.
Normally, we should not talk about MG_MAX_RECV_SIZE at all. It is just a safety measure which must not bear any other meaning.

jameshilliard:
> Sorry, but I strongly against the MG_EV_BUFFER_FULL message and using the MG_MAX_RECV_SIZE for anything other than it is used now. It is for putting a cap for RAM usage and killing accidentally stalled connections.

How about two variables, one for each limit and the ability to disable either entirely by setting it to 0?

cpq:
I am OK with the `c->full` approach, because it is easy to understand and it is flexible.

jameshilliard:
> I am OK with the `c->is_full` approach, because it is easy to understand and it is flexible.

How about `c->is_full` mixed with a dynamic `c->recv_max`(same semantics and purpose as `MG_MAX_RECV_SIZE` but adjustable on the fly per connection)? At least by making `c->recv_max` dynamic I can then conditionally bump it to a high enough value to prevent it from conflicting with streaming mode when setting `c->is_full` .

cpq:
I envision something like this:

```c
static void ev_handler(struct mg_connection *nc, int ev, void *ev_data, void *fn_data) {
  if (ev == MG_EV_HTTP_MSG) {
    if (..... non streaming handlers) {
      ...
    } else if (mg_http_match_uri(hm, "/upload")) {
      nc->label[0] = 'M';  // Streaming URI !
      multipart_upload_handler(nc, ev, ev_data, upload_handler, NULL);
    }
  } else if (nc->label[0] == 'M' && (ev == MG_EV_READ || ev == MG_EV_POLL || ev == MG_EV_CLOSE)) {
      multipart_upload_handler(nc, ev, ev_data, upload_handler, fn_data); // Drain streaming connection
      c->is_full = c->recv.len >= 2 * MG_IO_SIZE;  // Cap streaming buf size
  }
}
```

jameshilliard:
> I envision something like this:

Yeah, I think that with #1575 to allow for per handler `MG_MAX_RECV_SIZE` tuning would be fairly workable.

cpq:
Let me stress again - I don't see a reason to change `MG_MAX_RECV_SIZE` behavior. It should be set to some reasonable value, which is equal to e.g. max websocket message size - for example, 16kb. That should be enough. Having a per-connection limit is just an overkill.

jameshilliard:
> Let me stress again - I don't see a reason to change `MG_MAX_RECV_SIZE` behavior. It should be set to some reasonable value, which is equal to e.g. max websocket message size - for example, 16kb. That should be enough. Having a per-connection limit is just an overkill.

Tuning the buffer size is important for performance, small buffers can massively degrade performance for uploads especially under high RTT conditions and large buffers may not be desirable for handlers not dealing with upload handlers as you could run into RAM exhaustion issues if you tune them too high for all handlers, also allowing per connecting tuning allows one to limit aggregate recv buffer size in application logic when dealing with multiple connections as a static `MG_MAX_RECV_SIZE` is only able to tune on a per connection basis and would be insufficient to prevent ram exhaustion if there are too many connections.

PR #1484, opened 2022-03-02
Just assert data.txt exists.

jameshilliard:
It seems `ls -l` messes up github actions log output.


jameshilliard:
hmm, must not have tested that right, didn't seem to help

PR #1507, opened 2022-04-02
Handle MBEDTLS_ERR_NET_CONN_RESET errors

jameshilliard:
mbedtls expects this to be returned if the connection resets.

### DIFF
--- a/src/tls_mbed.c
+++ b/src/tls_mbed.c
@@ -24,15 +24,19 @@ void mg_tls_free(struct mg_connection *c) {
 }
 
 bool mg_sock_would_block(void);
+bool mg_sock_conn_reset(void);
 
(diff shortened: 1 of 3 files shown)


src/tls_mbed.c
```diff
@@ -41,8 +51,18 @@ static int mg_net_recv(void *ctx, unsigned char *buf, size_t len) {
   n = (int) recv(fd, buf, len, 0);
   MG_VERBOSE(("%lu n=%d, errno=%d", c->id, n, errno));
   if (n > 0) return n;
-  if (n < 0 && mg_sock_would_block()) return MBEDTLS_ERR_SSL_WANT_READ;
-  return MBEDTLS_ERR_NET_RECV_FAILED;
```
jameshilliard, line 45: A return of 0 is not an error, so this needs to be conditional on `if (n < 0)`.

PR #1518, opened 2022-04-07
Ensure MQTT_CMD_SUBACK is received before calling mg_mqtt_pub in tests

jameshilliard:
Should hopefully fix the random mqtt test failures.

### DIFF
--- a/test/unit_test.c
+++ b/test/unit_test.c
@@ -333,10 +333,25 @@ static void test_sntp(void) {
   ASSERT(mg_sntp_parse(NULL, 0) == -1);
 }
 
+struct mqtt_data {
+  char *buf;
+  int subscribed;
+  int published;
+};
+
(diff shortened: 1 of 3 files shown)


test/unit_test.c
```diff
@@ -363,32 +380,49 @@ static void test_mqtt(void) {
...
   // Set params
+  test_data.subscribed = 0;
+  test_data.published = 0;
   memset(buf, 0, sizeof(buf));
   memset(&opts, 0, sizeof(opts));
   opts.clean = true;
   opts.will_qos = 1;
   opts.will_retain = true;
   opts.keepalive = 20;
-  opts.client_id = mg_str("mg_client");
```
jameshilliard, line 383: Can't hard code the client_id here since tests run against the same public mqtt service in parallel causing breakage when the client_id's conflict.

PR #1543, opened 2022-05-07
Test with 4096 MG_MAX_RECV_BUF_SIZE

jameshilliard:
Expected to fail without #1476.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,11 +9,19 @@ jobs:
         cc: [gcc, clang]
         target: [test, test++, valgrind]
         ssl: [MBEDTLS, OPENSSL]
-    name: linux ${{ matrix.target }} CC=${{ matrix.cc }} SSL=${{ matrix.ssl }}
+        include:
+          - cc: gcc
+            target: test
+            ssl: MBEDTLS
+            extra: -DMG_MAX_RECV_BUF_SIZE=4096
+    name: >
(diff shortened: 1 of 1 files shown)


cpq:
The 4096 ought to fail, because unit test fetched test/data/ca.pem which is ~30kb in size.
When I set MG_MAX_RECV_BUF_SIZE to 35000 or more, tests pass.

jameshilliard:
> The 4096 ought to fail, because unit test fetched test/data/ca.pem which is ~30kb in size.

If you only process the data once fully received/buffered with `MG_EV_HTTP_MSG` yes that is true, but if using `MG_EV_HTTP_CHUNK` or something that drains the buffer when it hits `MG_MAX_RECV_BUF_SIZE` then it should not fail.

cpq:
That's correct.

If we submit your change, then the behavior for use cases that do not process chunks (and that's the majority) will change. When a limit is hit, instead of failiing, the processing will just hang. Is that true?

jameshilliard:
> If we submit your change, then the behavior for use cases that do not process chunks (and that's the majority) will change. 

Yeah, that's true, I made the retry behavior configurable in #1476 with the current behavior the default.

> When a limit is hit, instead of failiing, the processing will just hang. Is that true?

Depends if timeouts are implemented but yeah it would keep retrying.

cpq:
Closing this, as it works as expected.

PR #1561, opened 2022-05-24
Add optional poll() implementation under MG_ENABLE_POLL define

vkoskiv:
This adds the ability to substitute `select()` with `poll()` on supported
unix environments when `MG_ENABLE_POLL` is defined. A make flag called
`USE_POLL` is provided as well.
Using `poll()` removes the limitation of `FD_SETSIZE` concurrent sockets,
generally 1024 on Linux environments.

Tested with `make test IPV6=0 USE_POLL=1`
Marked as `draft` because I have not debugged a strange issue where substituting the call to `malloc()` in `mg_iotest()` with `alloca()` (as suggested by @cpq) results in the `test_http_client` and `test_dns_error` tests to fail.

### DIFF
--- a/src/sock.c
+++ b/src/sock.c
@@ -383,7 +383,7 @@ static void accept_conn(struct mg_mgr *mgr, struct mg_connection *lsn) {
 #endif
       MG_ERROR(("%lu accept failed, errno %d", lsn->id, MG_SOCK_ERRNO));
 #if (MG_ARCH != MG_ARCH_WIN32) && (MG_ARCH != MG_ARCH_FREERTOS_TCP) && \
-    (MG_ARCH != MG_ARCH_TIRTOS)
+    (MG_ARCH != MG_ARCH_TIRTOS) && !(MG_ENABLE_POLL)
   } else if ((long) fd >= FD_SETSIZE) {
     MG_ERROR(("%ld > %ld", (long) fd, (long) FD_SETSIZE));
     closesocket(fd);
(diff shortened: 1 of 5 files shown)


cpq (changes requested):
1. C++ needs some type casts - use `make clean all IPV6=0` to check
2. Make `MG_ENABLE_POLL` like the rest of `MG_ENABLE_*` vars:
   - define if in `src/config.h` as

```c
#ifndef MG_ENABLE_POLL
#define MG_ENABLE_POLL 0
#endif
```

And in the arch_unix.h, 

```c
#if MG_ENABLE_POLL
#include ...
#else
#include ... 
#endif
```

3. No need to path Makefile. You can `make clean test IPV6=0 TFLAGS=-DMG_ENABLE_POLL=1`

src/sock.c
```diff
@@ -475,6 +476,45 @@ static void mg_iotest(struct mg_mgr *mgr, int ms) {
...
+      }
+      i++;
+  }
+
+  if ((rc = poll(fds, n, ms)) < 0) {
+      MG_ERROR(("poll: %d %d", rc, MG_SOCK_ERRNO));
+  }
+
+  i = 0;
+  for (c = mgr->conns; c != NULL; c = c->next) {
```
cpq, line 507: If you stick i++ into the for(), then you don't need to care about increments in the loop body
for (c = mgr->conns; c != NULL; c = c->next, i++)

src/sock.c
```diff
@@ -475,6 +476,45 @@ static void mg_iotest(struct mg_mgr *mgr, int ms) {
     FreeRTOS_FD_CLR(c->fd, mgr->ss,
                     eSELECT_READ | eSELECT_EXCEPT | eSELECT_WRITE);
   }
+#else
+#ifdef MG_ENABLE_POLL
```
cpq, line 480: #if MG_ENABLE_POLL

cpq:
Integrated this change, thank you @vkoskiv

vkoskiv:
Thanks @cpq, I would have preferred to retain authorship of the changes, but at least the issue is now resolved :+1:

vkoskiv:
Alright, thanks for the good feedback! I amended the commit from before with your suggestions. Let me know if you have any more.
Tests pass now, and I also did some stress testing with 9000 concurrent websocket connections, and things worked fine.

PR #1568, opened 2022-05-31
adding new api to enable separation of query string with '&'.

behnejad:
adding mg_delimlist function. like mg_commalist but with delim as last parameter.
this way user can use mg_delimlist to separate query string with '&'.

sorry for opening pull request with out building.
take away lesson for my self: Always build before commit.


cpq:
Please rename `mg_delimlist` to `mg_split`.
Merge all commits into one.

cpq:
Thank you. Please sign a CLA at https://cesanta.com/cla.html , and I'll merge.

PR #1575, opened 2022-06-03
Make MG_MAX_RECV_SIZE dynamically configurable per connection.

jameshilliard:
Useful if different handlers in the same application need a different `MG_MAX_RECV_SIZE`.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -2758,7 +2758,7 @@ static void rx_udp(struct mip_if *ifp, struct pkt *pkt) {
   } else if (c != NULL) {
     c->rem.port = pkt->udp->sport;
     c->rem.ip = pkt->ip->src;
-    if (c->recv.len >= MG_MAX_RECV_SIZE) {
+    if (c->recv.len >= c->recv_max) {
       mg_error(c, "max_recv_buf_size reached");
     } else if (c->recv.size - c->recv.len < pkt->pay.len &&
                !mg_iobuf_resize(&c->recv, c->recv.len + pkt->pay.len)) {
@@ -2827,6 +2827,7 @@ static struct mg_connection *accept_conn(struct mg_connection *lsn,
   c->pfn_data = lsn->pfn_data;
   c->fn = lsn->fn;
   c->fn_data = lsn->fn_data;
+  c->recv_max = MG_MAX_RECV_SIZE;
   mg_call(c, MG_EV_OPEN, NULL);
   mg_call(c, MG_EV_ACCEPT, NULL);
   return c;
@@ -3563,6 +3564,7 @@ struct mg_connection *mg_connect(struct mg_mgr *mgr, const char *url,
     c->is_client = true;
     c->fd = (void *) (size_t) -1;  // Set to invalid socket
     c->fn_data = fn_data;
+    c->recv_max = MG_MAX_RECV_SIZE;
     MG_DEBUG(("%lu -1 %s", c->id, url));
     mg_call(c, MG_EV_OPEN, NULL);
     mg_resolve(c, url);
@@ -3585,6 +3587,7 @@ struct mg_connection *mg_listen(struct mg_mgr *mgr, const char *url,
     LIST_ADD_HEAD(struct mg_connection, &mgr->conns, c);
     c->fn = fn;
     c->fn_data = fn_data;
+    c->recv_max = MG_MAX_RECV_SIZE;
     mg_call(c, MG_EV_OPEN, NULL);
     MG_DEBUG(("%lu %p %s", c->id, c->fd, url));
   }
@@ -3598,6 +3601,7 @@ struct mg_connection *mg_wrapfd(struct mg_mgr *mgr, int fd,
     c->fd = (void *) (size_t) fd;
     c->fn = fn;
     c->fn_data = fn_data;
+    c->recv_max = MG_MAX_RECV_SIZE;
     mg_call(c, MG_EV_OPEN, NULL);
     LIST_ADD_HEAD(struct mg_connection, &mgr->conns, c);
   }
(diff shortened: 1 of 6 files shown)


cpq:
I am not in favor of this change.
Having such a fine granularity does not buy much on an embedded system. Anyone can simply DoS a device with a couple of hanging (half-finished) HTTP requests. Or just few extra users pull the Web UI page - each browser connection opens 4-5 simultaneous connections. So I don't think this change is practical, TBH

cpq:
Let's continue the RAM tuning conversation here. It is an important topic.

Could you share some concrete details - what MCU/board is it, what is free RAM value, and usage pattern?

jameshilliard:
> Anyone can simply DoS a device with a couple of hanging (half-finished) HTTP requests.

Depends on the hardware and use case I think, upload buffers may be desirable to have on the order of multiple megabytes, while other buffers(say for simple get requests) can be far smaller(a few kb).

cpq:
I concur. But let's talk some real numbers. Shall we take some real board and a real use case as an example?

jameshilliard:
> Could you share some concrete details - what MCU/board is it, what is free RAM value, and usage pattern?

Mostly x86_64 industrial and aarch64 boards, RAM typically between 512mb and 4GB, updates may be larger than RAM size and streamed directly to disk if larger than RAM size. Note we restrict the upload handler to a single upload at one time but want a reasonably large buffer for it for performance.

cpq:
Uh. I recently work with devices with 60-120 Kb free RAM, so 512M device, I assume, has megabytes of free RAM - depends on the runtime conditions of course.

In this situation I would not bother at all. Choose some sane limit for the MG_MAX_RECV_SIZE,  e.g. 65kb, and be done with it. Then,
- Using a default MG_IO_MAX of 2048 bytes is fine for that setup
- Every browser session starts 4-5 keep-alive connections
- All GET requests have no body, so they are only headers - usually 1-2 kb max, usually less
- As soon as GET request is done, mongoose deletes it from the recv buffer (see mg_iobuf_del() calls in http.c)
- So I would NOT worry about GET requests at all
- The only trouble are big POST and big WS messages
- The big POSTs you handle with streaming
- The big WS messages, well, cap! The suggested 64k is usually fine. For example, AWS IoT has a 8k cap for MQTT.
- Are we missing something?
- Yes, TLS! that is RAM-hungry. But that is a totally different topic.

cpq:
That's understood, thank you.

Nevertheless, I do not see any need for so highly granular configuration. First, it is an embedded linux, so we're talking about systems which have way more than 50-100 Kb free RAM. Second, it is a server, not client, which could be DoSed simply by some number of extra connections, no matter how much optimized they are. Third, Monogoose buffers are not preallocated. We're talking about limits here. The average case scenario uses very moderate RAM per connection.

What is the point of nano- optimisation if I can simply fire a couple of `nc` commands that saturate a device? Limiting the number of connection is not an option, too. All in all, there is no point in doing such optimisation cause it is unlikely to be triggered in the first place.

jameshilliard:
> Nevertheless, I do not see any need for so highly granular configuration. First, it is an embedded linux, so we're talking about systems which have way more than 50-100 Kb free RAM.

Well there's more RAM overall but also the update image sizes and RAM requirements are typically much higher than non-linux systems, the constraints in play vary a lot and I've seen quite a lot of systems that are pretty much running right up against resource limits(often for dumb reasons but it's still rather common).

> We're talking about limits here. The average case scenario uses very moderate RAM per connection.

Well this is for an updater which needs to handle multiple requests(for update status monitoring) while updating, but we want to loosen the limit for only a single connection while keeping the others reasonably low. I'm more concerned with trying to limit worst case scenario RAM usage than average case usage.

> What is the point of nano- optimisation if I can simply fire a couple of nc commands that saturate a device? Limiting the number of connection is not an option, too. All in all, there is no point in doing such optimisation cause it is unlikely to be triggered in the first place.

It's actually kinda easy to trigger accidentally, whatever is pushing the update package(such as some sort of mass update automation system or application) to mongoose could have a rapidly looping resend bug which causes the update request to be spammed at the webserver repeatedly, if all but the active upload connection has a low buffer size limit it's unlikely the system will run out of ram and crash during the update as the excess connections would get killed fairly quickly by going hitting the limit.

If say we want a 1MB or larger recv buffer for the active upload handler it wouldn't take that many retries being spammed to make the system run out of ram if we can't restrict the larger buffer to only the single active upload handler connection. For systems that don't use streaming updates this is also an issue as they will unpack update artifacts to RAM before writing them so will often be near max ram usage already. So we would only want to grow the upload buffer if no update is already being processed in the background.

cpq:
This change won't go it. It adds complexity and buys nothing.

jameshilliard:
> This change won't go in. It adds complexity and buys nothing.

Yeah, I guess it's probably somewhat overly use case specific.

I've def seen update failures similar to what I described(http buffers getting too big and triggering update failures to OOM crashes) get hit by a production line QA testing automation system being a bit overly aggressive with pushing updates for boards that had been preflashed before assembly but which needed to be updated to the latest firmware version before shipping/final burn in testing

Luckily the systems didn't hard brick since they had enough space to use redundant dual A/B partition rotation but power cycling was need in some cases to get things back into a workable state due to processes crashing.

I guess an alternative approach here is to make sure `MG_MAX_RECV_SIZE` is always set to the max buffer size limit one wants at the start and then in the event handler we'd need something like this for handlers dealing with everything other than the connection we want a large buffer on:
```c
if (c->recv.len >= SMALL_BUFFER_MAX_SIZE && ev == MG_EV_READ)
  mg_error(c, "max_recv_buf_size reached");
```

Main downside is just that we have to catch all `MG_EV_READ` events for which we want to clamp buffers rather than only having to unclamp the buffer explicitly as needed, but it should work I think.

> What is the point of nano- optimisation if I can simply fire a couple of nc commands that saturate a device? Limiting the number of connection is not an option, too.

> First, it is an embedded linux, so we're talking about systems which have way more than 50-100 Kb free RAM. Second, it is a server, not client, which could be DoSed simply by some number of extra connections, no matter how much optimized they are.

For devices with say 50-100 Kb free RAM I was thinking it may make sense to implement a simple buffer pool(and maybe connection pool as well to allow clamping max connections) attached to the `mg_mgr`, it would be a bit of a refactor but I think it would mostly just be a matter of having the `mg_iobuf` allocation/deallocation functions accept a `mg_connection` and do something like:

When allocating iobuffers check and decrement from recv_pool(basically just an integer like type for tracking total buffer allocations):
```c
if (allocation_size <= c->mgr->recv_pool)
  c->mgr->recv_pool -= allocation_size
else
  mg_error(c, "max_recv_pool_size reached");
```

When deallocating iobuffers return the allocations to the pool by incrementing the freed buffer size:
```c
c->mgr->recv_pool += allocation_size
```

PR #1579, opened 2022-06-05
Do not try to read from full connections.

jameshilliard:
Fixes a bug where setting `c->is_full = true` wasn't stopping reads.


cpq:
Integrated with a small change, thank you @jameshilliard !

PR #1586, opened 2022-06-10
fixed typo

scaprile:


scaprile:
Some checks are DNS timeouts, others, like the one on top, I have no clue

PR #1602, opened 2022-06-24
fixed watch() not being called after login

scaprile:


cpq:
I think there is a better way to do it: specifically, having only one place where watch() is called.
This place is the login() function:

```javascript
  const login = function(u) {
    document.cookie = `access_token=${u.token};path=/;max-age=3600`;
    setUser(u.user);
    return getconfig();
  };
```

There, after setUser(), we should call watch().
And remove watch() calls from any other place.
I think any codepath for the logged user should go through that point, and watch() will be called.

scaprile:
Got it

cpq:
Looks good!
Please rebase.. It looks like my recent commit made this conflict.

PR #1603, opened 2022-06-25
Use alloca instead of VLA

ryuukk:
Some compilers have problem dealing with VLAs ex: https://issues.dlang.org/show_bug.cgi?id=23213

Ideally a macro to check the feature should be used: https://port70.net/~nsz/c/c11/n1570.html#6.10.8.3

But i think one way of doing it is better for readability

### DIFF
--- a/src/sock.c
+++ b/src/sock.c
@@ -503,9 +503,11 @@ static void mg_iotest(struct mg_mgr *mgr, int ms) {
 #elif MG_ENABLE_POLL
   nfds_t n = 0;
   for (struct mg_connection *c = mgr->conns; c != NULL; c = c->next) n++;
-  struct pollfd fds[n == 0 ? 1 : n];  // Avoid zero-length VLA
+  int pollCount = n == 0 ? 1 : 1; // Avoid zero-length VLA
+  int pollSize = pollSize * sizeof(struct pollfd);
+  struct pollfd *fds = alloca(pollSize);
 
-  memset(fds, 0, sizeof(fds));
+  memset(fds, 0, pollSize);
   n = 0;
   for (struct mg_connection *c = mgr->conns; c != NULL; c = c->next) {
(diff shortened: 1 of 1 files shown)


src/sock.c
```diff
@@ -503,9 +503,11 @@ static void mg_iotest(struct mg_mgr *mgr, int ms) {
 #elif MG_ENABLE_POLL
   nfds_t n = 0;
   for (struct mg_connection *c = mgr->conns; c != NULL; c = c->next) n++;
-  struct pollfd fds[n == 0 ? 1 : n];  // Avoid zero-length VLA
+  int pollCount = n == 0 ? 1 : 1; // Avoid zero-length VLA
```
naydef: Probably it should be `int pollCount = n == 0 ? 1 : n; // Avoid zero-length VLA` maybe

src/sock.c
```diff
@@ -503,8 +503,8 @@ static void mg_iotest(struct mg_mgr *mgr, int ms) {
 #elif MG_ENABLE_POLL
   nfds_t n = 0;
   for (struct mg_connection *c = mgr->conns; c != NULL; c = c->next) n++;
-  int pollCount = n == 0 ? 1 : 1; // Avoid zero-length VLA
-  int pollSize = pollCount * (int) sizeof(struct pollfd);
+  size_t pollCount = n == 0 ? 1 : 1; // Avoid zero-length VLA
```
naydef, line 506: Probably it should be `size_t pollCount = n == 0 ? 1 : n; // Avoid zero-length VLA`

cpq:
Could you squash all commits into a single commit please?

cpq:
Please give a reason why VLA is not good.
Reopen this PR then.

cpq:
@ryuukk thank you. It would have helped to have that explanation in the first place in the PR description.
The recent code is already uses alloca, so I guess the issue is addressed.

PR #1608, opened 2022-06-28
fix incorrect WS packet payload

scaprile:
Only the first command (subscribe) succeeded.
7.7 has this same issue
Regarding failed checks: I've run my debug session, more info via mail


cpq:
Please sync and rebase: there should not be mongoose.{c,h} changes in this commit

scaprile:
That was weird... I never staged mongoose.*

PR #1612, opened 2022-07-01
Exported MQTT result codes

scaprile:
Since already exported MQTT_CMD_* are also not prefixed with MG_, I left them as they are.
Let me know if you want to prefix something.


cpq:
Please "make test" to regenerate mongoose.{c,h} and include them into this PR

cpq:
Hm.. The don't seem to match! mongoose.c/h has a newline before closing curly bracket, mqtt.c/h does not.
Please regenerate. Also, suggest to use clang-format before regeneration.

PR #1615, opened 2022-07-04
util: prefer monotonic clocks for mg_millis

jameshilliard:
Using non-monotonic clocks for elapsed time calculations tends to break in horrible ways on systems without battery backed clocks due to NTP syncing on system startup(CLOCK_REALTIME will effectively jump far into the future so we should avoid it when possible).

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -5386,6 +5386,12 @@ struct mg_str mg_url_pass(const char *url) {
 #endif
 
 
+#if MG_ARCH == MG_ARCH_UNIX && defined(__APPLE__)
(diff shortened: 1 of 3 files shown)


jameshilliard:
> MacOS however seem to support MONOTONIC_RAW.

Except it's not equivalent to the linux version, see #1617.

jameshilliard:
> client_id conflicts should only be present when there is more than one test is running.

Which is pretty much always...github actions runs tests in parallel...and anyone running a local test would also trigger a potential conflict.

PR #1627, opened 2022-07-11
added webUI REST basics example

scaprile:


examples/webui-rest/main.c
```diff
@@ -0,0 +1,52 @@
...
+  if (ev == MG_EV_HTTP_MSG) {
+    struct mg_http_message *hm = (struct mg_http_message *) ev_data;
+    if (mg_http_match_uri(hm, "/api/f1")) {
+      mg_http_reply(c, 200, "Content-Type: application/json\r\n",
+                    "{\"result\": %d}\n", 123);
+    } else if (mg_http_match_uri(hm, "/api/sum")) {
+      // Attempt to fetch a JSON array from the body, hm->body
+      struct mg_str params = mg_str("");
+      double num1, num2;
+      params = mg_str_n(hm->body.ptr, hm->body.len);
```
cpq, line 27: struct mg_str json = hm->body;
mg_json_get_num(json, ....);

cpq:
Very small nitpick

examples/webui-rest/main.c
```diff
@@ -0,0 +1,51 @@
...
+      mg_http_reply(c, 200, "Content-Type: application/json\r\n",
+                    "{\"result\": %d}\n", 123);
+    } else if (mg_http_match_uri(hm, "/api/sum")) {
+      // Attempt to fetch a JSON array from the body, hm->body
+      struct mg_str json = hm->body;
+      double num1, num2;
+      if (mg_json_get_num(json, "$[0]", &num1) &&
+          mg_json_get_num(json, "$[1]", &num2)) {
+        // Success! create a JSON response
+        char *resp = mg_mprintf("{%Q:%g}", "result", num1 + num2);
```
cpq, line 30: That's not needed. We can printf directly to the output buffer, no need to have an intermediate `resp`:

```c
mg_http_reply(c, 200, "...headers...",  "{%Q:%g}\n", "result", num1 + num2);
```

examples/webui-rest/main.c
```diff
@@ -0,0 +1,51 @@
...
+
+static const char *s_http_addr = "http://localhost:8000";  // HTTP port
+static const char *s_root_dir = "web_root";
+
+static void fn(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
+  if (ev == MG_EV_HTTP_MSG) {
+    struct mg_http_message *hm = (struct mg_http_message *) ev_data;
+    if (mg_http_match_uri(hm, "/api/f1")) {
+      mg_http_reply(c, 200, "Content-Type: application/json\r\n",
+                    "{\"result\": %d}\n", 123);
```
cpq, line 22:
```c
mg_http_reply(c, 200, "...headers...",  "{%Q:%d}\n", "result", 123);
```

cpq (changes requested):
Since recently, our mg_*printf* functions can print using extra specifiers, `%Q`, `%M`, `%H`, `%V` - quoted json-compatible string, function, hex, base64.  Let's use these abilities in the examples.

examples/webui-rest/main.c
scaprile: Yes, this one I first though it was nice to show %Q later, but it is better to change it

scaprile:
Since I rebased, the comparison to the previous push shows also some mongoose library changes, but the commit is clean.
I'll check for other examples that could be updated.

PR #1637, opened 2022-07-19
Add data push using REST example

scaprile:


cpq:
Thanks!
Can you elaborate, how is that different from the https://github.com/cesanta/mongoose/tree/master/examples/huge-response , please?

scaprile:
> This example, as far as I see, implements a "many small REST requests" scenario, which is not really a data push.

Nope, it actually implements your scenario 1, though the whole data is sent, not a diff. The client sends its current version and the server either responds with "no changes" or sends the new data. The client only refreshes its UI when new data arrives.

> 2. Send one long "polling" request,

That corresponds to the strategy used in the live-log example, I'm talking about that in the 3rd part of the tutorial (1st is WS, 2nd is scenario 1), where I link to that example and moved the description recently pushed to the http-server tutorial to this new tutorial, as you requested in https://github.com/cesanta/websites/pull/86#issuecomment-1180234608

Though the server code looks pretty similar (the underlying logic is the same but the choices when decision making are the opposite), the client code is totally different. This example actually does what you describe in scenario 1 and requested on the chat:

> 5. Data push. Shows how to implement server-side data push using various techniques: [...], and b) using polling - periodic REST function call that returns a "version", when "version" changes, a UI refreshes state

cpq:
Thanks! Then, please change server code comments. They still talk about the "large response" example

scaprile:
Oops, sorry about that, completely missed the header comments. Fixed.

PR #1644, opened 2022-07-25
Fix Makefile

scaprile:


examples/esp32/uart-bridge/Makefile
```diff
@@ -11,9 +11,10 @@ example:
 build: Makefile $(wildcard main/*)
 	docker run --rm $(DA) -v $(MNT):$(MNT) -w $(CWD) espressif/idf idf.py $(CMD)
 
-flash: build
-flash: CMD = flash monitor
+flash:
+flash: CMD = flash
 flash: DA = --device $(PORT)
+flash: build
```
cpq, line 18: Does this PR simply remove `monitor` target from CMD? Then is should be a one-line change.

cpq (changes requested):

examples/esp32/uart-bridge/Makefile
scaprile: No, I had to move `build` from the top to the bottom line, because it was calling `build` without the new parameters and not flashing. Then I added a "no dependencies" line on top, because sometimes it just stopped telling there was no action needed to be done.

Regarding `monitor`, it was failing with:
> Executing action: monitor
Detecting chip type... ESP32
--- idf_monitor on /dev/ttyUSB0 115200 ---
self.old = termios.tcgetattr(self.fd)
termios.error: (25, 'Inappropriate ioctl for device')

So I removed it since I can run _minicom_.
I first thought the example was using this UART but it doesn't, maybe it is something related to my version of Docker or USB support.

PR #1657, opened 2022-08-03
Resolve the bug that repeatedly creates and overwrites existing sockets using mg_wrapfd().

tilongzs:
When acting as a TCP(SSL) client, such as the following example code, the existing socket is repeatedly created and overwritten.

```
// Defined MG_ENABLE_OPENSSL

#include <iostream>
#include <thread>
#include "mongoose.h"

using namespace std;

static void cfn(struct mg_connection* c, int ev, void* ev_data, void* fn_data) {

	if (ev == MG_EV_OPEN) {
		MG_INFO(("CLIENT has been initialized"));
	}
	else if (ev == MG_EV_CONNECT) {
		MG_INFO(("CLIENT connected"));
	}
	else if (ev == MG_EV_READ) {
		struct mg_iobuf* r = &c->recv;
		MG_INFO(("CLIENT got data: %.*s", r->len, r->buf));
	}
	else if (ev == MG_EV_CLOSE) {
		MG_INFO(("CLIENT disconnected"));
	}
	else if (ev == MG_EV_ERROR) {
		MG_INFO(("CLIENT error: %s", (char*)ev_data));
	}
}

void timer_fn(void* arg)
{
	struct mg_connection* conn = (struct mg_connection*)arg;
	const int len = 1024;
	char* msg = new char[len] {0};

	if (mg_send(conn, msg, len))
	{
		MG_INFO(("sent"));
	}
	else
	{
		MG_ERROR(("sent failed"));
	}

	delete[] msg;
}

int main()
{
	mg_mgr* mgr = new mg_mgr;
	mg_mgr_init(mgr);
	thread([&, mgr]
		{
			while (true)
			{
				mg_mgr_poll(mgr, 100);
			}

			mg_mgr_free(mgr);
		}).detach();

	SOCKADDR_IN localAddr;
	localAddr.sin_family = AF_INET;
	localAddr.sin_port = htons(23301);
	inet_pton(AF_INET, "127.0.0.1", &localAddr.sin_addr);

	SOCKET sockfd = socket(AF_INET, SOCK_STREAM, 0);
	if (0 != ::bind(sockfd, (sockaddr*)&localAddr, sizeof(sockaddr)))
	{
		MG_ERROR(("bind failed"));
		return 0;
	}

	struct mg_connection* conn = mg_wrapfd(mgr, sockfd, cfn, nullptr);
	mg_resolve(conn, "tcp://127.0.0.1:23300");

	mg_tls_opts tlsOpts;
	memset(&tlsOpts, 0, sizeof(mg_tls_opts));
	mg_tls_init(conn, &tlsOpts);

	mg_timer_add(mgr, 2000, MG_TIMER_ONCE, timer_fn, conn);

	cin.get();
}
```

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -1,4 +1,4 @@
-// Copyright (c) 2004-2013 Sergey Lyubka
+﻿// Copyright (c) 2004-2013 Sergey Lyubka
 // Copyright (c) 2013-2022 Cesanta Software Limited
 // All rights reserved
 //
@@ -4299,11 +4299,15 @@ static void setsockopts(struct mg_connection *c) {
 }
 
 void mg_connect_resolved(struct mg_connection *c) {
+  c->is_client = true;
   // char buf[40];
   int type = c->is_udp ? SOCK_DGRAM : SOCK_STREAM;
   int rc, af = c->rem.is_ip6 ? AF_INET6 : AF_INET;
   // mg_straddr(&c->rem, buf, sizeof(buf));
-  c->fd = S2PTR(socket(af, type, 0));
+  if (!c->fd || INVALID_SOCKET == FD(c))
+  {
+    c->fd = S2PTR(socket(af, type, 0));
+  }
   c->is_resolving = 0;
   if (FD(c) == INVALID_SOCKET) {
     mg_error(c, "socket(): %d", MG_SOCK_ERRNO);


cpq:
@tilongzs thank you.
To be honest, I don't understand what the test program demonstrates.
Running on Mac gives:

```
~/src/mongoose$ c++ -std=c++20 /tmp/a.c mongoose.c -I. -DMG_ENABLE_OPENSSL -lssl -lcrypto -I /opt/homebrew/opt/openssl/include/ -L /opt/homebrew/opt/openssl/lib/ -o /tmp/a && /tmp/a
clang: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
clang: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
f5f8d2 2 a.c:12:cfn                     CLIENT has been initialized
f5f937 1 mongoose.c:404:mg_error        1 0x4 socket error
f5f937 2 a.c:21:cfn                     CLIENT error: socket error
f5f937 2 a.c:19:cfn                     CLIENT disconnected
f6011a 2 a.c:31:timer_fn                sent
```

So ..what ?
Please be more elaborate and explain, what is the intention of the client program, what's the expected result, and how your fix actually works.

tilongzs:
@cpq Sorry for not describing it clearly.
The above TCP client code, uses local port 23301 to connect to TCP server port 23300 and sends 1024 bytes of data after two seconds. On the TCP server side you will find that the actual connection is not to port 23301, because in the `mg_connect_resolved()` function, the created socket is overwritten by this line of code `c->fd = S2PTR(socket(af, type, 0));`.
Also, although mg_send() returns success, the data is not actually sent out. You will find that the TCP server did not receive it.
The code I submitted fixes these.

scaprile:
Can you please clarify the _intention_ in human terms of your program ? I mean, what is it that you are trying to do and why you can't do it with Mongoose function calls ?
> `mg_wrapfd()`: Wrap a given file descriptor fd into a connection, and add that connection to the event manager. An fd descriptor must suport send(), recv(), select() syscalls, and be non-blocking. Mongoose will treat it as a TCP socket. The c->rem and c->loc addresses will be empty.

So, your local address information will be overwritten by this call. I guess the idea of this function call was not to be applied on a bound socket.

`mg_resolve()` is not documented, why are you calling it ? And why on a socket you already bound ? 
Why do you bind the socket and then wrap it, why not just wrap it and `mg_listen()` ? 

> The above TCP client code, uses local port 23301 to connect to TCP server port 23300 and sends 1024 bytes of data after two seconds

What is the server and what is the client ? If I understand correctly, you are reusing a single connection (you need one for the server and one for the client), when you `mg_resolve()` you get a new socket on 23301 and the former one loses its wrapping; you've overwritten your connection. You are sending to 23301 cause that is what you asked for by calling `mg_resolve()`; you detached from your former socket. Of course [mg_send()](https://mongoose.ws/documentation/#mg_send) returns a positive number, it will just 

> `mg_send()`: Append data of size size to the c->send buffer. Return number of bytes appended.
Note: This function does not push data to the network. It only appends data to the output buffer. The data is being sent when mg_mgr_poll() is called. If mg_send() is called multiple times, the output buffer grows.

As long as you have memory, it will buffer your data; the event manager will try to send later and eventually TCP will timeout if there's no one listening.

I already sent you an example with the proper way of having a TCP client and a server together [here](https://github.com/cesanta/mongoose/discussions/1638#discussioncomment-3262049), and I used two connections.
I also commented on threading, and linked to the multi-threading tutorial, where all Mongoose function calls are on the very same thread.

tilongzs:
@scaprile Sorry for not describing it clearly.My native language is not English, and the translation software makes it unclear.
The sample code I provided is just the TCP client software code which uses the bound port 23301 to connect to the server side. The other TCP server software code is not provided and it listens on port 23300.
My intention is that the client uses the binded local port 23301 to connect to the server's port 23300.Instead of using the random local port used by `mg_connect()`.

> Why do you bind the socket and then wrap it, why not just wrap it and mg_listen() ?

I think I am using Mongoose in a wrong way. Let's take it step by step. If my client software uses `mg_listen() `to bind to local port 23301, it will return `struct mg_connection* conn`. so how do I use this `conn `to connect to the server next?

scaprile:
> If my client software uses `mg_listen() `to bind to local port 23301, it will return `struct mg_connection* conn`. so how do I use this `conn `to connect to the server next?

The thing is... if you _listen_, you are a _server_. Clients _connect_ to servers. The TCP example I sent you in the other question you posted, has a client and a server. The server listens, the client connects to the server

Things might perhaps be a little easier if you'd just describe what is it that you are trying to achieve. I find it hard to believe that anything you might think has not already been covered in any example/tutorial.

tilongzs:
The intention is to use the bound socket to connect to the server. 
If only `mg_connect()` is used, then a **random local port** is used.That's why I first create the socket by creating socket(), then bind() the local address (127.0.0.1) and port (**23301**), generate `struct mg_connection*` by `mg_wrapfd()`, and finally `mg_resolve()` to connect to the server (127.0.0.1: **23300**).
If you still don't understand what I need, then move to the discussion forum.

scaprile:
What I think I understand is that (for some reason), _you do need to use a specific source port when initiating a connection_ (like e.g. you are implementing an NFS client).
If that is the case, then create, bind, connect, and then wrap:

```
int client = socket(AF_INET, SOCK_STREAM, 0);
bind(client, (struct sockaddr *) &my_addr, sizeof(struct sockaddr_in))
connect(client, (struct sockaddr *) &his_addr, sizeof(his_addr))
struct mg_connection *c;
c = mg_wrapfd(&mgr, client, cfn, NULL);
```
The wrap will trigger an OPEN event on cfn

This worked for me, connecting to my server and closing: [sourceport.zip](https://github.com/cesanta/mongoose/files/9275675/sourceport.zip)

Perhaps there is a way to do it using Mongoose function calls, but I'm not aware of it.

tilongzs:
Yes.This worked for me too.
But adding openssl is a new problem. For example, insert these three lines of ssl code into your code.


```
        c = mg_wrapfd(&mgr, client, cfn, NULL);

// c->is_client = true;

	mg_tls_opts tlsOpts;
	memset(&tlsOpts, 0, sizeof(mg_tls_opts));
	mg_tls_init(c, &tlsOpts);

	while (mgr.conns != NULL)
		mg_mgr_poll(&mgr, 100);  
```

The server will receive the connection, but will not receive the data. Likewise, the client will not receive data from the server.
This is because `c->is_client` is **not** set to `true`, which in turn causes `mg_tls_handshake()` to not be executed.
`c->is_client = true;` is only executed in `mg_connect()`.

Even if we add `c->is_client = true;` after `mg_wrapfd()`, it will cause `SSL_read()` in `mg_tls_recv()` to block until the data from the server is received first.

scaprile:
So your question is : _How to connect to a server using TLS with a fixed and unique local port_ ?

If you would take the time to read the tutorials and do the examples you would probably find things usually do work when used as explained there.
I just read [the client section of the TLS tutorial](https://mongoose.ws/tutorials/tls/#tls-for-clients) and added what it is done there to the former example

1. TLS needs a certificate and you don't show whether you are providing one or not
2. All examples (and the tutorial) call mg_tls_init() at the connect event. Since we already connected, I called it at the open event.
So:
```
  if (ev == MG_EV_OPEN) {
    c->is_client = 1;            // since you wrapped an already connected socket, Mongoose has no way of knowing it is a client unless you tell it
    struct mg_tls_opts opts = {.ca = "ca.pem"};  // provide a proper certificate file, this is the one included when you clone the repo
    mg_tls_init(c, &opts);
    mg_send(c, "GET / HTTP/1.0\r\n\r\n", 18);
```

`make clean all SSL=MBETTLS` works for me: [sourceportTLS.zip](https://github.com/cesanta/mongoose/files/9276927/sourceportTLS.zip)
(The server IP is httpbin.org's resolved IP)

tilongzs:
> So your question is : How to connect to a server using TLS with a fixed and unique local port ?
Finally, yes.
But there is still the last problem that I have already mentioned. The server side will not receive the data unless the server side sends the data to the client **first**.
[server.zip](https://github.com/cesanta/mongoose/files/9278540/server.zip)

Here is the server-side code I wrote：
```
#include "mongoose.h"

// SERVER event handler
static void cfn(struct mg_connection* c, int ev, void* ev_data, void* fn_data) {
	if (ev == MG_EV_OPEN) {
		MG_INFO(("SERVER has been initialized"));
	}
	if (ev == MG_EV_ACCEPT) {
		MG_INFO(("CLIENT has been connected"));

		struct mg_tls_opts opts;
		memset(&opts, 0, sizeof(mg_tls_opts));
		opts.cert = "server.crt";
		opts.certkey = "server.key";
		mg_tls_init(c, &opts);

//		The server side will not receive the data unless the server side sends the data to the client first.
// 		mg_send(c, "ACCEPT\r\n\r\n", 10);
// 		MG_INFO(("MG_EV_ACCEPT::SERVER sent data"));
	}
	else if (ev == MG_EV_READ) {
		struct mg_iobuf* r = &c->recv;
		MG_INFO(("SERVER got data: %.*s", r->len, r->buf));

		mg_send(c, "HELLO\r\n\r\n", 9);
		MG_INFO(("MG_EV_READ::SERVER sent data"));
	}
	else if (ev == MG_EV_CLOSE) {
		MG_INFO(("SERVER disconnected"));
	}
	else if (ev == MG_EV_ERROR) {
		MG_INFO(("SERVER error: %s", (char*)ev_data));
	}
	(void)fn_data;
}

int main() {
	struct mg_mgr mgr;  // Event manager
	mg_log_set(MG_LL_DEBUG);  // use "3" for 7.7 and earlier
	mg_mgr_init(&mgr);        // Initialize event manager

	const char* url = "tcp://0.0.0.0:443";
	struct mg_connection* c = mg_listen(&mgr, url, cfn, NULL);
	
	while (mgr.conns != NULL)
		mg_mgr_poll(&mgr, 100);  // Infinite event loop, blocks for upto 100ms
								 // unless there is network activity
	mg_mgr_free(&mgr);         // Free resources

	return 0;
}

```

scaprile:
Is your server on the same process as your client ?
We did not configure the socket to be non-blocking, and that is a requirement. The process will block on a client call and the server won't run properly:

> Wrap a given file descriptor fd into a connection, and add that connection to the event manager. An fd descriptor must suport send(), recv(), select() syscalls, and **be non-blocking**. Mongoose will treat it as a TCP socket. The c->rem and c->loc addresses will be empty.

```
  int client = socket(AF_INET, SOCK_STREAM | SOCK_NONBLOCK, 0);
  bind(client, (struct sockaddr *) &my_addr, sizeof(struct sockaddr_in));
  while (connect(client, (struct sockaddr *) &his_addr, sizeof(his_addr)) == -1) ;
  c = mg_wrapfd(&mgr, client, cfn, NULL);
```
Anyway, you won't be able to busy wait on a real situation, so you should probably wrap your connect in a state machine on a timer event handler or use the POLL event on another connection.

Works for me: [clientserver.zip](https://github.com/cesanta/mongoose/files/9284086/clientserver.zip)

tilongzs:
> Is your server on the same process as your client ?
Not on the same process.

There is no `SOCK_NONBLOCK` on windows systems, so I used this way to set the socket to non-blocking mode.
```
int client = socket(AF_INET, SOCK_STREAM, 0);

	unsigned long ul = 1;
	if (SOCKET_ERROR == ioctlsocket(client, FIONBIO, (unsigned long*)&ul)) 
	{
		return 0;
	}
```
[client.zip](https://github.com/cesanta/mongoose/files/9286105/client.zip)

Although the server program shows that it is connected, the client program's` connect()` always returns -1 and` errno ! = EAGAIN`.

tilongzs:
Let's revisit the two changes I originally committed, both in the mg_connect_resolved() function.
One is the addition of c->is_client = true; this avoids the problem of c->is_client being false due to mg_connect() not being called. After all, if mg_connect_resolved() is executed, c must be client, right?
The other one is to do a check if c->fd already exists before c->fd = S2PTR(socket(af, type, 0)); to prevent it from being overwritten accidentally.

Finally you can use the way in the sample code I wrote to connect to the server, although you don't understand it.
```
SOCKET sockfd = socket(AF_INET, SOCK_STREAM, 0);
	if (0 != ::bind(sockfd, (sockaddr*)&localAddr, sizeof(sockaddr)))
	{
		MG_ERROR(("bind failed"));
		return 0;
	}

	struct mg_connection* conn = mg_wrapfd(mgr, sockfd, cfn, nullptr);
	mg_resolve(conn, "tcp://127.0.0.1:23300");

	mg_tls_opts tlsOpts;
	memset(&tlsOpts, 0, sizeof(mg_tls_opts));
	mg_tls_init(conn, &tlsOpts);
```

Sorry for not reading the official tutorial carefully, I always take my code for granted. I will close this commit and just use it as my own.

PR #1683, opened 2022-08-23
Show how to send data to a worker thread

scaprile:
As we've got a question on this

### DIFF
--- a/examples/multi-threaded/main.c
+++ b/examples/multi-threaded/main.c
@@ -24,11 +24,23 @@ static void start_thread(void (*f)(void *), void *p) {
 #endif
 }
 
+struct userdata {
+  int sock;  // Paired socket, worker thread owns it
+  struct mg_str
+      body;  // data to be processed
+};
+
 static void thread_function(void *param) {
-  int sock = (int) (size_t) param;  // Paired socket. We own it
-  sleep(2);                         // Simulate long execution
-  send(sock, "hi", 2, 0);           // Wakeup event manager
-  close(sock);                      // Close the connection
+  struct userdata *p = (struct userdata *) param;
+  sleep(2);  // Simulate long execution
+  // Respond, wakeup event manager
+  if (p->body.len)
+    send(p->sock, p->body.ptr, p->body.len, 0);  // if there is a body, echo it
+  else
+    send(p->sock, "hi", 2, 0);  // otherwise just respond "hi"
+  close(p->sock);               // Close the connection
+  free((void *) p->body.ptr);   // free body
+  free(p);                      // free userdata
 }
 
 static void link_conns(struct mg_connection *c1, struct mg_connection *c2) {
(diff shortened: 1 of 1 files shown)


examples/multi-threaded/main.c
```diff
@@ -61,14 +63,18 @@ static void pcb(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
 static void fn(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
   if (ev == MG_EV_HTTP_MSG) {
     struct mg_http_message *hm = (struct mg_http_message *) ev_data;
+      char param[64];
```
cpq, line 66: Why not passing the whole body, which could be a JSON message?
Like, 

```c
struct mg_str data = hm->body;
if (data.len > 512) data = mg_str("{}");
...
mg_http_reply(...., "%.*s\n", (int) data.len, data.ptr);
```

examples/multi-threaded/main.c
```diff
@@ -61,14 +63,18 @@ static void pcb(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
...
     if (mg_http_match_uri(hm, "/fast")) {
       // Single-threaded code path, for performance comparison
       // The /fast URI responds immediately
-      mg_http_reply(c, 200, "Host: foo.com\r\n", "hi\n");
+      mg_http_reply(c, 200, "Host: foo.com\r\n", param);
     } else {
       // Multithreading code path
       int sock = mg_mkpipe(c->mgr, pcb, c, true);             // Create pipe
       start_thread(thread_function, (void *) (size_t) sock);  // Start thread
+      mg_send(c->fn_data, param, strlen(param));
```
cpq, line 77: I'm not sure this is correct. c->fn_data is NULL, I think.
Do you intend to pass parameters to a thread function?
Then instead of passing a socket to `start_thread()`, we should alloc a struct, copy socket and parameter there, and pass a pointer to a struct as a thread function arg.

examples/multi-threaded/main.c
```diff
@@ -61,14 +63,18 @@ static void pcb(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
...
   if (ev == MG_EV_HTTP_MSG) {
     struct mg_http_message *hm = (struct mg_http_message *) ev_data;
+      char param[64];
+      mg_http_get_var(&hm->query, "param", param, sizeof(param));
+      if (param[0] == '\0') strcpy(param,"hi");
     if (mg_http_match_uri(hm, "/fast")) {
       // Single-threaded code path, for performance comparison
       // The /fast URI responds immediately
-      mg_http_reply(c, 200, "Host: foo.com\r\n", "hi\n");
+      mg_http_reply(c, 200, "Host: foo.com\r\n", param);
```
cpq, line 72: I'd keep the response as "hi\n"

examples/multi-threaded/main.c
scaprile: The OP question was how to access `ev_data` to be able to get the headers, that's why I thought on something that was simply adding `?param=something` to the URL. Yes, passing the body can be easily shown with curl, will do that.

examples/multi-threaded/main.c
scaprile: OK, just thought both "fast" and "threaded" should do the same.

examples/multi-threaded/main.c
cpq: Agree with those points. However I still think we should not do it this way.

Reasoning:
1. The anticipated use case of multi-threaded functionality is a single GET/POST request with some parameters, and a thread function performs long calculations
2. The situation when a client streams extra data , IMO, is significantly more complex, rare, and probably should be handled via websocket anyway. 
3. We should cover (1) in our example
4. The question asked in a discussion, IMO, also talks about situation (1). A thread function just needs to know some request parameters - and I think passing those as a message body is the way to go, it would be just an ordinary REST request which requires a long calculation
5. The examples' pipe event handler is already complex to understand, cause it interconnects a parent and pipe connections. Sending stuff via the c->fn_data is not obvious at all. A far more understood way is what I suggested, IMO: just  calloc thread params, and pass them to a thread function. That would be understood by anyone, and it would be easy to adopt - just change/add your stuff to the thread param struct. Yes it is an extra calloc, but it is simple and obvious.

So I still would argue to avoid using a shady send to c->fn_data.

scaprile: Understood, thanks for the explanation, will do that.

PR #1698, opened 2022-09-01
Fix mg_json_get for the array element result not as expected.

oakfire:
Because **ci** is pre-added, when array ends, it needs to be decremented by 1  before  entering the macro **MG_EOO**.

Before this modification,  finding "$.a[1]"in {"a":[1],"b":[2,3]}" would got the result b[0] .


gvanem:
This broke `test_json()` on Windows at least.
And why isn't this change in `src/json.c` too?

cpq:
@gvanem please pull the recent change, it should be fixed

cpq:
@gvanem please pull the recent change, it should be fixed

PR #1706, opened 2022-09-03
Introduce mg_timer_expired(). Adopt DHCP and link status to use it

cpq:


src/timer.c
```diff
@@ -15,27 +15,26 @@ void mg_timer_free(struct mg_timer **head, struct mg_timer *t) {
   if (*head) *head = t->next;
 }
 
+// t: expiration time, prd: period, now: current time. Return true if expired
+bool mg_timer_expired(uint64_t *t, uint64_t prd, uint64_t now) {
+  if (now + prd < *t) *t = 0;                    // Time wrapped? Reset timer
+  if (*t == 0) *t = now + prd;                   // Firt poll? Set expiration
+  if (*t > now) return false;                    // Not expired yet, return
+  *t = (now - *t) > prd ? now + prd : *t + prd;  // Next expiration time
+  return true;                                   // Expired, return true
```
scaprile, line 24:
```
  if (*t == 0) *t = now;                // First poll? Set expiration
  if (now < *t) t = 0;	                // time probably went backwards (or timer just wrapped around, reset
  if (now - *t < prd) return false;     // Not expired yet, return
  *t = *t + prd;  			// Next expiration time
  return true;                          // Expired, return true
```

scaprile (commented):
If you are OK, I'll pull and change

PR #1707, opened 2022-09-03
Use single-line comments

cpq:


cpq:
I am using `clang-format` utility, which uses a correct style specified in the `.clang-format` file in the repo root.

scaprile:
Yes, I also run it.
However, it didn't turn my comments into single-line comments... maybe my version is too old ? I'll check again. I have an editor window opened on the repo root so it picks up the .clang-format file.

PR #1713, opened 2022-09-07
display file modify date in "%Y/%m/%d %H:%M:%S".

behnejad:
adding compile time option to display file modification date as "%Y/%m/%d %H:%M:%S" in http directory indexing.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -563,12 +563,12 @@ static void printdirentry(const char *name, void *userdata) {
     } else {
       mg_snprintf(sz, sizeof(sz), "%lld", (uint64_t) size);
     }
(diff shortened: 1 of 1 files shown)


src/http.c
```diff
@@ -563,7 +563,19 @@ static void printdirentry(const char *name, void *userdata) {
     } else {
       mg_snprintf(sz, sizeof(sz), "%lld", (uint64_t) size);
     }
+#if defined(MG_HTTP_LOCALTIME)
```
cpq, line 566: Please name it `MG_HTTP_DIRLIST_TIME`

PR #1717, opened 2022-09-07
Fix MIP Q

scaprile:
Available room calculation in `q_write()` was faulty, it would brake as soon as pointers wrapped around and yield a room larger than the queue size. This caused that new incoming frames would overwrite those already in queue, and so `q_read()` would pop gazillion bytes, stepping over the stack and into no-memory land, causing a HardFault.

I used the easiest of the conventional approaches for lock-free ring buffers: wasting one byte

Note: prior pushes not updating mongoose.c sneaked into this commit

### DIFF
--- a/mip/mip.c
+++ b/mip/mip.c
@@ -180,7 +180,7 @@ static void q_copyout(struct queue *q, uint8_t *buf, size_t len, size_t tail) {
 
 static bool q_write(struct queue *q, const void *buf, size_t len) {
   bool success = false;
-  size_t left = q->len - q->head + q->tail;
+  size_t left = (q->len - q->head + q->tail -1) % q->len;
   if (len + sizeof(size_t) <= left) {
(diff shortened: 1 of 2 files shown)


mip/mip.c
```diff
@@ -180,7 +180,7 @@ static void q_copyout(struct queue *q, uint8_t *buf, size_t len, size_t tail) {
 
 static bool q_write(struct queue *q, const void *buf, size_t len) {
   bool success = false;
-  size_t left = q->len - q->head + q->tail;
+  size_t left = (q->len - q->head + q->tail -1) % q->len;
```
cpq, line 183: Can you add a unit test to tests/mip_test.c , that tests queue, that fails without this change , and passes with this change, please?

scaprile: Done, please see #1719

scaprile: Line 26 passes with this patch applied

PR #1718, opened 2022-09-07
Fix float in mg_mprintf

surfdude75:


cpq:
The fix is already pushed!
Thank you.

surfdude75:
Thanks for the merge.

PR #1719, opened 2022-09-07
Make mip_test fail with current queue algo

scaprile:
(used statics 'cause they make it easier to step by step)
line 26 will fail

### DIFF
--- a/test/mip_test.c
+++ b/test/mip_test.c
@@ -2,15 +2,17 @@
 #include "mongoose.c"
 
 static void test_queue(void) {
-  uint8_t buf[sizeof(size_t) + 5];
+  static uint8_t buf[sizeof(size_t) + sizeof(uint16_t) + 3 ]; // fit 1 element but not 2
   uint16_t val = 1234;
-  struct queue q = {buf, sizeof(buf), 0, 0};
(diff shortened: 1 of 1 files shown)


test/mip_test.c
```diff
@@ -2,15 +2,17 @@
 #include "mongoose.c"
 
 static void test_queue(void) {
-  uint8_t buf[sizeof(size_t) + 5];
+  static uint8_t buf[sizeof(size_t) + sizeof(uint16_t) + 3 ]; // fit 1 element but not 2
```
cpq, line 5: why static? we call this function only once, there is no point to memoise it

test/mip_test.c
```diff
@@ -2,15 +2,17 @@
...
   uint16_t val = 1234;
-  struct queue q = {buf, sizeof(buf), 0, 0};
+  static struct queue q = {buf, sizeof(buf), 0, 0};
```
cpq, line 7: likewise, why static?

test/mip_test.c
scaprile: I commented about it above, it makes it easier to step the code, otherwise those are optimized out.
Thought you might like to check.
Can go back to auto variables

PR #1722, opened 2022-09-09
Add mbedtls support to MIP

cpq:


mip/mip.c
```diff
@@ -545,44 +554,110 @@ static struct mg_connection *accept_conn(struct mg_connection *lsn,
...
   } else if (pkt->pay.len == 0) {
-  } else if (c->recv.size - c->recv.len < pkt->pay.len &&
-             !mg_iobuf_resize(&c->recv, c->recv.len + pkt->pay.len)) {
-    mg_error(c, "oom");
+    // TODO(cpq): handle this peer's ACK
   } else if (mg_ntohl(pkt->tcp->seq) != s->ack) {
-    mg_error(c, "oob: %x %x", mg_ntohl(pkt->tcp->seq), s->ack);
+    // TODO(cpq): peer sent us SEQ which we don't expect. Retransmit rather
+    // than close this connection
+    mg_error(c, "SEQ != ACK: %x %x", mg_ntohl(pkt->tcp->seq), s->ack);
```
scaprile, line 603: I'm a bit rusty on this, do we have a sliding window ? We should check the ACK is within our window and selectively retransmit the corresponding segment, and RST if actually OOB (out of the sliding window)

scaprile (commented):
Lots of changes out of my league, anything you want me to check/test ?

.github/workflows/test.yml
```diff
@@ -1,13 +1,13 @@
 name: build
-on: [push, pull_request]
+on: [push, pull_request_target]
```
jameshilliard, line 2: Why was this changed? It makes running tests in pull requests execute against the wrong branch(base branch context instead of pull request branch context).

PR #1726, opened 2022-09-10
Add MAC addr filtering to stm32 driver. Record free space in q profiler

cpq:


mip/mip.c
```diff
@@ -970,14 +972,11 @@ static struct queue qp;
...
-  else if (type == QP_FRAMEDROPPED)
-    ++drop_ctr;  // only incremented on IRQ calls
+  static uint16_t irq_ctr = 0, drop_ctr = 0;
+  struct qpentry e = {.timestamp = (uint32_t) mg_millis(),
+                      .type = (uint16_t) type,
+                      .len = (uint16_t) len};
+  if (type == QP_IRQTRIGGERED) e.len = ++irq_ctr;
   if (ovf) {
     e.type = (uint16_t) QP_QUEUEOVF;
     e.len = drop_ctr;
```
scaprile, line 982: drop_ctr is local and never incremented
(and we can't access ifp->dropped from here)

scaprile (commented):
Spotted a small inconsistency: when there is no room in the profiling queue we always log 0 (didn't actually check, just looked at the code)
I originally thought on post-processing the log, so never thought of counting profiling queue drops.
I used frame length to somehow identify which frame was being processed, logging queue room looks better.

mip/mip.c
```diff
@@ -202,7 +203,11 @@ static bool q_write(struct queue *q, const void *buf, size_t len) {
   return success;
 }
 
-static size_t q_avail(struct queue *q) {
+static inline size_t q_space(struct queue *q) {
+  return q->tail > q->head ? q->tail - q->head : q->tail + (q->len - q->head);
```
scaprile, line 207: This is the "modulo-less" way, but is off by 1
I'm wasting one byte to differentiate full from empty as otherwise both pointers point to the same location either on queue empty or queue full. As we won't reach that condition, it won't print q->len when it is actually full, but it always prints available room +1 . I guess we can live with that...

scaprile (commented):
One more nitpick

PR #1727, opened 2022-09-11
Pass MDC clock divider explicitly rather than guess

cpq:


mip/driver_stm32.h
```diff
@@ -0,0 +1,15 @@
...
+struct mip_driver_stm32 {
+  // MDC clock divider. MDC clock is derived from HCLK, must not exceed 2.5MHz
+  //    HCLK range    DIVIDER    mdc_cr VALUE
+  //    -------------------------------------
+  //    60-100 MHz    HCLK/42        0
+  //    100-150 MHz   HCLK/62        1
+  //    20-35 MHz     HCLK/16        2
+  //    35-60 MHz     HCLK/26        3
+  //    150-216 MHz   HCLK/102       4  <-- value for Nucleo-F* on max speed
+  //    216-310 MHz   HCLK/124       5
```
scaprile, line 12: These values are just probably a coincidence for an MCU with 216 max speed, actual values are taken using a similar approach to what we've taken. (Though they seem to use a fixed guard instead of a percentage)

cpq: The main motivation for this was: if user needs to set a custom value, they would need to read and understand the code of the driver, and what exactly define to set - which is not trivial.
This approach makes it clear, which tunables are there and what are the possible values.

scaprile: Totally agree, and the -1 option is even cooler, great idea !

cpq: Done

mip/driver_stm32.c
```diff
@@ -63,7 +49,62 @@ static void eth_write_phy(uint8_t addr, uint8_t reg, uint32_t val) {
   while (ETH->MACMIIAR & BIT(0)) spin(1);
 }
 
+static uint32_t get_hclk(void) {
+  struct rcc {
+    volatile uint32_t CR, PLLCFGR, CFGR;
+  } *RCC = (struct rcc *) 0x40023800;
+  uint32_t clk = 0, hsi = 16000000 /* 16 MHz */, hse = 8000000 /* 8MHz */;
```
scaprile, line 56: HSI is a fixed value but not common to all MCUs, and the user can trim it.
HSE  depends on user hardware, 8MHz is the one in these Nucleo boards.
We'll need to change HSI for the H7xx, but get_hclk() will also be different. (and the MAC controller is also different)

scaprile (approved):

PR #1728, opened 2022-09-11
Fix leaking non-active (e.g., bound) LWIP TCP PCBs

QRPp:
tcp_close_tcpip() only closed active TCP PCBs, leaking other types of TCP PCBs (notably bound only, not yet connected/used).  With time, this led to DoS situations.

Fix by giving up on the hacky way of searching the LWIP innards for the TCP PCB needing closing.  The searching was non-comprehensive, hence the problem in principle, and inherently racy: could close a meanwhile freed & reused PCB.

Pass the pointer to the struct mg_lwip_conn_state of the struct mg_connection being destroyed to tcp_close_tcpip() instead of the bare PCB pointer, so the PCB pointer stored within can be used.  Delay unassigning that pointer from the struct mg_connection until tcp_close_tcpip() finishes, so that a mg_lwip_tcp_error_cb() meanwhile can clear the PCB pointer stored within. That ensures no double-closing of the TCP PCB when a concurrent TCP error makes LWIP close the TCP PCB from the LWIP side.

NB!  At the moment, PCBs can still leak if tcp_close() fails, which is unlikely but possible as per the API.  Retrying tcp_close() should somehow be implemented.  For now, at least add logging of tcp_close() failures.

Also: remove the extraneous though benign double-calling of tcp_arg(…, NULL).

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -15594,25 +15594,24 @@ int mg_lwip_if_create_conn(struct mg_connection *nc) {
   return 1;
 }
 
-extern struct tcp_pcb *tcp_active_pcbs;
-
 static void tcp_close_tcpip(void *arg) {
   /*
    * A race with tcp error is possible:
    *  - mg_lwip_if_destroy_conn is called, schedules this function.
    *  - RST arrives, mg_lwip_tcp_error_cb is invoked and pcb is destroyed.
    *  - this function runs.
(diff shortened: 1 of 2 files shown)


QRPp:
Erm...  Is anything wrong with the fix or the PR?  Should the issue have been reported first perhaps?

rojer:
@cpq any particular reason this should not be merged into the mongoose-os branch?

PR #1742, opened 2022-09-19
Fuzzing MIP

cpq:


Makefile
```diff
@@ -80,7 +80,7 @@ unpacked:
 	$(CC) -I. mongoose.c test/unit_test.c -o $@
 
 fuzzer: mongoose.c mongoose.h Makefile test/fuzz.c
-	clang++ mongoose.c test/fuzz.c $(WARN) $(INCS) $(TFLAGS) -DMG_ENABLE_LINES -fsanitize=fuzzer,signed-integer-overflow,address -Wno-deprecated -Wno-vla-extension -o $@
+	$(CXX) test/fuzz.c $(WARN) $(INCS) $(TFLAGS) -fsanitize=fuzzer,signed-integer-overflow,address -Wno-deprecated -Wno-vla-extension -o $@
```
scaprile, line 83: Building the fuzzer fails, there are lots of "already defined" errors between fuzz.c and mongoose.o

mip/mip.c
```diff
@@ -313,8 +313,9 @@ static void arp_ask(struct mip_if *ifp, uint32_t ip) {
 
 static void onstatechange(struct mip_if *ifp) {
   if (ifp->state == MIP_STATE_READY) {
+    struct mg_addr addr = {};
```
scaprile, line 316: Looks like some compilers don't like this construction as it seems to be a GNU extension

cpq: Yeah, fixing those ...

Makefile
cpq: Yep, sent an issue to the oss-fuzz project and asked to change.
https://github.com/google/oss-fuzz/issues/8549

PR #1743, opened 2022-09-19
Stm32 nucleo f746z

scaprile:


cpq:
I've invited you to the mongoose-examples org instead - now you can fix any example directly.

cpq:
Is this PR still relevant?

scaprile:
Yes, though looks like I forgot to force push (fixed). This allows to follow the tutorial as git cloning is decoupled from build and flash can be done from this repo. Same as we already did for the F429

PR #1744, opened 2022-09-20
Add smtp client example

cpq:


examples/smtp-client/main.c
```diff
@@ -0,0 +1,80 @@
...
+  uint8_t *state = (uint8_t *) c->label;
+  if (ev == MG_EV_OPEN) {
+    // c->is_hexdumping = 1;
+  } else if (ev == MG_EV_READ) {
+    if (c->recv.len > 0 && c->recv.buf[c->recv.len - 1] == '\n') {
+      MG_INFO(("<-- %.*s", (int) c->recv.len - 2, c->recv.buf));
+      c->recv.len = 0;
+      if (*state == EHLO) {
+        mg_printf(c, "EHLO gmail.com\r\n");
+        *state = c->is_tls ? AUTH : STARTTLS;
```
scaprile, line 25: So the event/state progression is READ/EHLO, READ/STARTTLS, TLS_HS, READ/EHLO, READ/AUTH ?
You send EHLO on connection, then STARTTLS, then send EHLO again when TLS is ready, the server answers and you send EHLO one more time. Is this correct ? Looks like there is one too many.

What if authentication fails or recipient is rejected or whatever ? I know it is an example, perhaps we could just check the server responded OK and nicely exit on error.

Isn't a `switch (*state)` a bit more clear on such a long chain of if/else ?

examples/smtp-client/main.c
```diff
@@ -0,0 +1,80 @@
...
+        c->is_draining = 1;
+        MG_INFO(("end"));
+      }
+      MG_INFO(("--> %.*s", (int) c->send.len - 2, c->send.buf));
+    }
+  } else if (ev == MG_EV_CLOSE) {
+    s_quit = true;
+  } else if (ev == MG_EV_TLS_HS) {
+    MG_INFO(("TLS handshake done!"));
+    mg_printf(c, "EHLO gmail.com\r\n");
```
scaprile, line 68: EHLO needed to trigger a READ, I guess

examples/smtp-client/README.md
```diff
@@ -0,0 +1,5 @@
+# SMTP client example
+
+This example shows how to send emails via Gmail.
+Before running this example, open main.c and modify settings at the top
```
scaprile, line 4: Thought the new OAUTH policy also applied to SMTP. Great if it doesn't

examples/smtp-client/main.c
cpq: Correct. After STARTTLS and TLS upgrade, the whole sequence starts over, with EHLO.

Errors are not handled at all by this client.
I've tried switch/case, but it appeared to me as a more noisy approach

examples/smtp-client/README.md
cpq: At least this code works well.
Also the same Gmail SMTP service is used for subscription to our cloud services, vcon.io and mdash.net - so yeah, it works so far.

Issue #1733, opened 2022-09-18
Memory - corruptions  (OVERRUN) in mg_dtoa()


jameshilliard:
I'm seeing this get flagged by coverity:
https://github.com/cesanta/mongoose/blob/0a265e79a67d7bfcdca27f2ccb98ccb474677ec6/src/fmt.c#L150
```c
Overrunning array "buf" of 40 bytes at byte offset 40 using index "s + n" (which evaluates to 40).
```

scaprile:
Correct, though the user must request a 40+ digit number and it only overwrites already used variable space in the stack
See #1746

PR #1746, opened 2022-09-20
Fix issue #1733

scaprile:


cpq:
This fails unit test - generates warnings, please fix!

scaprile:
> This fails unit test - generates warnings, please fix!

Sorry about that, I tested the issue and the fix on a short piece of code without all the usual -W and it did not warn.

Issue #1737, opened 2022-09-18
Null pointer dereferences  (NULL_RETURNS) in mg_timer_add()


jameshilliard:
I'm seeing this get flagged by coverity:
https://github.com/cesanta/mongoose/blob/0a265e79a67d7bfcdca27f2ccb98ccb474677ec6/src/net.c#L227
```c
Dereferencing a pointer that might be "NULL" "t" when calling "mg_timer_init".
```

cpq:
Closing as fixed

jameshilliard:
I think we might also need #1756 to prevent this from getting flagged.

PR #1756, opened 2022-09-24
Use explicit struct sizeof in calloc in mg_timer_add()

jameshilliard:
Fixes a potential secondary issue in #1737.


cpq:

PR #1757, opened 2022-09-24
Use explicit struct sizeof in calloc in mg_tls_init()

jameshilliard:
This style appears to less likely to trip up static analysis tools.


scaprile:
Yes, but it has some maintenance limitations and in some cases renders the code less clear; there are interesting discussions about this in the usual forums.

jameshilliard:
> Yes, but it has some maintenance limitations and in some cases renders the code less clear; there are interesting discussions about this in the usual forums.

Yeah, it's a little bit more verbose but tradeoff might be worth it to reduce noise from analysis tools, kind of subjective though.

cpq:

PR #1759, opened 2022-09-24
Use explicit struct sizeof in calloc in mg_fs_open()

jameshilliard:
This style appears to less likely to trip up static analysis tools.


cpq:

PR #1766, opened 2022-09-25
Use platform native ntohl/ntohs/htons/htonl when available

### LINKED ISSUE #1738: (TAINTED_SCALAR) in mg_dns_parse()
I'm seeing this get flagged by coverity:
https://github.com/cesanta/mongoose/blob/0a265e79a67d7bfcdca27f2ccb98ccb474677ec6/src/dns.c#L113
```c
Using tainted variable "mg_ntohs(h->num_answers)" as a loop boundary.
```
https://github.com/cesanta/mongoose/blob/0a265e79a67d7bfcdca27f2ccb98ccb474677ec6/src/dns.c#L107
```c
Using tainted variable "mg_ntohs(h->num_questions)" as a loop boundary.
```

jameshilliard:
These functions need endian specific handling, lets use the platform native variants when available to ensure correct behavior.

Should fix #1738 on unix/win32 platforms, others may still have issues.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -5542,16 +5542,24 @@ char *mg_random_str(char *buf, size_t len) {
 }
 
 uint32_t mg_ntohl(uint32_t net) {
+#if MG_ARCH == MG_ARCH_WIN32 || MG_ARCH == MG_ARCH_UNIX
+  return ntohl(net);
+#else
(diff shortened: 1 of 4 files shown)


cpq:
No, we try to use the same codebase everywhere if possible.

PR #1769, opened 2022-09-27
add support encrypted openssl private key

kndozorov:
Add support for encrypted OpenSSL keys stored in file.


cpq:
@kndozorov Why not specifying the password directly? Meaning, new setting should be a password itself, not a file with a password. Some environments may not have a file system. Also, if a password is in the file, an app code can read it and set the option.

cpq:
No response, closing this

PR #1770, opened 2022-09-28
Enable/fix aton test on big endian systems

jameshilliard:
Not entirely sure this is correct but it makes the test pass at least.


cpq:
That's already fixed - thanks!

PR #1771, opened 2022-09-28
src/arch.h: fix endian macro and set default endian

jameshilliard:
This macro which was moved from md5.c/sha1.c doesn't work correctly as it was defining `BIG_ENDIAN` as `__LITTLE_ENDIAN` in some cases.

We should also set the default endian to `LITTLE_ENDIAN` so that any comparisions against `BYTE_ORDER` don't fail in unexpected ways.


cpq:
We won't use that maze of definitions

PR #1775, opened 2022-09-28
Fix endianness issues in unit test

cpq:


src/md5.c
```diff
@@ -1,21 +1,19 @@
...
-    uint32_t t = (uint32_t) ((unsigned) buf[3] << 8 | buf[2]) << 16 |
-                 ((unsigned) buf[1] << 8 | buf[0]);
-    *(uint32_t *) buf = t;
-    buf += 4;
-  } while (--longs);
-#else
-  (void) buf;
-  (void) longs;
-#endif
+  if (MG_BIG_ENDIAN) {
```
jameshilliard, line 7: Wouldn't this be better to use as a normal `#if`, `#else` macro so that it's more clear that it's a compile time conditional?

mongoose.h
```diff
@@ -85,6 +70,9 @@ extern "C" {
 #endif
 #endif  // !defined(MG_ARCH)
 
+// http://esr.ibiblio.org/?p=5095
+#define MG_BIG_ENDIAN (*(uint16_t *)"\0\xff" < 0x100)
```
jameshilliard, line 74: Hmm, there's a potential issue mentioned there:
> I thought the preprocessor would do all these operations, but it turns out this macro does expand to a small anount of code. It’s still pretty neat, though.

src/md5.c
cpq: A compiler optimises that anyways - so it'll be a no-op on a little endian systems.

PR #1777, opened 2022-09-28
Disable broken 32 bit tests and enable 32 bit little endian tests.

jameshilliard:
It appears there's a bunch of endian specific failing 32 bit tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,5 @@
 name: build
-on: [push, pull_request_target]
+on: [push, pull_request]
 jobs:
   linux:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -1,5 +1,5 @@
 name: build
-on: [push, pull_request_target]
+on: [push, pull_request]
```
jameshilliard, line 2: This fixes a [regression from #1722](https://github.com/cesanta/mongoose/pull/1722#discussion_r982966134).

.github/workflows/test.yml
```diff
@@ -19,29 +19,30 @@ jobs:
     - run: sudo apt-get update ; sudo apt-get install libmbedtls-dev
     - run: make ${{ matrix.target }}
   linux_bigendian:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
jameshilliard, line 22: So apparently `ubuntu-latest` is misleading and is actually ubuntu 20.04(which seems to have qemu issues).

jameshilliard:
FYI failing "pull_request_target" tests like "build / linux-bigendian ppc (pull_request_target)" should be safe to ignore as the failure will stop once merged due to "pull_request_target" being removed.

PR #1783, opened 2022-10-03
Update to latest Zephyr and their Docker repo

scaprile:
- Zephyr's new release (new for us, I guess) has different include paths, updated in arch_zephyr.c
- The application starts right away, and as Mongoose requests to listen on a socket while DHCP hasn't resolved an address yet, the socket remains opened on a non-existent address: client requests to plain TCP ports return an ICMP reject; client requests to TLS ports connect TCP and hang at the Client Hello stage; error messages showed in the logs.
  - To solve this, I added the network management and the app waits on a semaphore until a network event releases it (see main.c). Tested on actual hardware.
  - Their C libraries do not seem to provide strdup(), at least their minimal has an open PR to add it... so I updated the arch file to use mg_strdup() while it is not defined
  - Added MG_ARCH_ZEPHYR calling k_uptime_get() into mg_millis(), in order to have proper resolution when available

Moved to their Docker file
- It does not include the OS, only the toolchains. The OS gets cloned/updated from the repo using `west init/update`, handled through the Makefile.
- The Makefile supports building, running, and flashing using the Docker container.
- Default build is for a fake _QEMU_ board, which allows running it on Linux and test right away, requires host networking hocus pocus as we did with the *mip_tap* example. See (README.md)
  - The test has to be re-written, but we can actually test it works instead of just checking it builds; and without an actual board plugged in.

**NOTE:** The repo is initialized on the same directory as the project, so each project would have its own copy. It is large and cloning is time consuming, I would move it one level above and have a common zephyrproject dir for all zephyr-related projects. **What do you think ?** (On further examining their docs, they either put the app folder inside the zephyrproject dir (workspace application) or at a same level dir (freestanding application), so doing this will turn our apps into freestanding applications)

Whatever results from this can be applied to the other zephyr-related examples.


cpq:
W.r.t. common zephyr OS repo for all examples - yeah, I think it makes sense

PR #1787, opened 2022-10-06
Update Zephyr examples

scaprile:
Replicated http-server schema and Makefile with mods to use a single OS repo in top subtree
Renamed time() to ourtime()
Updated prj.conf when necessary
In http-client: removed NULL ptr dereference and added loop exit as probably intended
Added links to generic README file in top subtree


scaprile:
> Looks good! Though, please rename nucleo_* to nucleo- to keep the naming consistent with the rest of the repo.

If you refer to what is in the README, that is the name Zephyr assigns to the board, what I do is pass -b nameoftheboard to their build system. It was hardcoded in the Makefile before, as is qemu_x86 now...

PR #1790, opened 2022-10-07
Add device-dashboard example on Zephyr

scaprile:
Besides the number of sockets and fds, the number of connections has to be increased... but... connections use "contexts" and the number of them has also to be increased...
	CONFIG_NET_MAX_CONN=10
	CONFIG_NET_MAX_CONTEXTS=10
Will write a tutorial...


cpq:
There are failures like this: https://github.com/cesanta/mongoose/actions/runs/3205650827/jobs/5238435497 - are they expected? Looks like Makefile does not have a respective target.

scaprile:
Before I new that I could change the test.yml that is specific for the PR, I changed the global one.
The samples in the repo now do not have a 'zephyr' target in the Makefile, but the modified ones in the other PR do. The Zephyr tests should pass once merged and most current ones fail right now with that error; yes. Sorry about that, now I know I can test the tests along with the example in the same PR.
This add on does not have a test entry yet, I'll add it post merge

PR #1795, opened 2022-10-12
Use board nucleo_f429zi

cpq:


examples/zephyr/http-client/Makefile
```diff
@@ -1,59 +1,40 @@
...
 	true
 
-build:
-	cp $(MONGOOSE_ROOT)/mongoose.[c,h] src/
-	$(DOCKER) $(REPO) /bin/sh -c 'cd $(DOCKER_ZEPHYR_PATH)/zephyr && \
-	west build -b $(BOARD) -p auto $(DOCKER_PROJECT_PATH) $(OVERLAY) --build-dir $(DOCKER_PROJECT_PATH)/build'
+$(ZEPHYR_DIR):
+	mkdir -p $@
+	$(DOCKER) west init $@
+	$(DOCKER) sh -c 'cd $@ && west update'
```
scaprile, line 19: The reason I wrote it the way I did is because:
- west init has to be run only once, it takes forever, and creates the local repo
- west update has to be run after west init the first time, and every other time you want to keep up to date with their repo. It updates from all the subtrees in the Zephyr repo (Espressif, STM, etc) and also takes a fair amount of time even when just updating nothing.

If you want to get rid of the zephyr target, perhaps the best approach, at the cost of always updating the git repo, is to do as in the newer Makefiles cloning from mongoose-examples:

```
$(ZEPHYR_DIR):
	mkdir
	west init

build: | $(ZEPHYR_DIR)
	west update
```
The pipe is supposed to just check if the directory exists so west init will only be executed when the folder is missing

However, with properly written tests, there will be a noticeable speed up if only the first Zephyr example clones the repo and the others just use it and don't try to re-update it.

examples/zephyr/http-client/Makefile
```diff
@@ -1,59 +1,40 @@
...
-example:
+TOP_DIR ?= $(realpath $(CURDIR)/../../..)
+ZEPHYR_DIR ?= $(TOP_DIR)/zephyrproject
+DOCKER ?= docker run --rm \
+          -v $(TOP_DIR):$(TOP_DIR) \
+          -w $(realpath $(CURDIR)) \
+          -e ZEPHYR_BASE=$(ZEPHYR_DIR)/zephyr \
+          $(DOCKER_ARGS) zephyrprojectrtos/ci
+#BUILD_ARGS ?= -b qemu_x86 -DOVERLAY_CONFIG=overlay-e1000.conf
+BUILD_ARGS ?= -b nucleo_f429zi
```
scaprile, line 9: I see you now don't like the board-independent approach, what made you change your mind ?

> Default build is for a fake QEMU board, which allows running it on Linux and test right away, requires host networking hocus pocus as we did with the mip_tap example. See (README.md)

examples/zephyr/http-client/Makefile
```diff
@@ -1,59 +1,40 @@
...
+$(ZEPHYR_DIR):
+	mkdir -p $@
+	$(DOCKER) west init $@
+	$(DOCKER) sh -c 'cd $@ && west update'
 
+build: $(ZEPHYR_DIR)
+	cp $(TOP_DIR)/mongoose.[c,h] src/
+	$(DOCKER) west build $(BUILD_ARGS) -p auto $(realpath $(CURDIR))
+
+run: DOCKER_ARGS ?= --cap-add=NET_ADMIN --device /dev/net/tun --net=host -it
```
scaprile, line 25: This is only needed for the virtual qemu board, that uses the tun/tap interface and shares the host network. But you just removed it

examples/zephyr/http-client/Makefile
```diff
@@ -1,59 +1,40 @@
...
 run:
-	$(DOCKER) -it $(SHARETUN) $(REPO) /bin/sh -c 'cd $(DOCKER_ZEPHYR_PATH) && \
-	west build -t run --build-dir $(DOCKER_PROJECT_PATH)/build'
```
scaprile, line 33: I didn't re-test but at some point I needed to pass the --build-dir argument for the project to be built under the application path and not inside the Zephyr project tree

examples/zephyr/http-client/Makefile
```diff
@@ -1,59 +1,40 @@
...
+	$(DOCKER) west build -t run
 
 clean:
-	$(DOCKER) $(REPO) /bin/sh -c 'cd $(DOCKER_ZEPHYR_PATH) && \
-	west build -t clean --build-dir $(DOCKER_PROJECT_PATH)/build'
+	rm -rf $(TOP_DIR)/.west $(TOP_DIR)/zephyrproject build
```
scaprile, line 30: They have their own partial cleaning (_clean_, keeping board config) and full cleaning (_pristine_, erasing everything)
If you go `rm` that doesn't work on some systems (like mine...) where Docker runs as root, as you don't own those Docker generated files and need to sudo to rm them. The clean operation through Docker is to keep everything the Zephyr way, so Zephyr developers are happy. There is a noticeable speedup when you clean and recompile vs when you pristine and recompile

examples/zephyr/http-client/Makefile
```diff
@@ -1,59 +1,40 @@
...
 
 pristine:
-	$(DOCKER) $(REPO) /bin/sh -c 'cd $(DOCKER_ZEPHYR_PATH) && \
-	west build -t pristine --build-dir $(DOCKER_PROJECT_PATH)/build'
-
+	$(DOCKER) west build -t pristine
```
scaprile, line 33: I didn't re-test, but at some point all these commands must be called from the proper Zephyr directory and giving a proper project path

examples/zephyr/http-client/Makefile
```diff
@@ -1,59 +1,40 @@
...
-
-SHAREUSB ?= --privileged -v /dev/bus/usb/:/dev/bus/usb
-SHARETUN ?= --cap-add=NET_ADMIN --device /dev/net/tun --net=host
-DOCKER ?= docker run --rm -v $(PROJECT_PATH):$(DOCKER_PROJECT_PATH) -v $(ZEPHYR_PATH):$(DOCKER_ZEPHYR_PATH)
-REPO ?= zephyrprojectrtos/ci
-
-
-example:
+TOP_DIR ?= $(realpath $(CURDIR)/../../..)
+ZEPHYR_DIR ?= $(TOP_DIR)/zephyrproject
```
scaprile, line 2: The idea was to have zephyrproject under the examples/zephyr level, parallel to the apps directories. This is what Zephyr calls "freestanding application" so Zephyr developers are happy (they seem to have their own jargon...)

examples/zephyr/http-client/Makefile
```diff
@@ -1,59 +1,40 @@
...
-	$(DOCKER) $(REPO) /bin/sh -c 'cd $(DOCKER_ZEPHYR_PATH) && \
-	west build -t pristine --build-dir $(DOCKER_PROJECT_PATH)/build'
-
+	$(DOCKER) west build -t pristine
 
+flash debug: DOCKER_ARGS ?= --privileged -v /dev/bus/usb/:/dev/bus/usb -it
 flash:
-	$(DOCKER) $(SHAREUSB) $(REPO) /bin/sh -c 'cd $(DOCKER_ZEPHYR_PATH) && \
-	west flash --build-dir $(DOCKER_PROJECT_PATH)/build'
+	$(DOCKER) west flash --build-dir $(realpath $(CURDIR))/build
```
scaprile, line 37: I didn't know you can do that double target variable assignment magic
At this point I'm lost between paths, looks like a crossroads... that is why I used variable names

examples/zephyr/http-client/Makefile
```diff
@@ -1,59 +1,40 @@
...
-
-.PHONY: build flash zephyr clean pristine run
-
-zephyr:
-ifeq ($(wildcard $(ZEPHYR_PATH)/.*),)
-	mkdir $(ZEPHYR_PATH)
-	$(DOCKER) $(REPO) /bin/sh -c 'cd $(DOCKER_PROJECT_DIR) && west init ./$(ZEPHYR_DIR)'
-endif
-	$(DOCKER) $(REPO) /bin/sh -c 'cd $(DOCKER_ZEPHYR_PATH) && west update'
+	$(DOCKER) west debug --build-dir build
```
scaprile, line 40: At some point those commands I needed to call from within a Zephyr project directory (so the `cd`) and pass the target build path.

examples/zephyr/http-client/Makefile
```diff
@@ -1,59 +1,40 @@
...
 
+build: $(ZEPHYR_DIR)
+	cp $(TOP_DIR)/mongoose.[c,h] src/
+	$(DOCKER) west build $(BUILD_ARGS) -p auto $(realpath $(CURDIR))
+
+run: DOCKER_ARGS ?= --cap-add=NET_ADMIN --device /dev/net/tun --net=host -it
 run:
-	$(DOCKER) -it $(SHARETUN) $(REPO) /bin/sh -c 'cd $(DOCKER_ZEPHYR_PATH) && \
-	west build -t run --build-dir $(DOCKER_PROJECT_PATH)/build'
+	$(DOCKER) west build -t run
```
scaprile, line 27: This won't do anything on an F429.
However, on a qemu board, if we can setup the tun/tap in the hosting Ubuntu at Github, qemu is built into the Zephyr Docker image, so we can run on an emulated board and actually test the example works without having to have an actual board connected to the Internet and flash it and etc.
Actually, this fake board was great for spotting the null ptr dereferences in http-client I fixed later, the 386 MMU raised an exception and QEMU showed a nice crash report that led me to the problem. You can also debug on this virtual board running a gdb server in QEMU and a gdb client on another instance of the Docker image (to use the same gdb version)

examples/zephyr/http-client/Makefile
cpq: The intention is to run it against some embedded target - f429 is just a random pick.
I don't see much value in testing some generic i386 target, we have plenty of such tests already.

examples/zephyr/http-client/Makefile
cpq: I believe this approach implements this idea. The `zephyrproject` dir is created in the mongoose repo root, and can be reused by other examples.

I think running `west update` on every build is a overkill; it executes extremely slowly on my machine (takes minutes)

examples/zephyr/http-client/Makefile
cpq: Yeah, main thing for us is to test the embedded build - we're not much interested in running the x86 emulator.

examples/zephyr/http-client/Makefile
cpq: I think using "BUILD_ARGS" we still can build / test with qemu target?

examples/zephyr/http-client/Makefile
cpq: Yeah I overlooked this... Added west pristine and west clean. I think since docker is run with ZEPHYR_BASE environment set, we're good and don't need to `cd ...`

examples/zephyr/http-client/Makefile
scaprile: Yes, I thought on integrating those tests on the Github pipeline, we can add that in the make command I guess

examples/zephyr/http-client/Makefile
cpq: Restored clean and pristine targets. Added `purge` target which deletes everything

examples/zephyr/http-client/Makefile
cpq: Works OK for me

examples/zephyr/http-client/Makefile
cpq: docker is started with CWD equal to the project's dir, and ZEPHYR_BASE set. As far as I know, `build` dir in the CWD is the default. So  I don't think we should be setting anything.

examples/zephyr/http-client/Makefile
```diff
@@ -1,59 +1,39 @@
...
-	$(DOCKER) $(REPO) /bin/sh -c 'cd $(DOCKER_ZEPHYR_PATH) && \
-	west build -t clean --build-dir $(DOCKER_PROJECT_PATH)/build'
-
-pristine:
-	$(DOCKER) $(REPO) /bin/sh -c 'cd $(DOCKER_ZEPHYR_PATH) && \
-	west build -t pristine --build-dir $(DOCKER_PROJECT_PATH)/build'
-
+	-$(DOCKER) west build -t clean
+	-$(DOCKER) west build -t pristine
+	rm -rf $(TOP_DIR)/.west $(ZEPHYR_DIR) build
```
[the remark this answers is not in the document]
cpq: Added purge, left clean and pristine to what west does.

examples/zephyr/http-client/Makefile
scaprile: Yeah, fetched and checked out your z repo and it works for me too, great

PR #1805, opened 2022-10-19
Fix unintended socket close on systems with valid fd=0 descriptor numbering

scaprile:
Scenario:
- When a DNS request times out, mg_error() is called
- mg_error() marks the connection as is_closing, but c->fd=0
- The event manager checks for INVALID_SOCKET, but c-> fd has a valid socket number.

On Unix systems, 0 corresponds to stdin and this goes unnoticed. On lwIP, 0 is the first assigned number which corresponds to the listening socket in many examples. If the example has outside connections resolving IP addresses, and there is a DNS timeout, the listening socket is closed.

Action:
- mg_connect marks socket as invalid before starting the resolution process
  - c->fd = S2PTR(INVALID_SOCKET);

For the sake of keeping things the same, mg_wrapfd is modified to also use S2PTR

Side effect:
- sock.c needs to be included in mongoose.c before net.c; so the Makefile has been modified as a Q&D workaround.


Makefile
```diff
@@ -141,7 +141,9 @@ uninstall:
 	rm -rf $(DESTDIR)$(PREFIX)/lib/libmongoose.a $(DESTDIR)$(PREFIX)/lib/libmongoose.so.$(VERSION) $(DESTDIR)$(PREFIX)/include/mongoose.h $(DESTDIR)$(PREFIX)/lib/libmongoose.so
 
 mongoose.c: Makefile $(wildcard src/*) $(wildcard mip/*.c)
-	(cat src/license.h; echo; echo '#include "mongoose.h"' ; (for F in src/*.c mip/*.c ; do echo; echo '#ifdef MG_ENABLE_LINES'; echo "#line 1 \"$$F\""; echo '#endif'; cat $$F | sed -e 's,#include ".*,,'; done))> $@
+	(cat src/license.h; echo; echo '#include "mongoose.h"' ; \
+	echo; echo '#ifdef MG_ENABLE_LINES'; echo "#line 1 \"src/sock.c\""; echo '#endif'; cat src/sock.c | sed -e 's,#include ".*,,'; \
+	(for F in src/*.c mip/*.c ; do [ $$F != "src/sock.c" ] && (echo; echo '#ifdef MG_ENABLE_LINES'; echo "#line 1 \"$$F\""; echo '#endif'; cat $$F | sed -e 's,#include ".*,,'); done))> $@
```
cpq, line 146: Please leave this intact. That does not worth it, to get a trivial S2PTR macro.

src/net.c
```diff
@@ -174,6 +174,7 @@ struct mg_connection *mg_connect(struct mg_mgr *mgr, const char *url,
   } else {
     LIST_ADD_HEAD(struct mg_connection, &mgr->conns, c);
     c->is_udp = (strncmp(url, "udp:", 4) == 0);
+    c->fd = S2PTR(INVALID_SOCKET);
```
cpq, line 177: I'd do

```c
  c->fd = (void *) (size_t) -1;  // Set to INVALID_SOCKET
```

Makefile
scaprile: INVALID_SOCKET is the one

src/net.c
scaprile: But INVALID_SOCKET is platform dependent:
MG_ARCH_FREERTOS_TCP defines it to FREERTOS_INVALID_SOCKET
MG_ARCH_WIN32 && MG_ENABLE_WINSOCK and MG_ARCH_TIRTOS seem to have a built-in value
and else it is defined to -1, so just using -1 might fail in 
https://github.com/cesanta/mongoose/blob/4271aa5b1f97f09cddf8688977ef43bd541f1887/mongoose.c#L4248-L4259

cpq: Hardcoding to -1 seems fine to me

PR #1840, opened 2022-11-08
Update mip_test.c

jfsimon1981:
HTTP fetch using MIP stack.

### DIFF
--- a/mip/mip.c
+++ b/mip/mip.c
@@ -299,11 +299,12 @@ static struct ip *tx_ip(struct mip_if *ifp, uint8_t proto, uint32_t ip_src,
   struct eth *eth = (struct eth *) ifp->tx.ptr;
   struct ip *ip = (struct ip *) (eth + 1);
   uint8_t *mac = arp_cache_find(ifp, ip_dst);  // Dst IP in ARP cache ?
-  if (!mac && (ip_dst & ifp->mask)) arp_ask(ifp, ip_dst);  // Same net, lookup
-  if (!mac) mac = arp_cache_find(ifp, ifp->gw);            // Use gateway MAC
-  if (mac) memcpy(eth->dst, mac, sizeof(eth->dst));        // Found? Use it
-  if (!mac) memset(eth->dst, 255, sizeof(eth->dst));       // No? Use broadcast
-  memcpy(eth->src, ifp->mac, sizeof(eth->src));  // TODO(cpq): ARP lookup
+  if (!mac && ((ip_dst & ifp->mask) == (ifp->ip & ifp->mask)))
+    arp_ask(ifp, ip_dst);                             // Same net, lookup
+  if (!mac) mac = arp_cache_find(ifp, ifp->gw);       // Use gateway MAC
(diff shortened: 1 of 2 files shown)


jfsimon1981:
HTTP fetch with MIP stack.
Compilation warnings (void* to int conversion).
Additional Makefile WARN flags -Wno-pointer-to-int-cast -Wno-int-to-pointer-cast

test/mip_test.c
```diff
@@ -38,6 +54,8 @@ static void test_queue(void) {
   assert(q_avail(&q) == 0);
 }
 
+// *****************************************************************************
```
cpq, line 57: Please remove this.
And below

test/mip_test.c
```diff
@@ -49,9 +67,108 @@ static void test_statechange(void) {
   onstatechange(&iface);
 }
 
+// *****************************************************************************
+// Implement tap_rx() tap_tx() tap_up()
+
+static size_t tap_rx(void *buf, size_t len, void *userdata) {
+  ssize_t received = read((int) userdata, buf, len); // Original
```
cpq, line 74: As discussed, please try this to squash the warning:

```c
read((int) (size_t) userdata, buf, len);
```

If that does not work, search for other ways on how to do a cast without a warning.

test/mip_test.c
```diff
@@ -49,9 +67,108 @@ static void test_statechange(void) {
...
+  } else if (ev == MG_EV_HTTP_MSG) {
+    struct mg_http_message *hm = (struct mg_http_message *) ev_data;
+    // MG_INFO(("-->[%.*s]", (int) hm->message.len, hm->message.ptr));
+    // ASSERT(mg_vcmp(&hm->method, "HTTP/1.1") == 0);
+    // ASSERT(mg_vcmp(&hm->uri, "301") == 0);
+    *fn_data_int_ptr = mg_http_status(hm);
+  } else if (ev == MG_EV_CLOSE) {
+    if (*fn_data_int_ptr == 0) *fn_data_int_ptr = 888;
+  } else if (ev == MG_EV_ERROR) {
+    if (*fn_data_int_ptr == 0) *fn_data_int_ptr = 777;
```
cpq, line 113: That's a copy/paste from the unit_test which is not used in this test. Please remove EVERYTHING that is not used. Leave only relevant code.

test/mip_test.c
```diff
@@ -49,9 +67,108 @@ static void test_statechange(void) {
...
+  int ok = 0;
+  struct mg_connection *conn;
+
+  int pc = 100; // Wait sufficiently to acquire HTTP lease
+  while ((pc--)>0) {
+    mg_mgr_poll(&mgr, 100);
+    usleep(10000); // 10 ms
+  }
+
+  conn = mg_http_connect(&mgr, "http://cesanta.com", f_http_fetch, &ok);
```
cpq, line 153: We might do several HTTP requests, to fetch small and large chunks, with and without SSL.
Therefore, please make a function `char *fetch(struct mg_mgr *mgr, const char *url, const char *post_data)` that can be called multiple times. The function should return full HTTP message including headers.

cpq (changes requested):

PR #1841, opened 2022-11-09
Endianness test

cpq:


jameshilliard:
Was this supposed to fix the broken endian tests? It looks like it just disables tests entirely for the tests that don't work on big endian and 32 bit platforms.

cpq:
No, the fix is under way - there are issues opened.
@jameshilliard thanks for the earlier contribution, but your way of disabling large chunks of the unit test is just not how it should be fixed :)

PR #1842, opened 2022-11-09
Add nucleo-f429zi-freertos-mip example

cpq:


examples/stm32/nucleo-f429zi-freertos-mip/Makefile
```diff
@@ -0,0 +1,33 @@
...
+LDFLAGS ?= -Tlink.ld -nostartfiles -nostdlib --specs nano.specs -lc -lgcc -Wl,--gc-sections -Wl,-Map=$@.map
+SOURCES = main.c startup.c syscalls.c ../../../mongoose.c
+
+FREERTOS_VERSION ?= V10.5.0
+FREERTOS_REPO ?= https://github.com/FreeRTOS/FreeRTOS-Kernel 
+
+build example: firmware.elf
+
+firmware.elf: FreeRTOS-Kernel $(SOURCES) 
+	arm-none-eabi-gcc -o $@ $(SOURCES) $(CFLAGS) $(LDFLAGS) \
```
scaprile: This requires having the ARM toolchain installed. Why not keep using the docker version ? (Yes, those nasty warnings, but we'll get rid of them, eventually...)

examples/stm32/nucleo-f429zi-freertos-mip/mcu.h
```diff
@@ -0,0 +1,161 @@
...
+  if (uart == UART3) RCC->APB1ENR |= BIT(18);
+
+  if (uart == UART1) af = 4, tx = PIN('A', 9), rx = PIN('A', 10);
+  if (uart == UART2) af = 4, tx = PIN('A', 2), rx = PIN('A', 3);
+  if (uart == UART3) af = 7, tx = PIN('D', 8), rx = PIN('D', 9);
+
+  gpio_init(tx, GPIO_MODE_AF, GPIO_OTYPE_PUSH_PULL, GPIO_SPEED_HIGH, 0, af);
+  gpio_init(rx, GPIO_MODE_AF, GPIO_OTYPE_PUSH_PULL, GPIO_SPEED_HIGH, 0, af);
+  uart->CR1 = 0;                           // Disable this UART
+  uart->BRR = FREQ / APB2_PRE / baud;      // FREQ is a CPU frequency
```
scaprile, line 120: Same as #1839 , happy coincidence
```
uart->BRR = FREQ / 4 / baud
```

examples/stm32/nucleo-f429zi-freertos-mip/mcu.h
```diff
@@ -0,0 +1,161 @@
...
+static inline bool timer_expired(uint32_t *t, uint32_t prd, uint32_t now) {
+  if (now + prd < *t) *t = 0;                    // Time wrapped? Reset timer
+  if (*t == 0) *t = now + prd;                   // Firt poll? Set expiration
+  if (*t > now) return false;                    // Not expired yet, return
+  *t = (now - *t) > prd ? now + prd : *t + prd;  // Next expiration time
+  return true;                                   // Expired, return true
+}
+
+static inline void clock_init(void) {                 // Set clock frequency
+  SCB->CPACR |= ((3UL << 10 * 2) | (3UL << 11 * 2));  // Enable FPU
```
scaprile, line 150: `DSB` + `ISB` is recommended after this to purge the pipeline, though I don't see any FP operations in the near future

examples/stm32/nucleo-f429zi-freertos-mip/mcu.h
```diff
@@ -0,0 +1,161 @@
...
+  return (uint8_t) (uart->DR & 255);
+}
+
+static inline bool timer_expired(uint32_t *t, uint32_t prd, uint32_t now) {
+  if (now + prd < *t) *t = 0;                    // Time wrapped? Reset timer
+  if (*t == 0) *t = now + prd;                   // Firt poll? Set expiration
+  if (*t > now) return false;                    // Not expired yet, return
+  *t = (now - *t) > prd ? now + prd : *t + prd;  // Next expiration time
+  return true;                                   // Expired, return true
+}
```
scaprile, line 147: This is problematic, and I don't see you use it. Is it needed ?
If _now_ is guaranteed not to go backwards (don't know how FreeRTOS handles timing but I think it is monotonic. BTW, don't they have functions for this ?) and the function is called frequently enough so it can't meet more than one overflow between calls, unsigned difference survives the overflow, arithmetics are on our side and these hold:
```
	last_time = MS_TIMER;

	if(MS_TIMER - last_time > interval)
	    printf("Timer expired\n");

	if(MS_TIMER - interval > last_time)
	    printf("Timer expired\n");
```
Using signed for the variables (not the counter), this holds for intervals upto MAXINT:
```
int32_t deadline;
int32_t interval = 5000; // milliseconds

	deadline = MS_TIMER + interval;
	if((int32_t)(MS_TIMER - deadline) > 0)
	    printf("Timer expired\n");
```
I would remove this function anyway

PS: Source: http://www.scaprile.com/2022/04/29/timer-expired-i-think/ nice guy, btw...

examples/stm32/nucleo-f429zi-freertos-mip/mcu.h
```diff
@@ -0,0 +1,161 @@
...
+static inline void spin(volatile uint32_t count) {
+  while (count--) asm("nop");
+}
+
+static inline void systick_init(uint32_t ticks) {
+  if ((ticks - 1) > 0xffffff) return;  // Systick timer is 24 bit
+  SysTick->LOAD = ticks - 1;
+  SysTick->VAL = 0;
+  SysTick->CTRL = BIT(0) | BIT(1) | BIT(2);  // Enable systick
+  RCC->APB2ENR |= BIT(14);                   // Enable SYSCFG
```
scaprile, line 39: ` RCC->APB2ENR |= BIT(14);`
This doesn't really belong here, and is duplicated in _main.c_

examples/stm32/nucleo-f429zi-freertos-mip/main.c
```diff
@@ -0,0 +1,77 @@
...
+    MG_INFO(("blink %s,  RAM: %u", (char *) args, xPortGetFreeHeapSize()));
+  }
+}
+
+int main(void) {
+  clock_init();               // Set clock to max of 180 MHz
+  systick_init(FREQ / 1000);  // Tick every 1 ms
+  uart_init(UART3, 115200);   // Initialise UART
+  xTaskCreate(blinker, "blinker", 512, ":)", configMAX_PRIORITIES - 1, NULL);
+  xTaskCreate(server, "server", 8192, 0, configMAX_PRIORITIES - 1, NULL);
```
scaprile, line 74: FreeRTOS task size is in words, 128 should be enough for the blinker, and the web server... the device-dashboard works with 2048 words

examples/stm32/nucleo-f429zi-freertos-mip/Makefile
cpq: Docker added, yes

examples/stm32/nucleo-f429zi-freertos-mip/mcu.h
cpq: Yeah so I'd like to have a minimal code that does the task

examples/stm32/nucleo-f429zi-freertos-mip/mcu.h
cpq: Yep agreed , removed!

examples/stm32/nucleo-f429zi-freertos-mip/mcu.h
cpq: Well.. Yes, it is duplicated in main.c but for a different reason - to enable ethernet.
Here, we enable systick. 
I'll remove the main.c line

examples/stm32/nucleo-f429zi-freertos-mip/main.c
cpq: Good point Sergio, I did not realise that it is in words!
Modified.

examples/stm32/nucleo-f429zi-freertos-mip/mcu.h
scaprile: I beg to differ. I would remove it from here and set it at main.c

Rationale:
In main.c you do act on the SYSCFG to enable RMII, so it makes sense to enable the SYSCFG clock there.
SysTick is ARM, it does not have a clock gating and (usually) does not depend on vendor peripherals. Plus, you are acting on this _after_ setting the SysTick registers, if it were actually needed you'd have met Mr. HardFault.
From 5.2 at the datasheet:
> The RCC feeds the external clock of the Cortex System Timer (SysTick) with the AHB clock
(HCLK) divided by 8. The SysTick can work either with this clock or with the Cortex clock
(HCLK), configurable in the SysTick control and status register.

SysTick control and status register is also ARM's (SYST_CSR in Ozone), we don't act upon it and defaults to CLJSOURCE=1, receiving the Cortex clock. https://developer.arm.com/documentation/ddi0413/d/system-control/system-control-register-descriptions/systick-control-and-status-register
 
Since we calculate ticks from FREQ and they seem correct, it is receiving the Cortex clock, as usual.
Verified everything works if removing this line here and leaving it on main.c

cpq: Done!

PR #1850, opened 2022-11-11
Fix incorrect local network address detection

scaprile:
An IP address is not local if it does not belong in the local address space.
The local address space is the local IP address anded with the IP mask.
The destination IP address is then anded with the IP mask and if it resides in the local address space, both should be equal. Probably there is a more performant way to do this.

clang format added some artifacts, and now that the TM4C driver has been merged, mongoose.[ch] include it


cpq:
LGTM (looks good to me), please merge

PR #1869, opened 2022-11-18
Add missing include

scaprile:
linux/if uses struct sockaddr that is defined in sys/socket.h. This breaks building on some systems (RHEL7). It went unnoticed as it was hidden by mongoose.h being included first


jfsimon1981 (approved):
All right, OpenBSD / Debian compilations.

cpq (approved):

PR #1874, opened 2022-11-21
Fix wildcard in Makefile

scaprile:
The construction `*.c` works OK on Un*x derivatives but fails on Windows, as it is the shell who expands the wildcard.
The construction `$(wildcard *.c)` seems to work everywhere as it is make who expands the wildcard.


cpq:
Are you sure that works when sources are NOT yet cloned, and you run `make build` ?

scaprile:
Yes, verified in Linux and Windows (mingw32-make from MinGW-W64)

PR #1883, opened 2022-11-25
Fix sending ARP request for IP broadcasts

scaprile:
Like e.g.: at startup when sending DHCP Request
Returning a broadcast MAC from the cache lookup avoids this behavior

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -6759,9 +6759,14 @@ static inline void arp_cache_dump(const uint8_t *p) {
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


mongoose.c
```diff
@@ -6759,9 +6759,14 @@ static inline void arp_cache_dump(const uint8_t *p) {
 }
 #endif
 
+static const uint8_t bcastmac[6] = {0xFF,0xFF,0xFF,0xFF,0xFF,0xFF};
```
cpq, line 6762: Move this inside arp_cache_find() ?

cpq (approved):

PR #1904, opened 2022-12-05
Improve TM4C driver

scaprile:
- Fix clock guessing
- Frames are now read in order, and only host owned descriptors are checked
- Read as many frames as possible (but don't loop forever)
- Discard bad frames
- Fixed incorrect handling of RU bit and exit out of suspended condition
- If a frame arrives while processing, it will now generate a new interrupt (formerly it would wait for another frame to arrive as we reset the flag on exit)


cpq (commented):
Only a very minor style comment.
LGTM, feel free to address a comment and merge.

PR #1958, opened 2023-01-10
Close connection on received TCP RST frame

psykokwak-com:
Do not reset TCP KEEPALIVE timeout on TCP send frame

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -7049,7 +7049,7 @@ long mg_io_send(struct mg_connection *c, const void *buf, size_t len) {
   if (tx_tcp(ifp, c->rem.ip, TH_PUSH | TH_ACK, c->loc.port, c->rem.port,
              mg_htonl(s->seq), mg_htonl(s->ack), buf, len) > 0) {
     s->seq += (uint32_t) len;
-    settmout(c, MIP_TTYPE_KEEPALIVE);
+    //settmout(c, MIP_TTYPE_KEEPALIVE);
   } else {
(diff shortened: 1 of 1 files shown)


scaprile:
We did that for a reason and will properly implement it in due time

psykokwak-com:
Ok but in the meantime we can have scenarios of resource saturation because of the absence of detection of the RST and the absence of wachdog (keepalive). And it's not cool at all.
My bugfix allows at least to avoid falling in this case. I think you should integrate this PR the time to improve the implementation of the MIP stack.

scaprile:
TCP Keepalives should only be sent when idle oe in response to a keepalive from the other end, that is one of the reasons of that timer reset being there.
Regarding the RST, please explain where and why the RST is not detected, add a capture file if possible and if something is not RFC-compliant, please tell us why

psykokwak-com:
If mongoose is streaming data over a TCP connection and the other side sends an "RST" frame (to disconnect), then mongoose ignores the frame and continues to stream "in the void". And as the watchdog resets each time a TCP frame is sent, it will never detect that there is no one in front.
And worse: if there is a break in the link between the two parties, mongoose will also continue to stream without ever stopping.
The two modifications that I propose are used to correct (while waiting for better) these two problems.

scaprile:
#1959

psykokwak-com:
Great. Thank you.
And what about the KEEPALIVE in case of no RST|FIN frame ?

scaprile:
#1960

PR #1967, opened 2023-01-16
Add example Dockerfile to build and run http-server

scaprile:


cpq:
Thanks Sergio!
Is this Dockerfile used by something, like, a tutorial?

scaprile:
> Thanks Sergio! Is this Dockerfile used by something, like, a tutorial?

No, but we could add it to the HTTP server tutorial, what do you think ?

cpq:
@scaprile to be honest, I think docker would be an overkill..

I think we should also kill all docker usage for ARM builds, and rely on ARM GCC installed

scaprile:
Moving to the chat...

PR #1968, opened 2023-01-16
Don't retrigger ACK timer

scaprile:
When delaying ACKs, start counting on the first incoming segment instead of the last one.

We should actually also ACK when our window has been exhausted but that is a bit more involved


psykokwak-com:
> We should actually also ACK when our window has been exhausted but that is a bit more involved

Yep. It will considerably increase the performance.

PR #1986, opened 2023-01-21
Add examples/rp2040/pico-w5500 to hwtest

cpq:


scaprile (approved):
The README file still talks about "example" instead of the new (and standard) "firmware"

PR #1988, opened 2023-01-22
Add nucleo-h743zi-baremetal boilerplate, with HW test

cpq:


examples/stm32/nucleo-h743zi-baremetal/Makefile
```diff
@@ -0,0 +1,38 @@
+CFLAGS  ?=  -W -Wall -Wextra -Werror -Wundef -Wshadow -Wdouble-promotion \
+						-Wformat-truncation -fno-common -Wconversion -Wno-sign-conversion \
+						-g3 -Os -ffunction-sections -fdata-sections \
+            -I. -Ih7/Include -Icmsis/CMSIS/Core/Include \
+            -mcpu=cortex-m7 -mthumb -mfloat-abi=soft \
```
scaprile, line 5: TODO(): hard FPU

examples/stm32/nucleo-h743zi-baremetal/main.c
```diff
@@ -0,0 +1,40 @@
...
+  mg_log_set(MG_LL_DEBUG);  // Set log level
+  mg_timer_add(&mgr, BLINK_PERIOD_MS, MG_TIMER_REPEAT, timer_cb, NULL);
+
+  MG_INFO(("Starting event loop"));
+  for (;;) {
+    mg_mgr_poll(&mgr, 0);
+  }
+
+  return 0;
+}
```
scaprile, line 40: No CMSIS ?

scaprile (approved):
Sadly I don't have a board to help

examples/stm32/nucleo-h743zi-baremetal/Makefile
cpq: Yeah, there is FPU and other TODOs in the mcu.h

examples/stm32/nucleo-h743zi-baremetal/main.c
cpq: Well, we could pull both ARM Core cmsis headers and stm32h74x headers.
But that would make this example different from F4 and F7, which do not use CMSIS headers.
Shall we use CMSIS for all cases?

PR #1995, opened 2023-01-29
Add mg_print_ip, mg_print_ip_port helper functions

cpq:
Our mg_*printf API has a non-standard extensions: `%M`, `%I`, `%H`, `%A`, etc etc.

We might want more functionality.
For example, we currently have `%Q` specifier for double-quoted string - we might want to add a non-quoted variant for strings, to handle cases e.g. `mg_printf(c, "{%Q: \"hello, %q\"}", "message", "escape, but not quote");` 

Extra functionality is good.
But non-standard is bad.

We should move towards using a minimum non-standard features.
This PR makes `%I` specifier not needed anymore:

Before:

```c
mg_printf(c, "%I", c->rem.ip);  // Prints IPv4 address. Caller must be sure it is IPv4
```
After:

```c
mg_printf(c, "%M", mg_print_ip, &c->rem);  // Prints IP address, IPv4 or IPv6
```

This PR cleans up `%I` usage in `src/` and `test/`. 
Examples, and README still have `I%`

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -207,6 +207,25 @@ struct mg_timer *mg_timer_add(struct mg_mgr *mgr, uint64_t milliseconds,
   return t;
 }
 
+size_t mg_print_ip(void (*out)(char, void *), void *arg, va_list *ap) {
+  struct mg_addr *addr = va_arg(*ap, struct mg_addr *);
+  if (addr->is_ip6) {
+    uint16_t *p = (uint16_t *) addr->ip6;
+    return mg_xprintf(out, arg, "[%x:%x:%x:%x:%x:%x:%x:%x]", mg_htons(p[0]),
+                      mg_htons(p[1]), mg_htons(p[2]), mg_htons(p[3]),
+                      mg_htons(p[4]), mg_htons(p[5]), mg_htons(p[6]),
+                      mg_htons(p[7]));
+  } else {
+    uint8_t *p = (uint8_t *) &addr->ip;
+    return mg_xprintf(out, arg, "%d.%d.%d.%d", p[0], p[1], p[2], p[3]);
+  }
+}
+
+size_t mg_print_ip_port(void (*out)(char, void *), void *arg, va_list *ap) {
+  struct mg_addr *a = va_arg(*ap, struct mg_addr *);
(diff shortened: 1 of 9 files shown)


src/tls_mbed.c
```diff
@@ -66,7 +66,7 @@ static int mbed_rng(void *ctx, unsigned char *buf, size_t len) {
 
 static void debug_cb(void *c, int lev, const char *s, int n, const char *s2) {
   n = (int) strlen(s2) - 1;
-  MG_VERBOSE(("%lu %d %.*s", ((struct mg_connection *) c)->id, lev, n, s2));
+  MG_INFO(("%lu %d %.*s", ((struct mg_connection *) c)->id, lev, n, s2));
```
scaprile, line 69: Changing _VERBOSE to _INFO inside a "debug" callback raised my eyebrows.
Intended or test leftover ?

src/net.c
```diff
@@ -207,6 +207,25 @@ struct mg_timer *mg_timer_add(struct mg_mgr *mgr, uint64_t milliseconds,
...
 }
 
+size_t mg_print_ip(void (*out)(char, void *), void *arg, va_list *ap) {
+  struct mg_addr *addr = va_arg(*ap, struct mg_addr *);
+  if (addr->is_ip6) {
+    uint16_t *p = (uint16_t *) addr->ip6;
+    return mg_xprintf(out, arg, "[%x:%x:%x:%x:%x:%x:%x:%x]", mg_htons(p[0]),
+                      mg_htons(p[1]), mg_htons(p[2]), mg_htons(p[3]),
+                      mg_htons(p[4]), mg_htons(p[5]), mg_htons(p[6]),
+                      mg_htons(p[7]));
```
scaprile, line 217: ntohs ?

scaprile (commented):
I'll take care of the examples and tutorials

src/tls_mbed.c
cpq: That's fine, cause if this callback is called, we want to see what it prints.

src/net.c
scaprile: I know, it is just a reversal, it just makes the one who sees the code know in which format was the origin and in which format the destination will be... it helps when following the code for maintenance.
Anyway, I can change it later if you want to push this as is.

Issue #1997, opened 2023-01-29
Kill `%A`, create `mg_print_mac`


cpq:
Add `mg_print_mac()` function, and get rid of `%A` specifier everywhere.
Document `mg_print_mac()`.

Move all `%M` helper functions to `src/util.c`, just to have them in one place - including `mg_print_ip*`.

Issue #1996, opened 2023-01-29
Kill `%I`, use `mg_print_ip()`


cpq:
Change `%I` usage in `examples/*` to `mg_print_ip`
Remove `%I` from doc/README.md, and instead, document `mg_print_ip()` and `mg_print_ip_port()`

PR #2003, opened 2023-01-31
Do not use %I %A in MIP

scaprile:
util.c:  Added mg_print_ip4 (and mg_print_ip6) as it was needed by MIP. Refactored mg_print_ip to use it

#1996 #1997


cpq:
LGTM

cpq (approved):

PR #2008, opened 2023-02-01
Merged commit for driver i.MX RT1020.

jfsimon1981:
Driver for i.MX RT1020 MIP Bare
Evaluation board MIMXRT1020-EVK


mip/driver_imx_rt1020.c
```diff
@@ -0,0 +1,302 @@
...
+void delay(uint32_t);
+void delay (uint32_t di) {
+  volatile int dno = 0; // Prevent optimization
+  for (uint32_t i = 0; i < di; i++)
+    for (int j=0; j<20; j++) // PLLx20 (500 MHz/24MHz)
+      dno++;
+}
+
+void wait_phy_complete(void);
+void wait_phy_complete(void) {
```
scaprile, line 45: static

mip/driver_imx_rt1020.c
```diff
@@ -0,0 +1,302 @@
...
+{
+    uint16_t length;  // Data length
+    uint16_t control; // Control and status
+    uint32_t *buffer; // Data ptr
+} enet_bd_struct_t;
+
+// Descriptor and buffer globals, in non-cached area, 64 bits aligned.
+
+__attribute__((section("NonCacheable,\"aw\",%nobits @"))) enet_bd_struct_t rx_buffer_descriptor[(ENET_RXBD_NUM)] __attribute__((aligned((64U))));
+__attribute__((section("NonCacheable,\"aw\",%nobits @"))) enet_bd_struct_t tx_buffer_descriptor[(ENET_TXBD_NUM)] __attribute__((aligned((64U))));
```
scaprile, line 104: Is this ARM GCC specific ? Hope Keil handles it

mip/driver_imx_rt1020.c
```diff
@@ -0,0 +1,302 @@
...
+  // ENET Reset, wait complete
+  ENET->ECR |= BIT_SET(0);
+  while((ENET->ECR & BIT_SET(0)) != 0) {}
+
+  // Re-latches the pin strapping pin values
+  ENET->ECR |= BIT_SET(0);
+  while((ENET->ECR & BIT_SET(0)) != 0) {}
+
+  // Setup MII/RMII MDC clock divider (<= 2.5MHz).
+  ENET->MSCR = 0x130; // HOLDTIME 2 clk, Preamble enable, MDC MII_Speed Div 0x30
```
scaprile, line 121: This is clock speed dependent

mip/driver_imx_rt1020.c
```diff
@@ -0,0 +1,302 @@
...
+
+      if (!linkup) {
+        MG_ERROR(("Error: Link not ready"));
+      }
+      else {
+        MG_INFO(("Link ready"));
+      }
+    }
+  }
+
```
scaprile, line 168: Is this really necessary ? We poll the PHY every second for link status

mip/driver_imx_rt1020.c
```diff
@@ -0,0 +1,302 @@
...
+};
+
+#undef ENET
+#define ENET ((struct imx_rt1020_enet *) (uintptr_t) 0x402D8000u)
+
+#undef BIT_SET
+#define BIT_SET(x) ((uint32_t) 1 << (x))
+
+#undef BIT_CLEAR
+#define BIT_CLEAR(x) ((uint32_t) ~(1U << (x)))
```
scaprile, line 15: These names are confusing, look like they set and clear bits. I would add the word _IS_ but why not using `BIT()` ?

mip/driver_imx_rt1020.c
```diff
@@ -0,0 +1,302 @@
...
+
+void ENET_IRQHandler(void) {
+  ENET->EIMR = 0;           // Mask interrupts.
+  uint32_t eir = ENET->EIR; // Read EIR
+  ENET->EIR = 0xffffffff;   // Clear interrupts
+
+  if (eir & EIMR_RX_ERR) // Global mask used
+  {
+    if (rx_buffer_descriptor[s_rxno].control & BIT_SET(15)) {
+      ENET->EIMR = EIMR_RX_ERR; // Enable interrupts
```
scaprile, line 265: I personally don't like fiddling with enabling and disabling interrupts, particularly within their own handler. AFAIK IRQs in ARM are not reentrant, there is an _active_ flag in the NVIC.

mip/driver_imx_rt1020.c
```diff
@@ -0,0 +1,302 @@
...
+  (void) userdata;
+  return bsr & BIT_SET(2) ? 1 : 0;
+}
+
+// Set receive callback for interrupt-driven drivers
+static void mip_driver_imx_rt1020_setrx(void (*rx)(void *, size_t, void *),
+                                   void *rxdata) {
+  s_rx = rx;
+  s_rxdata = rxdata;
+}
```
scaprile, line 295: Oh, this is really old stuff... time passes quickly (and chances are few)...

mip/driver_imx_rt1020.c
jfsimon1981: Bit set does mask like 001000 and bit clear like 110111.

mip/driver_imx_rt1020.c
jfsimon1981: Right i posted code from branch, let me work on update.

mip/driver_imx_rt1020.c
scaprile: I mean:
BIT_SET -> means "to set bits"
BIT_IS_SET or IS_SET -> means "check if the bit is set and tell me"
We already have a BIT macro, |= BIT() sets bits and &= ~BIT() clears bits.

mip/driver_imx_rt1020.c
jfsimon1981: Ok, i need to check this further. The IRQ goes back a few function calls away. It has not to be re-entrant.

mip/driver_imx_rt1020.c
jfsimon1981: I see you suggest, so let just DHCP/Network code in general do the timout/reties and delete this graceful retry in init right ?
You sure that's better ? i can update of course.

mip/driver_imx_rt1020.c
jfsimon1981: Yes so far the clock is only 50 MHz, there's peripheral clock being used in there.
I'll have to work on this to be more general. Though it should be working all right for the 10/100M setup.

mip/driver_imx_rt1020.c
scaprile: BTW, that is a quotation from a Toto song... I should have changed it to "and changes aren't few"

mip/driver_imx_rt1020.c
jfsimon1981: Done

mip/driver_imx_rt1020.c
jfsimon1981: Ok, i updated it

mip/driver_imx_rt1020.c
scaprile: Yes, both Keil compilers support that dialect.
This is one of the reasons CMSIS exists...

jfsimon1981: Ok thanks Sergio

mip/driver_stm32.c
```diff
@@ -1,5 +1,6 @@
 #include "mip.h"
 
+#ifndef MG_ENABLE_DRIVER_IMXRT1020
 #if MG_ENABLE_MIP && \
     (!defined(MG_ENABLE_DRIVER_TM4C) || MG_ENABLE_DRIVER_TM4C == 0)
```
scaprile, line 5: See #2020 , I modded this driver to check for its own enable macro.
That macro is defined in mip.h to 1 when no other clashing driver is selected

mip/driver_tm4c.c
```diff
@@ -1,5 +1,6 @@
 #include "mip.h"
 
+#ifndef MG_ENABLE_DRIVER_IMXRT1020
```
scaprile, line 3: and we don't need this, this driver is disabled by default

mip/driver_w5500.c
```diff
@@ -1,5 +1,6 @@
 #include "mip.h"
 
+#ifndef MG_ENABLE_DRIVER_IMXRT1020
```
scaprile, line 3: We don't need this, your driver should not clash with this one (no other driver does)

mip/driver_tm4c.c
jfsimon1981: Name clashes with enum/phy accessors. Driver enum/functions need be unique'd.
Is that wishable ?

mip/driver_w5500.c
jfsimon1981: This needs enums and phy functions to be prefixed. I have the draft for this but driver code needs be unique'd.

scaprile: STM32, STM32H, and TM4C, do not clash with this driver, and do not have prefixes

mip/driver_tm4c.c
scaprile: This driver is only enabled when selected, and won't be selected with the imx driver at the same time; I don't see why

scaprile: It could be left enabled as its IRQ handler has a different name, we'll cross that bridge when we get there,

jfsimon1981: Driver's included as soon as MG_ENABLE_MIP. 2 other drivers have name collisions This happens with enums and phy_read phy_write functions (stm32+w5500 come with MIP definition).

Drivers can go with prefix or unique inclusion way. I opted for the second, but i also drafted the 1st solution.
This was discussed a bit. I'll leave it open for the moment. This commit isn't pushed.

jfsimon1981:
This driver doesn't merge anymore.
The branch will probably need to be dropped.
Branch with comments kept for review only.
No further commit on this one ...

Sergio, you modded all drivers it seems, to the way i planned originally, hence i can just discard side modifications to stm/ti as rt1020 shouldn't collide in the future.

PR #2020, opened 2023-02-03
Add STM32H743ZI driver and example

scaprile:
Minimum changes to existing code
Hardware test enabled and working
STM32 driver remains as default, `mip.h` disables it if other driver is selected
Non-ARM drivers not affected
New driver is STM32H and requires `MG_ENABLE_DRIVER_STM32H=1`, defined in Makefile


examples/stm32/nucleo-h743zi-baremetal/Makefile
```diff
@@ -25,11 +25,11 @@ flash: firmware.bin
...
 # Requires env variable VCON_API_KEY set
 DEVICE_URL ?= https://dash.vcon.io/api/v3/devices/6
-test: update
-	curl --fail-with-body -su :$(VCON_API_KEY) $(DEVICE_URL)/tx?t=5 | tee /tmp/output.txt
-	grep 'ticks:' /tmp/output.txt
-#	grep 'Ethernet: up' /tmp/output.txt
-#	grep 'MQTT connected' /tmp/output.txt
+test:	EXTRA_CFLAGS += -DUART_DEBUG=UART1
+test:	update
+	curl --fail -su :$(VCON_API_KEY) $(DEVICE_URL)/tx?t=5 | tee /tmp/output.txt
```
cpq, line 30: Nit: use `--fail-with-body` as it gives a clue about the failure

scaprile: done
(my version does not support either form...)

examples/stm32/nucleo-h743zi-baremetal/README.md
```diff
@@ -0,0 +1,41 @@
+# Baremetal webserver on NUCLEO-H743ZI
```
cpq, line 1: As soon as we have a tutorial for H7, this README should be a one-line reference to that tutorial. This ensures we have a single place of authority , which does not rot , and things do not go out of sync with each other

cpq (approved):

examples/stm32/nucleo-h743zi-baremetal/README.md
scaprile: Good idea, will do the same for F4 and F7

PR #2023, opened 2023-02-04
Reduce the number of redundant test runs. Disable SNTP test on macos …

cpq:
…- it is flaky


scaprile (commented):
I get a bit lost, shouldn't we also check `^mip` for changes ?
We are now checking only on push, that means there won't be an extra check when the PR is created, great!. Any side-effects ?

PR #2024, opened 2023-02-06
Use CMSIS headers for f429 and f746 baremetal examples

cpq:


examples/stm32/nucleo-f429zi-baremetal/main.c
```diff
@@ -31,20 +31,23 @@ void EXTI15_10_IRQHandler(void) {  // External interrupt handler
   gpio_write(LED1, gpio_read(BTN1));  // No debounce. Turn LED if button pressed
 }
 
-static void timer_cb(void *arg) {
+void SystemInit(void) {                  // Called automatically by startup code
+  clock_init();                          // Set clock to 180MHz
+  SysTick_Config(SYS_FREQUENCY / 1000);  // Increment s_ticks every ms
+}
```
scaprile, line 37: Yes... but the user might pull-in the provided system_devicename.c ... if this is going to be a skeleton we should perhaps write a system file or use it. Some set the clock to highest frequency, some don't...

examples/stm32/nucleo-f429zi-baremetal/hal.h
```diff
@@ -146,22 +92,28 @@ static inline void irq_exti_attach(uint16_t pin) {
...
+USART_TypeDef {
   volatile uint32_t SR, DR, BRR, CR1, CR2, CR3, GTPR;
 };
-#define UART1 ((struct uart *) 0x40011000)
-#define UART2 ((struct uart *) 0x40004400)
-#define UART3 ((struct uart *) 0x40004800)
+#define UART1 ((USART_TypeDef *) 0x40011000)
+#define UART2 ((USART_TypeDef *) 0x40004400)
+#define UART3 ((USART_TypeDef *) 0x40004800)
+#endif
```
scaprile, line 106: Why keeping these ? (and others)

examples/stm32/nucleo-f429zi-baremetal/Makefile
```diff
@@ -1,44 +1,60 @@
...
+CMSIS_DEVICE_VERSION ?= v2.6.8   # ST MCU peripheral definitions
+CMSIS_DEVICE_REPO ?= https://github.com/STMicroelectronics/cmsis_device_f4
+
+CFLAGS ?= -W -Wall -Wextra -Werror -Wundef -Wshadow -Wdouble-promotion \
+          -Wformat-truncation -fno-common -Wconversion \
+          -g3 -Os -ffunction-sections -fdata-sections \
+          -I . -I cmsis_core/CMSIS/Core/Include -I cmsis_device_f4/Include \
+          -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16 $(EXTRA_CFLAGS)
+LDFLAGS ?= -Tlink.ld -nostdlib -nostartfiles --specs nano.specs -lc -lgcc -Wl,--gc-sections -Wl,-Map=$@.map
+SOURCES = main.c syscalls.c cmsis_device_f4/Source/Templates/gcc/startup_stm32f429xx.s
```
scaprile, line 13: We are introducing assembly files, syntax changes between compilers, we should perhaps add this in a different line with a comment; I mean, if this is going to be a skeleton

examples/stm32/nucleo-f429zi-baremetal/main.c
cpq: I think providing per-clock examples would be an overkill, TBH.
I do not expect any conflicts, if our examples are used as skeletons.
Conflicts can happen if user integrates our code into their existing example, which is a different use case, IMO.

examples/stm32/nucleo-f429zi-baremetal/hal.h
cpq: Good point! let me clean that up in my next PR which is coming ...

examples/stm32/nucleo-f429zi-baremetal/Makefile
cpq: Good point too! Will fix in my next PR.

examples/stm32/nucleo-f429zi-baremetal/main.c
scaprile: Oh, no, no, just the highest speed...
I meant keeping system stuff in a system file of our own or integrating with the vendor system file

scaprile: Perhaps just putting this into our system.c is the easiest/fastest. If the user wants to integrate with his example will know at first sight that there is system stuff to handle.
Probably the `clock_init()` function itself should move to that system file; when vendors do set the clock, they do it in the system_xxxx.c file

PR #2027, opened 2023-02-06
Introduce sysinit.c, better debug log, more verbose Makefile

cpq:
+ eport randomizing

### DIFF
--- a/examples/stm32/nucleo-f746zg-baremetal/main.c
+++ b/examples/stm32/nucleo-f746zg-baremetal/main.c
@@ -4,53 +4,35 @@
 #include "hal.h"
 #include "mongoose.h"
 
-#define LED1 PIN('B', 0)      // On-board LED pin (green)
-#define LED2 PIN('B', 7)      // On-board LED pin (blue)
-#define LED3 PIN('B', 14)     // On-board LED pin (red)
-#define BTN1 PIN('C', 13)     // On-board user button
+#define LED PIN('B', 7)       // On-board LED pin
 #define BLINK_PERIOD_MS 1000  // LED blinking period in millis
 
-static uint64_t s_ticks, s_exti;  // Counters, increased by IRQ handlers
-
-uint64_t mg_millis(void) {  // Declare our own uptime function
-  return s_ticks;           // Return number of milliseconds since boot
-}
-
-void HardFault_Handler(void) {                // Escalated fault handler
-  gpio_output(LED3);                          // Setup red LED
-  for (;;) spin(2999999), gpio_toggle(LED3);  // Blink LED infinitely
-}
-
+static uint64_t s_ticks;      // Milliseconds since boot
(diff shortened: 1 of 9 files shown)


mip/mip.c
```diff
@@ -888,6 +896,14 @@ size_t mip_driver_rx(void *buf, size_t len, struct mip_if *ifp) {
 }
 
 void mip_init(struct mg_mgr *mgr, struct mip_if *ifp) {
+  // If MAC address is not set, make a random one
+  if (ifp->mac[0] == 0 && ifp->mac[1] == 0 && ifp->mac[2] == 0 &&
+      ifp->mac[3] == 0 && ifp->mac[4] == 0 && ifp->mac[5] == 0) {
+    mg_random(ifp->mac, sizeof(ifp->mac));
+    ifp->mac[0] &= (uint8_t) ~1;  // 1st byte must be even (unicast)
```
scaprile, line 903: How does this make it even ? There are other restrictions too; what about setting it to 02 that works for sure ?

mip/mip.c
```diff
@@ -901,6 +917,12 @@ void mip_init(struct mg_mgr *mgr, struct mip_if *ifp) {
     ifp->mgr = mgr;
     mgr->extraconnsize = sizeof(struct connstate);
     if (ifp->ip == 0) ifp->enable_dhcp_client = true;
+
+    // Randomise initial ephemeral port
+    uint16_t jitter;
+    mg_random(&jitter, sizeof(jitter));
+    ifp->eport = MIP_ETHEMERAL_PORT + (jitter % (0xffffu - MIP_ETHEMERAL_PORT));
+
```
scaprile, line 925: We should change ETHEMERAL to EPHEMERAL

examples/stm32/nucleo-f746zg-baremetal/main.c
```diff
@@ -4,79 +4,65 @@
...
 
-void SystemInit(void) {                  // Called automatically by startup code
-  clock_init();                          // Set clock to 180MHz
-  SysTick_Config(SYS_FREQUENCY / 1000);  // Increment s_ticks every ms
+void mg_random(void *buf, size_t len) {  // Use on-board RNG
+  for (size_t n = 0; n < len; n += sizeof(uint32_t)) {
+    uint32_t r = rng_read();
+    memcpy((char *) buf + n, &r, n + sizeof(r) > len ? len - n : sizeof(r));
+  }
 }
```
scaprile, line 24: This should perhaps be added to the driver, as all code using it will need it. Besides MG_ENABLE_CUSTOM_RANDOM we could resort to the driver's function when MG_ENABLE_MIP is defined; if the driver does not have one it falls back to Mongoose default.

mip/mip.c
scaprile: Yeap, this will work for the STMs but won't be very useful without some entropy on other platforms without an RNG

PR #2034, opened 2023-02-08
Add struct mg_queue and API, change examples/multi-threaded, add tests

cpq:


src/queue.c
```diff
@@ -0,0 +1,46 @@
...
+}
+
+void mg_qconsume(struct mg_queue *q) {
+  unsigned long tail, len = mg_qnext(q, &tail);
+  if (len > 0) {
+    tail += mg_align(len, sizeof(unsigned long));
+    if (tail >= q->head) tail = q->head;
+    q->tail = tail;
+  }
+}
```
scaprile, line 46: This is a bit convoluted... (for me) and it's late, so please excuse if I write nonsense

You ask for room. Then you pass that as a parameter. Once there is room, that won't change, it can only get better when the consumer does his thing; so, once there is room, return the amount, and q-> head is still the place to put that (if it fits); unless I'm not seeing something, there is no need to pass parameter head to qproduce()

q->tail should never be greater than q->head, at least equal, and that happens either when the queue is empty and when the queue is full, that is why some people leave one byte "reserved", to prevent the tail = head condition on queue full. With this as it is, you may purge a full queue if no msg has been consumed when it fills, as you see that as an empty queue

You push messages and the head advances, you pop messages and the tail advances, when head is at the end of the queue, even though the queue may be empty, you return only the room between head and end. Many frames will be dropped until one fits that size. You should reset head when there is no room for a whole frame AND tail is above that (head > len - MAX_PKT_SIZE -  sizeof(UL)) && (tail >= MAX_PKT_SIZE + sizeof(UL))

scaprile (commented):
Testing with a CRC is quite cool !

examples/multi-threaded/main.c
```diff
@@ -1,19 +1,26 @@
...
 // some time to simulate long processing time, produces an output and
 // hands over that output to the request handler function.
 //
+// We use a single queue for the bi-directional communication of the connection
+// and a worker thread. A connection starts a worker thread and writes a string
+// value (a name) into the queue. A worker thread reads a name from a queue,
+// and writes a string message "Hello, <name>" back.  A connection waits for
+// that message to appear, and sends it back to the client.
+//
+// The 1st byte of the message in the queue specifies the direction.
```
scaprile, line 15: This works because there is always only one outstanding message in the queue...

examples/multi-threaded/main.c
```diff
@@ -24,79 +31,73 @@ static void start_thread(void (*f)(void *), void *p) {
...
+  MG_INFO(("Got space %lu in the queue, offset %lu", len, ofs));
+  len = mg_snprintf((char *) q->buf + ofs, len, "%cHello %s", FROM_WORKER,
+                    name.ptr);  // Copy message, prepend direction marker
+  mg_qproduce(q, ofs, len);     // Signal that we've written a message
+  free((char *) name.ptr);      // Free stored name
+
+  // Wait until connection reads the message, then deallocate the queue
+  while ((len = mg_qnext(q, &ofs)) != MG_QUEUE_EMPTY) usleep(1000);
+  MG_INFO(("done, deallocating queue"));
+  free(q);
```
scaprile, line 61: I'd rather free() where I calloc(), what would a real life application for this ? Does it make sense that the worker waits for the master to have consumed the response to release the queue ? I would kill the worker task asap and let the master decide whether to reuse the queue or dealloc.

src/queue.c
```diff
@@ -0,0 +1,46 @@
...
+#define MG_MEMORY_BARRIER() asm("dsb; isb;")
+#else
+#define MG_MEMORY_BARRIER() (void) 0
+#endif
+
+// Align given value up
+static unsigned long mg_align(unsigned long value, unsigned long align) {
+  return (value + align - 1) / align * align;
+}
+
```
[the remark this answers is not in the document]
cpq: That is correct only for one case: when `align == sizeof(unsigned long)`. Which also must be power of 2.
My formula works for any `align`. For example,  if `align == 7:`: `mg_align(11, 7) == 14`

examples/multi-threaded/main.c
cpq: Yes. That works only in the "dialogue" mode.

src/queue.c
cpq: I am not sure you're correct. `you return only the room between head and end` - that's wrong.

When tail == head, it returns the full queue length:
```c
unsigned long mg_qspace(struct mg_queue *q, unsigned long *ofs) {
  *ofs = q->tail >= q->head ? sizeof(unsigned long)
                            : q->head + sizeof(unsigned long);
  return *ofs > q->len ? 0 : q->len - *ofs;
}
```

Anyway, what's your suggested API?

src/queue.c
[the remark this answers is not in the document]
scaprile: Yes, of course, but you are only aligning to `unsigned long` and that I guess is because of the alignment needed for some DMA controllers. I fail to imagine why you would align to 7 or other non-power-of-2 number

examples/multi-threaded/main.c
scaprile: I would add a note to avoid github issues accumulating

examples/multi-threaded/main.c
scaprile: Got it

src/queue.c
```diff
@@ -0,0 +1,40 @@
...
+//
+//    |------| ML | message1 |  ML | message2 |--- free space ---|
+//    ^      ^                                ^                  ^
+//   buf    tail                             head               len
+
+size_t mg_queue_space(struct mg_queue *q, char **buf) {
+  size_t ofs;
+  if (q->head > 0 && q->tail >= q->head) {  // All messages read?
+    q->head = 0;                            // Yes. Reset head first
+    q->tail = 0;                            // Now reset the tail
```
scaprile, line 15: This one is the atomic bomb, I can see clearly now (the rain is gone)

scaprile (commented):
Printing to the queue is _so_ cool !

PR #2037, opened 2023-02-08
Randseed

jfsimon1981:
mg_entropy() gathers from 3 sources and seeds mg_random().
Tested ok with hot restart (reset button) and usage before mif_init()
hence can provide enough entropy to get randomized mac.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -1,5 +1,33 @@
 #include "util.h"
 
+#define MG_ENTROPY_STACK_S 512
+#define MG_ENTROPY_HEAP_S 2048
+
+uint64_t mg_entropy(void);
+uint64_t mg_entropy() {
+	static uint64_t entropy;
+	const unsigned int lbss = MG_ENTROPY_STACK_S;  // Size of a large buffer of unint64_t on stack
+	const unsigned int lbhs = MG_ENTROPY_HEAP_S; // Size of a large buffer of unint64_t on heap
+	// Mix from source 1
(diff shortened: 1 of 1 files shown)


src/util.c
```diff
@@ -15,6 +16,14 @@ void mg_random(void *buf, size_t len) {
...
   }
+#else
+  static uint64_t millis_last;
+  volatile uint64_t millis = mg_millis();
+  // Seed/reseed on timout
+  if (!millis_last || ((millis - millis_last) > MG_RESEED_PERIOD_MS)) {
+    srand((unsigned int)millis);
+    millis_last = millis;
+  }
 #endif
```
scaprile, line 27: My belief is that this will re-seed on call after a timeout.
So far we have two issues: the most important is getting a random ephemeral port on start, to avoid a known "flaw", the other is to generate a random MAC address on startup. That is, we just use it on startup.
What we need is a random seed, like e.g.: re-seeding on DHCP response would solve the ephemeral ports issue, but not the MAC issue.

jfsimon1981: Correct, as discussed yesterday, even phy auto-negociation can be very predictable on some hardware and wont's solve it. This needs more investigation.

jfsimon1981: Updated per discuss, reading heap/stack works, mixing timer makes it better.
That works on the NXP before network access, whereas mg_millis() was too predictable.
Thanks for the idea that's the draft implementation of it.

PR #2038, opened 2023-02-08
Use CMSIS for H7

cpq:


examples/stm32/nucleo-h743zi-baremetal/hal.h
```diff
@@ -0,0 +1,148 @@
...
+  D2PPRE1 = 4,
+  D2PPRE2 = 4,
+  D3PPRE = 4
+};
+// PLL1_P: odd division factors are not allowed (8.7.13) (according to Cube, '1'
+// is also an "odd division factor").
+// Make sure your chip is revision 'V', otherwise set PLL1_N = 400
+enum { PLL1_HSI = 64, PLL1_M = 32, PLL1_N = 480, PLL1_P = 2 };
+#define FLASH_LATENCY 0x24  // WRHIGHFREQ LATENCY
+#define SYS_FREQUENCY ((PLL1_HSI * PLL1_N / PLL1_M / PLL1_P / D1CPRE) * 1000000)
```
scaprile, line 41: There is no "sys" in H7, that is why I used "cpu".
There are three domains each one with its AHB clock and the CPU has a different clock. There is also a prescaler between the source (which is called "sys" in the datasheets) and the cpu, so using "sys" is misleading.

examples/stm32/nucleo-h743zi-baremetal/main.c
```diff
@@ -1,67 +1,74 @@
...
-  clock_init();
-  systick_init(CPU_FREQUENCY / 1000);  // Increment s_ticks every ms
-  gpio_output(LED2);                   // Setup LED
-  uart_init(UART_DEBUG, 115200);       // Initialise debug printf
+uint64_t mg_millis(void) {  // Let Mongoose use our uptime function
+  return s_ticks;           // Return number of milliseconds since boot
+}
 
-  char rev = chiprev();
-  MG_INFO(("Chip revision: %c, max cpu clock: %u MHz", rev, (rev == 'V') ? 480 : 400));
```
scaprile, line 29: You need to know the chip revision, revY does not run at 480MHz but at 400MHz, that is why I put that there

examples/stm32/nucleo-h743zi-baremetal/main.c
```diff
@@ -1,67 +1,74 @@
...
+           names[ifp->state], mg_print_ip4, &ifp->ip, ifp->nrecv, ifp->nsent,
+           ifp->ndrop, ifp->nerr));
+}
 
-  // Initialise Ethernet. Enable MAC GPIO pins, see UM2407 for MB1364 boards
-  // section 6.6.7:
-  // https://www.st.com/resource/en/user_manual/um2407-stm32h7-nucleo144-boards-mb1364-stmicroelectronics.pdf
+static void ethernet_init(void) {
+  // Initialise Ethernet. Enable MAC GPIO pins, see
+  // https://www.farnell.com/datasheets/2014265.pdf section 6.10
```
scaprile, line 41: I'd rather send the reader to ST than to Farnell as an authoritative source for the datasheet
How come sections are different ?

examples/stm32/nucleo-h743zi-baremetal/main.c
```diff
@@ -1,67 +1,74 @@
...
-              GPIO_PULL_NONE, 11);
+              GPIO_PULL_NONE, 11);  // 11 is the Ethernet function
   }
-  nvic_enable_irq(61);                          // Setup Ethernet IRQ handler
-  RCC->APB4ENR |= BIT(1);                       // Enable SYSCFG
+  NVIC_EnableIRQ(ETH_IRQn);                     // Setup Ethernet IRQ handler
   SETBITS(SYSCFG->PMCR, 7 << 21, 4 << 21);      // Use RMII (12.3.1)
   RCC->AHB1ENR |= BIT(15) | BIT(16) | BIT(17);  // Enable Ethernet clocks
-  RCC->AHB1RSTR |= BIT(15);                     // ETHMAC force reset
-  RCC->AHB1RSTR &= ~BIT(15);                    // ETHMAC release reset
```
scaprile, line 47: We used to reset the MAC controller

examples/stm32/nucleo-h743zi-baremetal/hal.h
cpq: Got it - reverted to CPU_FREQUENCY

examples/stm32/nucleo-h743zi-baremetal/main.c
cpq: Don't we show the CPU clock here?

```c
  MG_INFO(("Starting, CPU freq %g MHz", (double) SystemCoreClock / 1000000));
```

If the SystemCoreClock holds a wrong number, then it is SystemInit() that should be fixed insted.

examples/stm32/nucleo-h743zi-baremetal/main.c
cpq: Restored! Was a copy-paste error.

examples/stm32/nucleo-h743zi-baremetal/main.c
cpq: Do we need that ? Isn't MAC controller reset after a reboot?

examples/stm32/nucleo-h743zi-baremetal/main.c
scaprile: I mean, if the chip does some weirdo crashes, the user will see a nice message before that... of if it does not boot, a clue to search why.

The function does not change freq according to chip revno, as the PLL numbers are fixed enums decided at compile time.

examples/stm32/nucleo-h743zi-baremetal/main.c
scaprile: Yeah, I thought that, but was too lazy to go check if that was the 429 datasheet...

examples/stm32/nucleo-h743zi-baremetal/main.c
scaprile: Yes, it should, but we've been doing it, I thought it was for consistency. No resets from now on

examples/stm32/nucleo-h743zi-baremetal/main.c
cpq: Restored!

examples/stm32/nucleo-h743zi-baremetal/hal.h
```diff
@@ -0,0 +1,167 @@
...
+}
+
+static inline char chiprev(void) {
+  uint16_t rev = (uint16_t) (((uint32_t) DBGMCU->IDCODE) >> 16);
+  if (rev == 0x1003) return 'Y';
+  if (rev == 0x2003) return 'V';
+  return '?';
+}
+
+#define UUID ((uint32_t *) 0x1ff1e800)  // Unique 96-bit chip ID. TRM 61.1
```
scaprile, line 157:
```
/*!< Device electronic signature memory map */
#define UID_BASE                  (0x1FF1E800UL)            /*!< Unique device ID register base address */
```
From CMSIS include file...

cpq: Done!

PR #2042, opened 2023-02-09
Use linux ephemeral range (start from 32768). More TCP debug

cpq:


scaprile:
> The RFC 6056 says that the range for ephemeral ports should be 1024–65535.[2]
The Internet Assigned Numbers Authority (IANA) and RFC 6335 suggests the range 49152–65535 (2^15 + 2^14 to 2^16 − 1) for dynamic or private ports.[3][4]
Many Linux kernels use the port range 32768–60999.[note 1][5] FreeBSD has used the IANA port range since release 4.6. Previous versions, including the Berkeley Software Distribution (BSD), use ports 1024–5000 as ephemeral ports.[6][7]

PR #2066, opened 2023-02-14
Refactor queue

cpq:


src/queue.c
```diff
@@ -1,40 +1,82 @@
...
+  if (q->head >= q->tail && q->head + len + hs <= q->size) {
+    space = q->size - q->head - hs;  // There is enough space
+  } else if (q->head >= q->tail && q->tail > hs) {
+    mg_queue_write_len(q, 0);  // Not enough space ahead
+    q->head = 0;               // Wrap head to the beginning
   }
-  ofs = q->head + sizeof(size_t);
-  if (buf != NULL) *buf = q->buf + ofs;
-  return ofs > q->len ? 0 : q->len - ofs;
+  if (q->head + hs + len < q->tail) space = q->tail - q->head - hs;
```
scaprile, line 53: `else if` ?

scaprile (approved):
Exquisite !
Ottimo !

src/queue.c
cpq: Nope! On line 51, we wrap head, which can make the condition on line 53 become true.
If we make `else if`, we won't hit that path.

PR #2067, opened 2023-02-14
Use RNG in examples/pico-w5500

cpq:


scaprile:
It is `MG_ENABLE_CUSTOM_RANDOM`
But... since we have the RP2040 ARCH, I would just put this inside mg_random, what do you think ? (Same as Linux and ESP32)

cpq:
Makes total sense to move this into the RP2040 arch, and demand `pico_rand` library to be included.
Then every RP2040 will be affected - to include a lib.
TBD in subsequent PRs.

PR #2075, opened 2023-02-17
adding time_spec to mg_http_serve_opts for choosing how to display last modified date

behnejad:
adding time_spec to mg_http_serve_opts for choosing how to display last modified date.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -562,19 +562,11 @@ static void printdirentry(const char *name, void *userdata) {
     } else {
       mg_snprintf(sz, sizeof(sz), "%lld", (uint64_t) size);
     }
-    if (d->opts->time_spec == MG_TIMESPEC_LOCAL) {
-      char time_str[30];
-      struct tm *time_info = localtime(&t);
-      strftime(time_str, sizeof time_str, "%Y/%m/%d %H:%M:%S", time_info);
-      mg_snprintf(mod, sizeof(mod), "%s", time_str);
-    } else if (d->opts->time_spec == MG_TIMESPEC_UTC) {
-      char time_str[30];
-      struct tm *time_info = gmtime(&t);
-      strftime(time_str, sizeof time_str, "%Y/%m/%d %H:%M:%S", time_info);
-      mg_snprintf(mod, sizeof(mod), "%s", time_str);
-    } else {
-      mg_snprintf(mod, sizeof(mod), "%ld", (unsigned long) t);
-    }
+
+    char time_str[100];
+    struct tm *time_info = gmtime(&t);
+    strftime(time_str, sizeof time_str, MG_HTTP_DIRLIST_TIME_FMT, time_info);
+    mg_snprintf(mod, sizeof(mod), "%s", time_str);
 
     n = (int) mg_url_encode(name, strlen(name), path, sizeof(path));
     mg_printf(d->c,
--- a/src/http.c
+++ b/src/http.c
@@ -562,23 +562,20 @@ static void printdirentry(const char *name, void *userdata) {
     } else {
       mg_snprintf(sz, sizeof(sz), "%lld", (uint64_t) size);
     }
-#if defined(MG_HTTP_DIRLIST_TIME)
-    {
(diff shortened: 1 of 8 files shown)


scaprile:
Since they're mutually exclusive, I vote for having a single `MG_HTTP_DIRLIST_TIME` with possible values
```
enum {
    MG_TIMESPEC_OFF = 0 ,
    MG_TIMESPEC_LOCAL,
    MG_TIMESPEC_UTC
}
```
That can also be used in other places if and when necessary

cpq:
why not strftime format string? When NULL, we'll use a default.

```c
  const char *time_fmt;
```

src/http.c
```diff
@@ -562,23 +562,20 @@ static void printdirentry(const char *name, void *userdata) {
...
-      char time_str[30];
+    } else if (d->opts->time_spec == MG_TIMESPEC_UTC) {
+      char time_str[MG_HTTP_TIME_FMT_SIZE];
       struct tm *time_info = gmtime(&t);
-      strftime(time_str, sizeof time_str, "%Y/%m/%d %H:%M:%S", time_info);
+      strftime(time_str, sizeof time_str, d->opts->time_fmt ? d->opts->time_fmt : "%Y/%m/%d %H:%M:%S", time_info);
       mg_snprintf(mod, sizeof(mod), "%s", time_str);
+    } else {
+      mg_snprintf(mod, sizeof(mod), "%ld", (unsigned long) t);
     }
```
scaprile, line 577: Perhaps
```
if (d->opts->time_spec == MG_TIMESPEC_OFF) {
  mg_snprintf(mod, sizeof(mod), "%ld", (unsigned long) t);
} else {
  char time_str[MG_HTTP_TIME_FMT_SIZE];
  struct tm *time_info = (d->opts->time_spec == MG_TIMESPEC_LOCAL) ? localtime(&t) : gmtime(&t);
  strftime(time_str, sizeof(time_str), d->opts->time_fmt ? d->opts->time_fmt : "%Y/%m/%d %H:%M:%S", time_info);
  mg_snprintf(mod, sizeof(mod), "%s", time_str);
}
```
?

scaprile (commented):
Looks nice

cpq:
Please get rid of everything that is not necessary. MG_TIMESPEC_ENUM should go.  MG_HTTP_TIME_FMT_SIZE should go.
Make a buffer 100 bytes long, do not introduce yet another tunable - we have enough of them. `time_fmt` is the only thing that should be introduced by this PR into the API, nothing else.

scaprile:
How do you suggest to choose between seconds, formatted local, and formatted UTC without the enum ?

behnejad:
I think displaying time in seconds is not necessary, since we want it to be human readable so MG_TIMESPEC_OFF is not needed.
gmtime and localtime can be chosen at compile time or hard coded.
Time format can also be hard coded or chosen at compile time or defined dynamically by developer with default format (personally prefer).
It is possible to let user set time function with registering a callback, if it is not set, localtime will be default callback.

cpq:
I like the idea of the compile-time format option!
A dynamic format, IMO, is an overkill.
Then, we could have a build option `MG_DIR_LIST_TIME_FMT` set to "%Y/%m/%d %H:%M:%S" by default?

Or maybe we just don't do any of that, and use time format that is traditionally used by other servers? @behnejad what exactly made you do this PR? What format you prefer to see?

behnejad:
I like to use "%Y/%m/%d %H:%M:%S" format with localtime and by seeing other webservers, they have hard coded format.
Before this PR, I opened another one to add time format, by the time i though it is good to have an option to choose which time (local / gm) to use. But now i think it is overkill like you said.
How about having a macro for time format and use localtime only without any options?

scaprile:
> I think displaying time in seconds is not necessary, since we want it to be human readable so MG_TIMESPEC_OFF is not needed.

We actually have systems that don't have the notion of that, only uptime is available. Why would those systems want to show a directory is another question, but if possible I would like to keep the unformatted plain uptime option, as formatting that may send us back to the loved 70's and I'm too old to rock'n'roll.

PR #2076, opened 2023-02-19
Use mg_queue in mip

cpq:


src/tcpip/tcpip.c
```diff
@@ -790,9 +746,17 @@ static void mg_tcpip_poll(struct mg_tcpip_if *ifp, uint64_t uptime_ms) {
...
+    size_t len =
+        ifp->driver->rx(ifp->recv_queue.buf, ifp->recv_queue.size, ifp);
+    if (len > 0) mg_tcpip_rx(ifp, ifp->recv_queue.buf, len);
+  } else {  // Interrupt-based driver. Fills recv queue itself
+    char *buf;
+    size_t len = mg_queue_next(&ifp->recv_queue, &buf);
+    if (len > 0) {
+      mg_tcpip_rx(ifp, buf, len);
+      mg_queue_del(&ifp->recv_queue, len);
+    }
```
scaprile, line 759: We agreed on providing a single entry point and always read from rx(); the driver would pull from the queue if it pushed to it (IRQ drivers and RP2040 RMII example)
When introducing the RP2040 RMII example, we decided to change the API as the new one was more homogeneous, flexible, and consistent

src/tcpip/tcpip.c
```diff
@@ -823,20 +787,20 @@ static void mg_tcpip_poll(struct mg_tcpip_if *ifp, uint64_t uptime_ms) {
...
 void mg_tcpip_qwrite(void *buf, size_t len, struct mg_tcpip_if *ifp) {
-  if (q_write(&ifp->queue, buf, len)) {
+  char *p;
+  if (mg_queue_book(&ifp->recv_queue, &p, len) >= len) {
+    memcpy(p, buf, len);
+    mg_queue_add(&ifp->recv_queue, len);
     ifp->nrecv++;
   } else {
     ifp->ndrop++;
+    MG_ERROR(("Recv queue full: %lu/%lu", len, ifp->recv_queue.size));
```
scaprile, line 797: This is IRQ code, I think incrementing drop counter is enough.

src/tcpip/tcpip.c
```diff
@@ -823,20 +787,20 @@ static void mg_tcpip_poll(struct mg_tcpip_if *ifp, uint64_t uptime_ms) {
...
-size_t mg_tcpip_qread(void *buf, struct mg_tcpip_if *ifp) {
-  return q_read(&ifp->queue, buf);
-}
-
 size_t mg_tcpip_driver_rx(void *buf, size_t len, struct mg_tcpip_if *ifp) {
-  return mg_tcpip_qread((void *) ifp->rx.ptr, ifp);
-  (void) len, (void) buf;
+  (void) buf, (void) len, (void) ifp;
+  return 0;
 }
```
scaprile, line 804: This allows drivers to just put a pointer to this function in the function structure, a generic poll from queue function that works for all IRQ based drivers, and avoid implementing that themselves.

src/tcpip/tcpip.c
```diff
@@ -851,10 +815,10 @@ void mg_tcpip_init(struct mg_mgr *mgr, struct mg_tcpip_if *ifp) {
...
     MG_ERROR(("driver init failed"));
   } else {
-    size_t maxpktsize = 1540;
-    ifp->rx.ptr = (char *) calloc(1, maxpktsize), ifp->rx.len = maxpktsize;
-    ifp->tx.ptr = (char *) calloc(1, maxpktsize), ifp->tx.len = maxpktsize;
-    if (ifp->queue.len) ifp->queue.buf = (uint8_t *) calloc(1, ifp->queue.len);
+    size_t framesize = 1540;
+    ifp->tx.ptr = (char *) calloc(1, framesize), ifp->tx.len = framesize;
+    ifp->recv_queue.size = ifp->driver->rx ? framesize : 8192;
+    ifp->recv_queue.buf = (char *) calloc(1, ifp->recv_queue.size);
```
scaprile, line 821: We've been there, and opted for doing that in the driver, because some drivers don't need a queue and all had the rx() function.

test/fuzz.c
```diff
@@ -81,8 +81,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
     mg_tcpip_rx(&mif, pkt, size);
     mg_mgr_free(&mgr);
     free(pkt);
-    free((char *) mif.rx.ptr);
-    free((char *) mif.tx.ptr);
```
scaprile, line 85: No free(tx.ptr) ?

src/tcpip/driver_stm32.c
```diff
@@ -212,7 +211,7 @@ void ETH_IRQHandler(void) {
   ETH->DMARPDR = 0;     // and resume RX
 }
 
-struct mg_tcpip_driver mg_tcpip_driver_stm32 = {
-    mg_tcpip_driver_stm32_init, mg_tcpip_driver_stm32_tx, mg_tcpip_driver_rx,
-    mg_tcpip_driver_stm32_up};
```
scaprile, line 217: No changes needed if we keep the current API.
You did not take into account the RP2040 RMII example, then we decided to change the API as the new one was more homogeneous, flexible, and consistent.

src/tcpip/tcpip.c
scaprile: This is done by the IRQ-driven drivers using the queue by calling the function removed above. This way, IRQ or no IRQ, queue or no queue, the API stays the same, we call the driver rx() function.
Plus, drivers that don't use the queue (RP2040 WiFi IIRC) don't waste memory by having a queue they don't use.

scaprile (changes requested):

src/tcpip/tcpip.c
scaprile: OK, this one will go as soon as we fix the RMII example...

test/fuzz.c
cpq: mg_tcpip_free() does it

src/tcpip/tcpip.c
cpq: Agreed, removed

scaprile (commented):
AFAICS, we need to find a way for the RMII example and then get rid of mg_tcpip_driver_rx()
Another thumbs up point of this to be removed function was that it allowed calling other handlers like TinyUSB's and the Pico-W WiFi driver's. Nothing that can't be done as it is done now adding those calls after the Mongoose mgr call in the main loop...

scaprile (approved):

PR #2095, opened 2023-02-27
Fix RNDIS examples

scaprile:
workflow in #2076 did not detect build problems. Fixed in #2094 , tested here


scaprile:
RMII example just builds, qread function has been removed so it can't run. Will be fixed in next PR

Issue #1605, opened 2022-06-28
[Windows] 'unit_test.exe' failure


gvanem:
The `unit_test.exe` program always fails for me.
Running as `set V=5 & unit_test.exe` I get this output:
```
-> [9876543210,7]
...
3cab2d23 3 sock.c:301:write_conn        4 0x2b8 29:0 29 err 0
3cab2d33 4 sock.c:608:mg_mgr_poll       1 -- tchrc
3cab2d43 4 sock.c:608:mg_mgr_poll       5 r- tchrc
3cab2d52 3 sock.c:291:read_conn         5 0x2b4 0:0 29 err 0
3cab2d62 4 http.c:660:uri_to_path2      5 /foo/version.h -> ./src/version.h 3
3cab2d81 4 sock.c:608:mg_mgr_poll       4 -- tchrc
3cab2d91 4 sock.c:608:mg_mgr_poll       1 -- tchrc
3cab2da0 4 sock.c:608:mg_mgr_poll       5 -w tchrc
3cab2db0 3 sock.c:301:write_conn        5 0x2b4 129:0 129 err 0
3cab2dcf 4 sock.c:608:mg_mgr_poll       4 -- tchrc
3cab2def 4 sock.c:608:mg_mgr_poll       1 -- tchrc
3cab2dfe 4 sock.c:608:mg_mgr_poll       5 -- tchrc
3cab2dfe 4 sock.c:608:mg_mgr_poll       4 r- tchrc
3cab2dfe 3 sock.c:291:read_conn         4 0x2b8 0:0 129 err 0
3cab2dfe 3 net.c:163:mg_close_conn      4 closed
3cab2e0e 4 sock.c:608:mg_mgr_poll       1 -- tchrc
3cab2e0e 4 sock.c:608:mg_mgr_poll       5 r- tchrc
3cab2e0e 3 sock.c:291:read_conn         5 0x2b4 0:0 -1 err 0
3cab2e0e 3 net.c:163:mg_close_conn      5 closed
3cab2e0e 4 sock.c:608:mg_mgr_poll       1 -- tchrc
FAILURE test/unit_test.c:2270: cmpbody(buf, expected) == 0
```

Seems the cause is DOS line-ending of `src/version.h`. If I first do a `dos2unix src\version.h`, it passes.
Can the unit_test be suited for this case? Or are we expected to have git always convert files to Unix line-endings?
`core.autocrlf=false` or something?

[...]

PR #2102, opened 2023-03-02
Fix unit test under (actual) Windows

scaprile:
#1605

### DIFF
--- a/test/unit_test.c
+++ b/test/unit_test.c
@@ -2770,6 +2770,9 @@ int main(void) {
   const char *debug_level = getenv("V");
   if (debug_level == NULL) debug_level = "3";
   mg_log_set(atoi(debug_level));
+#if MG_ARCH == MG_ARCH_WIN32
+  system ("dos2unix.exe --quiet src/version.h");
+#endif
 
   test_json();
   test_queue();


cpq:
I don't like this fix.
It changes a source file under us, it just should be done like this.

Is the underlying issue with Git changing line endings on windows?
Then, maybe we can use another file to test against, not version.h - maybe some generated binary file, which is not touched by Git.

scaprile:
Agreed, I didn't like this either but I tried to insert a `\r` instead and it broke on our tests (I think because git actually works on Linux, only the test works on wine).
Yes, I also think a binary file is the proper solution, we could try to register the version file as "binary" (-b in CVS...) so it is not changed by git clients, but a specific binary file will be future proof and less human error prone.
I'll change this on Monday

cpq:
Yeah, we can make some dummy file in the test/.. dir - or use an existing one, which Git does not mess up with

scaprile:
The whole point of the test is to check rewrites: 2392 -> `opts.root_dir = "./test/data,/foo=./src";`
The test in line 2406 should also fail as git should also be modifying a.txt unless it has been marked as binary somehow...
The failing test in line 2408 intentionally tries to fetch from src to test rewriting.
https://github.com/cesanta/mongoose/blob/f68d83dbf065b07d1d7b5c222562b52a705a6e3e/test/unit_test.c#L2387-L2414

scaprile:
Exactly.
The whole Makefile does not expect to be run under Windows, neither do unit tests. Manually trimming the build process and running on Windows fails on line 2406 as expected; perhaps in many other places.

PR #2110, opened 2023-03-08
Gen MAC from board ID, standardize

scaprile:
Added MAC generation from board_id (which comes from the board 's flash ID)
Standardized example texts and structure


cpq (approved):
This looks good to me!

PR #2125, opened 2023-03-17
Treat ms=-1 as indefinite timeout in mg_mgr_poll

strager:
If mg_mgr_poll is called with the 'ms' parameter set to '-1', the behavior differs between operating systems:

* FreeRTOS: (I don't know)
* epoll (Linux): wait indefinitely
* poll (macOS, BSDs): wait indefinitely
* select (Windows, other OSs): wait for -1000 microseconds (???)

Change the behavior of mg_mgr_poll(mgr, -1) on Windows (and some other operating systems) to instead block indefinitely, matching the behavior of POSIX platforms.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -4668,12 +4668,18 @@ static void mg_iotest(struct mg_mgr *mgr, int ms) {
     }
   }
 #else
-  struct timeval tv = {ms / 1000, (ms % 1000) * 1000}, tv_zero = {0, 0};
+  struct timeval tv_buffer, tv_zero = {0, 0};
+  struct timeval *tv = NULL;
   struct mg_connection *c;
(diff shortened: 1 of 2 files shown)


scaprile:
I don't think we'd like this, as it will break timers.

strager:
> I don't think we'd like this, as it will break timers.

This problem only applies if the user uses timers in the first place, right?

The [documentation of mg_timer_add](http://mongoose.ws/documentation/#mg_timer_add) states the following:

> NOTE: Make sure that the timer interval is equal or more to the mg_mgr_poll() timeout.

If mg_timer_add is never called, then this constraint on mg_mgr_poll's timeout doesn't apply.

If Mongoose uses timers internally, I can see my patch being an issue.

Issue #1520, opened 2022-04-08
Manager state not fully contained in struct mg_mgr


sebhub:
It should be possible to use multiple struct mg_mgr structures in a multi-threaded application.

This doesn't work since the timer maintenance is global (g_timers, oldnow). It should move to struct mg_mgr.

The logging support uses global variables (s_fn, s_spec, s_fn_param). They should move to struct mg_mgr.

The mg_fs_packed and mg_fs_posix objects should be made read-only.

The active DNS requests should move to struct mg_mgr (s_reqs).

hex, sort_js_code, and sort_js_code2 should be made const (for example: static const char sort_js_code[]).

The MQTT s_id counter should probably move to struct mg_mgr.

I suggest to use `nm` tool to review that there are no read-write global objects in the mongoose.o module.

Issue #1266, opened 2021-04-16
Feature Request: mg_http_reply() response reasons


oamc:
`  char *reason = "OK";

  switch (code) {
    case 100: reason = "Continue"; break;
    case 101: reason = "Switching Protocols"; break;
    case 102: reason = "Processing"; break;
    case 200: reason = "OK"; break;
    case 201: reason = "Created"; break;
    case 202: reason = "Accepted"; break;
    case 203: reason = "Non-authoritative Information"; break;
    case 204: reason = "No Content"; break;
    case 205: reason = "Reset Content"; break;
    case 206: reason = "Partial Content"; break;
    case 207: reason = "Multi-Status"; break;
    case 208: reason = "Already Reported"; break;
    case 226: reason = "IM Used"; break;
    case 300: reason = "Multiple Choices"; break;
    case 301: reason = "Moved Permanently"; break;
    case 302: reason = "Found"; break;
    case 303: reason = "See Other"; break;
    case 304: reason = "Not Modified"; break;
    case 305: reason = "Use Proxy"; break;
    case 307: reason = "Temporary Redirect"; break;
    case 308: reason = "Permanent Redirect"; break;
    case 400: reason = "Bad Request"; break;
    case 401: reason = "Unauthorized"; break;
    case 402: reason = "Payment Required"; break;
    case 403: reason = "Forbidden"; break;
    case 404: reason = "Not Found"; break;
    case 405: reason = "Method Not Allowed"; break;
    case 406: reason = "Not Acceptable"; break;
    case 407: reason = "Proxy Authentication Required"; break;
    case 408: reason = "Request Timeout"; break;
    case 409: reason = "Conflict"; break;
    case 410: reason = "Gone"; break;
    case 411: reason = "Length Required"; break;
    case 412: reason = "Precondition Failed"; break;
    case 413: reason = "Payload Too Large"; break;
    case 414: reason = "Request-URI Too Long"; break;
    case 415: reason = "Unsupported Media Type"; break;
    case 416: reason = "Requested Range Not Satisfiable"; break;
    case 417: reason = "Expectation Failed"; break;
    case 418: reason = "I'm a teapot"; break;
    case 421: reason = "Misdirected Request"; break;
    case 422: reason = "Unprocessable Entity"; break;
    case 423: reason = "Locked"; break;
    case 424: reason = "Failed Dependency"; break;
    case 426: reason = "Upgrade Required"; break;
    case 428: reason = "Precondition Required"; break;
[...]

PR #2142, opened 2023-03-29
Reintroduce missing HTTP status code strings

james-advmach:
These were added in 1115514 in response to #1266 but were then removed in ef44f90.

There's nothing to suggest why they were removed in the commit message or the discussion around #1520. Perhaps they were removed by accident?


cpq:
Closing this.
Appreciate your attention!
If you have a specific code(s) that you need covered in your product, please file a PR. That would be compelling enough to get into the mainstream.

scaprile:
> the reason they were removed is that they are useless. Those codes that are used by the library already, are in.
> Can you provide a **compelling evidence / use case**, why codes e.g. `412 Precondition Failed` or `410 Gone` might be required?

Why ? In which case ? Because of what ? What is the use case for it ?

> If we call `mg_http_reply` with, say, status code 410 ("Gone"), then it'll reply to the client with the wrong string. It'll say `HTTP/1.1 410 OK`, which is not right!

If you nevertheless have/want to use one of those codes, use [mg_printf](https://mongoose.ws/documentation/#mg_printf-mg_vprintf) instead:
```
 mg_printf(c, "HTTP/1.1 410 Gone\r\nContent-length: 0\r\n\r\n");
```
Wireshark says:
```
Transmission Control Protocol, Src Port: 8000, Dst Port: 39962, Seq: 1, Ack: 449, Len: 40
Hypertext Transfer Protocol
    HTTP/1.1 410 Gone\r\n
    Content-length: 0\r\n
    \r\n
    [HTTP response 1/1]
```
Wouldn't that work for you in such a case ?

scaprile:
That function does not return any value, here is the list of supported codes, you may add a 200 and change the default "OK" at your wish:
https://github.com/cesanta/mongoose/blob/5cebce3f4bf677727c1c2ee2db519c44a3452f42/src/http.c#L296-L315

cpq:
@rgilton if you want to use rarely used reason strings for 410 and others, create your own utility function.

I dont really see the value in those reason strings, other than bloating the codebase.

cpq:
@rgilton 

Reading https://datatracker.ietf.org/doc/html/rfc2616#section-6.1.1 

```
  The individual values of the numeric status codes defined for
   HTTP/1.1, and an example set of corresponding Reason-Phrase's, are
   presented below. The reason phrases listed here are only
   recommendations -- they MAY be replaced by local equivalents without
   affecting the protocol.
```

I read it as that the implementation (mongoose library) can use whatever reason phrase it pleases.

cpq:
@rgilton that sounds reasonable! Can you elaborate on your use case please, how exactly reason phrases interfere with your  program logic? Do you fetch reason phrases from the HTTP responses - and if yes, why?

@scaprile could you add a note to the https://mongoose.ws/documentation/#mg_http_reply docs please, stating that only a subset of reason phrases are unique. With a permalink to the [github code ](https://github.com/cesanta/mongoose/blob/5cebce3f4bf677727c1c2ee2db519c44a3452f42/src/http.c#L296-L315)

PR #2162, opened 2023-04-21
Check return values of calloc()

szsam:


cpq:
1. If you decide for a proper error handling, then you also need to handle the failure part - log the failure, let the user know, etc
2. That is too much for the demo app - that's why we do not do a proper error handling, to make examples emphasize the main point.

PR #2160, opened 2023-04-20
MQTT: support for user properties feature added

robertc2000:

PR #2163, opened 2023-04-24
Revert. Some tests not passing

scaprile:
Reverts cesanta/mongoose#2160
Windows and hw tests were not passing and we were red.


cpq:
I've pushed fixed so this revert is not required  - closing

PR #2170, opened 2023-04-27
Check return values of calloc() in mongoose.c

szsam:
This PR patches mongoose.c, which does not belong to a demo app.
(This PR is a part of #2162, which has been closed.)


scaprile:
Yeah, seen that portion in your former PR, there seems to be no check in the embedded filesystem, we'll take care of that.

PR #2197, opened 2023-05-16
Added check for negative value for HTTP Content-Length header

allanpark:
Added check for negative value for HTTP Content-Length. HTTP "Bad request"(400) will be sent as responce.


test/unit_test.c
```diff
@@ -790,6 +790,10 @@ static void test_http_server(void) {
   ASSERT(fetch(&mgr, buf, url, "GET /..ddot HTTP/1.0\n\n") == 301);
   ASSERT(fetch(&mgr, buf, url, "GET /..ddot/ HTTP/1.0\n\n") == 200);
   ASSERT(cmpbody(buf, "hi\n") == 0);
+  ASSERT(fetch(&mgr, buf, url, "GET /a.txt HTTP/1.0\n"
+        "Content-Length: -123\n\n") == 400);
+  ASSERT(fetch(&mgr, buf, url, "POST /a.txt HTTP/1.0\n"
+               "Content-Length: -123\n\n") == 400);
```
cpq, line 796: If server just aborts the connection, fetch() should return 0, IMO. That's what we should check for.

src/http.c
```diff
@@ -950,7 +953,9 @@ static void http_cb(struct mg_connection *c, int ev, void *evd, void *fnd) {
       bool next = false;
       int hlen = mg_http_parse((char *) c->recv.buf, c->recv.len, &hm);
       if (hlen < 0) {
-        mg_error(c, "HTTP parse:\n%.*s", (int) c->recv.len, c->recv.buf);
+        mg_printf(c, "HTTP/1.0 400 Bad Request\r\n\r\n");
+        MG_ERROR(("%lu HTTP malformed message", c->id));
+        mg_iobuf_del(&c->recv, 0, hm.message.len);
```
scaprile, line 958: I would keep the former message as close as possible.
We are no longer closing the connection as before, is this OK/better ?

scaprile (commented):
Please run `make test` and cancel it, so it just builds the amalgamated sources, and include `mongoose.c` in the PR

PR #2200, opened 2023-05-17
Add MG_ESC helper macro

cpq:


scaprile (approved):
That is cool, certainly worth updating all related tutorials !

PR #2205, opened 2023-05-19
improve negative Content-length handling

scaprile:
Handling numbers outside int64_t range requires overflow indication, as 0 can be valid in some circumstances
Modified `mg_to64()` and `mg_tou64()` to be passed a `bool *` and set it to true on error. A `NULL` can be passed when not interested.
Changed `mg_http_parse_headers()` to use `mg_to64()`, as all numbers of interest will be `>= 0`, we exit on error indication
Modified the number parsing to accomodate a closer to full range interval
Passed `NULL` in other invocations of `mg_to64()`
Added further checkings in `unit_test.c`, both for `mg_to64()`/`mg_tou64()` and for incorrect/unhandled `Content-length`s

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -1243,6 +1243,22 @@ struct mg_fs mg_fs_posix = {p_stat,  p_list, p_open,   p_close,  p_read,
 
 
 
+bool mg_to_size_t(struct mg_str str, size_t *val);
+bool mg_to_size_t(struct mg_str str, size_t *val) {
+  uint64_t result = 0, max = 1844674407370955160 /* (UINT64_MAX-9)/10 */;
+  size_t i = 0;
+  while (i < str.len && (str.ptr[i] == ' ' || str.ptr[i] == '\t')) i++;
+  if (i < str.len && str.ptr[i] == '-') return false;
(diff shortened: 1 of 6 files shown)


cpq:
`mg_to64()` is not used anywhere else, other than in http.c
I suggest to remote it from the API. Both mg_to64 and mg_tou64. Move mg_tou64 to http.c

I'd make it `bool toul(struct mg_str, size_t *)`, and change this

```c
    int64_t content_len = mg_to64(*cl);
    if(content_len < 0) return -1;
    hm->body.len = (size_t) content_len;
    hm->message.len = (size_t) req_len + hm->body.len;
```

to this:

```c
    if (to_size_t(*cl, &hm->body.len) == false) return -1;  // See #.....
    hm->message.len = (size_t) req_len + hm->body.len;
```

scaprile:
Removed signed to64 as it doesn't have a real use
Left new `to_size_t()` as global as it has to be tested
Reduced tests accordingly
Modified `getrange()` to also work on `size_t` and its caller to pass and work on `size_t`

scaprile:
Done, removed; thanks for letting us know.

PR #2223, opened 2023-05-30
Fix Secure

scaprile:
Confirmed working on hardware (HTTP), Linux and MIP-TAP (HTTP/HTTPS)

### DIFF
--- a/examples/device-dashboard/net.c
+++ b/examples/device-dashboard/net.c
@@ -123,18 +123,20 @@ static struct user *authenticate(struct mg_http_message *hm) {
 static void handle_login(struct mg_connection *c, struct user *u) {
   char cookie[256];
   mg_snprintf(cookie, sizeof(cookie),
-              "Set-Cookie: access_token=%s;Path=/;"
-              "HttpOnly;SameSite=Lax;Max-Age=%d\r\n",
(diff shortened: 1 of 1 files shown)


examples/device-dashboard/net.c
```diff
@@ -130,11 +130,13 @@ static void handle_login(struct mg_connection *c, struct user *u) {
...
-                "Secure; HttpOnly; Max-Age=0; \r\n",
-                "true\n");
+  char cookie[256];
+  mg_snprintf(cookie, sizeof(cookie),
+              "Set-Cookie: access_token=; Path=/; "
+              "Expires=Thu, 01 Jan 1970 00:00:00 UTC; "
+              "%sHttpOnly; Max-Age=0; \r\n",
+              c->is_tls ? "Secure; " : "");
+  mg_http_reply(c, 200, cookie, "true\n");
 }
```
cpq, line 140: DO this in `handle_login()` too.

scaprile: `Secure` wasn't there in `handle_login()` before, only in `handle_logout()`; could this have been the issue ?

cpq: It was not there because I've removed it at some stage, but forgot to remove in handle_logout

PR #2228, opened 2023-06-01
Add TLS=1 option

cpq:


examples/device-dashboard/net.c
```diff
@@ -72,6 +72,15 @@ static int event_next(int no, struct event *e) {
...
 
+// This is for newlib and TLS (mbedTLS)
+int _gettimeofday(struct timeval *tv, void *tz) {
+  uint64_t now = mg_millis() + s_boot_timestamp;
+  (void) tz;
+  tv->tv_sec = (time_t) (now / 1000);
+  tv->tv_usec = (unsigned long) ((now % 1000) * 1000);
+  return 0;
+}
+
```
scaprile, line 83: This might wreak havoc in other environments, besides, I'm not comfortable with overriding a library call
We've already have problems when overriding time()

examples/stm32/nucleo-f746zg-baremetal/Makefile
```diff
@@ -37,6 +37,15 @@ cmsis_core:     # ARM CMSIS core headers
...
 	git clone --depth 1 -b v1.2.8 https://github.com/STMicroelectronics/cmsis_device_f7 $@
+mbedtls:        # mbedTLS library
+	git clone --depth 1 -b v2.28.2 https://github.com/mbed-tls/mbedtls $@
+
+ifeq ($(TLS),1)
+CFLAGS += -DMG_ENABLE_MBEDTLS=1 -DMBEDTLS_CONFIG_FILE=\"mbedtls_config.h\"
+CFLAGS += -Wno-conversion -Imbedtls/include
+SOURCES += mbedtls/library/*.c
+firmware.elf: mbedtls
+endif
```
scaprile, line 48: We are instructing to use CFLAGS_EXTRA in OS examples. I would do the same here, though the user would just add:
`CFLAGS_EXTRA=-DMG_ENABLE_MBEDTLS=1` (with of course no linking stuff)
This will keep the examples the same in both OS and embedded worlds; we cant catch `CFLAGS_EXTRA` existence and being `==-DMG_ENABLE_MBEDTLS=1` to add extra flags/sources

examples/device-dashboard/Makefile
```diff
@@ -42,7 +42,6 @@ packed_fs.c: $(wildcard web_root/*) Makefile web_root/main.css web_root/bundle.j
 # Pull and build mbedTLS library. See https://mongoose.ws/tutorials/tls/#how-to-build for TLS build options
 mbedtls:
 	git clone --depth 1 -b v2.28.2 https://github.com/mbed-tls/mbedtls $@
-	$(MAKE) -C mbedtls/library
```
scaprile, line 45: This is actually needed to build with a local copy of the library, the third option in Linux/Mac, only one in Windows.
There is no other way to build in Windows, and that is one of the reasons for that section in the tutorial
We tried other ways and this one was the cleanest one running on all three major OSs.

scaprile (changes requested):

examples/device-dashboard/Makefile
```diff
@@ -9,13 +9,11 @@ CFLAGS = -W -Wall -Wextra -g -I.        # Build options
 CFLAGS_MONGOOSE +=
 
 ifeq ($(OS),Windows_NT)         # Windows settings. Assume MinGW compiler. To use VC: make CC=cl CFLAGS=/MD OUT=/Feprog.exe
-  PROG ?= example.exe           # Use .exe suffix for the binary
+  PROG = example.exe            # Use .exe suffix for the binary
```
scaprile, line 12: Besides not letting the user change the name, this implies changing all other tutorials so all are consistent and we don't diverge again having lots of different ways of doing the same
Is it really necessary ?

examples/device-dashboard/Makefile
```diff
@@ -39,10 +37,14 @@ packed_fs.c: $(wildcard web_root/*) Makefile web_root/main.css web_root/bundle.j
...
-# Pull and build mbedTLS library. See https://mongoose.ws/tutorials/tls/#how-to-build for TLS build options
 mbedtls:
 	git clone --depth 1 -b v2.28.2 https://github.com/mbed-tls/mbedtls $@
-	$(MAKE) -C mbedtls/library
+
+ifeq ($(TLS), mbedtls)
+CFLAGS += -DMG_ENABLE_MBEDTLS=1 -Wno-conversion -Imbedtls/include
+CFLAGS += -DMBEDTLS_CONFIG_FILE=\"mbedtls_config.h\" mbedtls/library/*.c
+$(PROG): mbedtls
+endif
```
scaprile, line 47: `ifxx ($(CFLAGS_EXTRA), MBEDTLS)` ( if CFLAGS_EXTRA contains MBEDTLS)
so `make clean all CFLAGS_EXTRA= -DMG_ENABLE_MBEDTLS=1`just works

examples/stm32/nucleo-f746zg-baremetal/Makefile
```diff
@@ -37,6 +37,14 @@ cmsis_core:     # ARM CMSIS core headers
...
 cmsis_f7:       # ST CMSIS headers for STM32F7 series
 	git clone --depth 1 -b v1.2.8 https://github.com/STMicroelectronics/cmsis_device_f7 $@
+mbedtls:        # mbedTLS library
+	git clone --depth 1 -b v2.28.2 https://github.com/mbed-tls/mbedtls $@
+
+ifeq ($(TLS), mbedtls)
+CFLAGS += -DMG_ENABLE_MBEDTLS=1 -Wno-conversion -Imbedtls/include
+CFLAGS += -DMBEDTLS_CONFIG_FILE=\"mbedtls_config.h\" mbedtls/library/*.c
+firmware.elf: mbedtls
+endif
```
scaprile, line 47: Same as above for the generic version

scaprile (changes requested):

examples/device-dashboard/Makefile
cpq: I don't like this - snooping into CFLAGS_EXTRA

examples/device-dashboard/net.c
cpq: This is not a library call. This one is with the underscore, it is a newlib "syscall"

examples/stm32/nucleo-f746zg-baremetal/Makefile
cpq: For the pre-built libs, using CFLAGS_EXTRA is probably fine. Cause this is really just CFLAGS_EXTRA

When we're pulling an extra repo and adding extra sources, it is not really extra CFLAGS anymore

examples/device-dashboard/net.c
scaprile: Yes, of course, we should put it in `syscalls.c`.
I'll propose it later

PR #2230, opened 2023-06-02
RL: fix accept() not looping on BSD_EWOULDBLOCK

scaprile:
https://www.keil.com/pack/doc/mw/Network/html/group__using__network__sockets__bsd__func.html
`accept()`:
> In blocking mode, which is enabled by default, this function waits for a connection request. In non blocking mode, you must call the accept function again if the error code BSD_EWOULDBLOCK is returned.


cpq:
@scaprile can you add this comment to the changed source code line please?

```
In blocking mode, which is enabled by default, this function waits for a connection request. In non blocking mode, you must call the accept function again if the error code BSD_EWOULDBLOCK is returned.
```

PR #2240, opened 2023-06-07
Renew dhcp lease

cpq:


src/tcpip/tcpip.c
```diff
@@ -176,10 +176,7 @@ static void onstatechange(struct mg_tcpip_if *ifp) {
   if (ifp->state == MG_TCPIP_STATE_READY) {
     MG_INFO(("READY, IP: %M", mg_print_ip4, &ifp->ip));
     MG_INFO(("       GW: %M", mg_print_ip4, &ifp->gw));
-    if (ifp->lease_expire > ifp->now) {
-      MG_INFO(
-          ("       Lease: %lld sec", (ifp->lease_expire - ifp->now) / 1000));
-    }
+    MG_INFO(("      MAC: %M", mg_print_mac, &ifp->mac));
```
scaprile, line 179: We are printing the MAC only on those examples where the user would not know it (we generate it).
This is not useful info when the user has set the MAC and makes us print it twice when it is.

src/tcpip/tcpip.c
```diff
@@ -347,11 +346,14 @@ static void rx_dhcp_client(struct mg_tcpip_if *ifp, struct pkt *pkt) {
...
       memcpy(&lease, p + 2, sizeof(lease));
-      ifp->lease_expire = ifp->now + mg_ntohl(lease) * 1000;
+      lease = mg_ntohl(lease);
+      ifp->lease_expire = ifp->now + lease * 1000;
+      MG_INFO(("Lease: %u sec (%lld)", lease, ifp->lease_expire / 1000));
     }
     p += p[1] + 2;
   }
   if (ip && mask && gw && ifp->ip == 0) {
+    MG_INFO(("OO"));
```
scaprile, line 356: I think this one is a leftover

scaprile (commented):
This should work when the boards are always on. Let's go for it !
I fail to see the problem when the board starts with new firmware and asks for an IP. It should be given a new one if it expired; and if given one that just expires in seconds, it will get a new one on next reboot.

src/tcpip/tcpip.c
cpq: It actually might be useful to see logs, to spot conflicts or outright bad MAC settings

src/tcpip/tcpip.c
cpq: It is !

PR #2250, opened 2023-06-09
optimized mg_addr structure

robertc2000:


test/unit_test.c
```diff
@@ -1887,12 +1887,20 @@ static void test_str(void) {
 
   {
     char buf[100];
-    struct mg_addr a = {mg_htons(3), mg_htonl(0x2000001), {1, 100, 33}, false};
+    struct mg_addr a;
```
cpq, line 1890: No need to add an addr and then  memcpy it.
Write IP address as is:

```c
struct mg_addr a = {{20, 0, 0, 1, ....}, mg_htons(3), false};
```

PR #2271, opened 2023-06-23
extend http status codes

Sir2B:
add 414, 503

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -1555,8 +1555,8 @@ static const char *mg_http_status_code_str(int status_code) {
     case 304: return "Not Modified";
     case 400: return "Bad Request";
     case 401: return "Unauthorized";
-    case 403: return "URI Too Long";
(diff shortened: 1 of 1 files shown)


scaprile:
#2142 Please read that PR thread and provides a compelling reason with an interesting use case for those.

mongoose.c
```diff
@@ -1555,11 +1555,13 @@ static const char *mg_http_status_code_str(int status_code) {
     case 304: return "Not Modified";
     case 400: return "Bad Request";
     case 401: return "Unauthorized";
-    case 403: return "Forbidden";
+    case 403: return "URI Too Long";
+    case 414: return "Forbidden";
```
cpq, line 1559: I think you've got these two mixed

Sir2B:
Ok I find it a little bit weird that I got a `414 OK` back, but I understand your reasons.

PR #2278, opened 2023-06-27
fix header truncation

### LINKED ISSUE #2275: Header Truncation
## Description
The headers can be truncated in the example HTTP server by sending a header with no name.

## Instruction to Reproduce
```console
# Download and build the server
$ git clone https://github.com/cesanta/mongoose && cd mongoose

# Start the server
$ make &

# Observe that the server times out because the Content-Length header is interpreted:
$ printf 'GET / HTTP/1.1\r\nContent-Length: 1\r\n\r\n' | nc localhost 8000

# Observe that the server does not time out, because it stops interpreting headers after the empty header:
$ printf 'GET / HTTP/1.1\r\n:\r\nContent-Length: 1\r\n\r\n' | nc localhost 8000
```

## Impact
See [this CVE in HAProxy](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-25725), which describes a very similar issue. When Mongoose is used to implement a proxy, this behavior should be a cause for similar concern.

## Proposed Solution:
Make the HTTP parser fail when it encounters a header with an empty name, as is suggested by the RFCs.

## Versions
Mongoose 7.10, Linux/musl

## Note
You requested that we make this a public GitHub issue.

scaprile:
fixes #2275

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -1431,22 +1431,24 @@ struct mg_str *mg_http_get_header(struct mg_http_message *h, const char *name) {
   return NULL;
 }
 
-static void mg_http_parse_headers(const char *s, const char *end,
+static bool mg_http_parse_headers(const char *s, const char *end,
                                   struct mg_http_header *h, int max_headers) {
   int i;
   for (i = 0; i < max_headers; i++) {
     struct mg_str k, v, tmp;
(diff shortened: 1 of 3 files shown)


cpq:
@scaprile unit test please!

PR #2284, opened 2023-06-30
enhanced path sanitizing

robertc2000:


src/str.c
```diff
@@ -195,12 +195,15 @@ char *mg_remove_double_dots(char *s) {
...
-        }
+        } else if (s[0] == '.' && s[1] == '.') {
+          if (s[2] == '\0' || s[2] == '/' || s[2] == '\\')
+            s += 2;
+          else 
+            break;
+          }
+          else {
+            break;
+          }
```
scaprile, line 206: Code-style: Whenever there is an `else` clause, use braces for both then and else paths. The else goes on the same line as the brace, short functions can be read without scrolling. Please run clang-format

Is there a reason I'm missing for the second `if` ? The `else` action is the same for both nested `if`s, why not keep the `&&` as before ?

If possible, I would add a test case in the unit tests, to make sure this works.

scaprile (changes requested):

src/http.c
```diff
@@ -726,11 +728,21 @@ static int uri_to_path2(struct mg_connection *c, struct mg_http_message *hm,
   size_t n = mg_snprintf(path, path_size, "%.*s", (int) dir.len, dir.ptr);
   if (n > path_size) n = path_size;
   path[path_size - 1] = '\0';
-  if (n + 2 < path_size) path[n++] = '/', path[n] = '\0';
+  if (n + 2 < path_size && path[n-1] != '/') path[n++] = '/', path[n] = '\0';
   mg_url_decode(hm->uri.ptr + url.len, hm->uri.len - url.len, path + n,
                 path_size - n, 0);
   path[path_size - 1] = '\0';  // Double-check
-  mg_remove_double_dots(path);
+  if (!mg_path_is_sane(path)) {
```
cpq, line 735: Please simplify this:

```c
if (!mg_path_is_sane(path)) {
  mg_http_reply(c, 418, "", "Invalid path");
  return -1;
}
```

src/http.c
```diff
@@ -726,11 +728,21 @@ static int uri_to_path2(struct mg_connection *c, struct mg_http_message *hm,
   size_t n = mg_snprintf(path, path_size, "%.*s", (int) dir.len, dir.ptr);
   if (n > path_size) n = path_size;
```
cpq, line 729: Fail here, too. There is no point truncating the path.

```c
if (n > path_size) {
  mg_http_reply(c, 418, "", "path is too big");
  return -1;
}
```

src/http.c
```diff
@@ -726,11 +728,21 @@ static int uri_to_path2(struct mg_connection *c, struct mg_http_message *hm,
...
   path[path_size - 1] = '\0';
-  if (n + 2 < path_size) path[n++] = '/', path[n] = '\0';
+  if (n + 2 < path_size && path[n-1] != '/') path[n++] = '/', path[n] = '\0';
```
cpq, line 731: Please add a comment here
```c
// Terminate root dir with /
```

cpq (changes requested):

examples/file-upload-multiple-posts/main.c
```diff
@@ -12,11 +12,13 @@ static void cb(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
     if (mg_http_match_uri(hm, "/upload")) {
       char path[80], name[64];
       mg_http_get_var(&hm->query, "name", name, sizeof(name));
+      mg_snprintf(path, sizeof(path), "/tmp/%s", name);
       if (name[0] == '\0') {
         mg_http_reply(c, 400, "", "%s", "name required");
+      } else if (!mg_path_is_sane(path)) {
+        mg_http_reply(c, 400, "", "%s", "invalid path");
```
scaprile, line 19: As per the comment below on http.c, shouldn't this one also be 418 ?

test/unit_test.c
```diff
@@ -643,11 +643,13 @@ static void eh1(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
     } else if (mg_http_match_uri(hm, "/upload")) {
       char path[80], name[64];
       mg_http_get_var(&hm->query, "name", name, sizeof(name));
+      mg_snprintf(path, sizeof(path), "./%s", name);
       if (name[0] == '\0') {
         mg_http_reply(c, 400, "", "%s", "name required");
+      } else if (!mg_path_is_sane(path)) {
+        mg_http_reply(c, 400, "", "%s", "invalid path");
```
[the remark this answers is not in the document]
robertc2000: Isn't 400 more relevant than 418?

test/unit_test.c
```diff
@@ -854,11 +856,11 @@ static void test_http_server(void) {
...
-  ASSERT(fetch(&mgr, buf, url, "GET /../fuzz.c HTTP/1.0\n\n") == 404);
-  ASSERT(fetch(&mgr, buf, url, "GET /.%%2e/fuzz.c HTTP/1.0\n\n") == 404);
-  ASSERT(fetch(&mgr, buf, url, "GET /.%%2e%%2ffuzz.c HTTP/1.0\n\n") == 404);
-  ASSERT(fetch(&mgr, buf, url, "GET /..%%2f%%20fuzz.c HTTP/1.0\n\n") == 404);
-  ASSERT(fetch(&mgr, buf, url, "GET /..%%2ffuzz.c%%20 HTTP/1.0\n\n") == 404);
+  ASSERT(fetch(&mgr, buf, url, "GET /../fuzz.c HTTP/1.0\n\n") == 400);
+  ASSERT(fetch(&mgr, buf, url, "GET /.%%2e/fuzz.c HTTP/1.0\n\n") == 400);
+  ASSERT(fetch(&mgr, buf, url, "GET /.%%2e%%2ffuzz.c HTTP/1.0\n\n") == 400);
+  ASSERT(fetch(&mgr, buf, url, "GET /..%%2f%%20fuzz.c HTTP/1.0\n\n") == 400);
+  ASSERT(fetch(&mgr, buf, url, "GET /..%%2ffuzz.c%%20 HTTP/1.0\n\n") == 400);
```
scaprile, line 863: 418 ?

test/unit_test.c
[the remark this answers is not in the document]
scaprile: If we consider "deceptive routing path", probably yes.
We are not a teapot anyway... and 403 is more appropriate for blocking.

PR #2294, opened 2023-07-07
Fix #2292 - fail on stray \r character in headers

### LINKED ISSUE #2292: Mongoose terminates HTTP headers improperly
From RFC 9112:
> Although the line terminator for the start-line and fields is the sequence CRLF, a recipient MAY recognize a single LF as a line terminator and ignore any preceding CR.
> A sender MUST NOT generate a bare CR (a CR character not immediately followed by LF) within any protocol elements other than the content. A recipient of such a bare CR MUST consider that element to be invalid or replace each bare CR with SP before processing the element or forwarding the message.

Mongoose terminates headers on bare CR, in violation of the standard. For example, the following request is parsed incorrectly:
```
GET / HTTP/1.1\r\n
Whatever: whatever\rContent-Length: 10\r\n
\r\n
0123456789
```

Mongoose interprets this message as having a Content-Length header, and consequently interprets the 0123456789 as being the request body. A standards-compliant HTTP server would either replace the '\r' with a space, and thus see a `Whatever: whatever Content-Length: 10` header, or reject the message and respond 400.

cpq:


src/http.c
```diff
@@ -234,9 +239,17 @@ int mg_http_parse(const char *s, size_t len, struct mg_http_message *hm) {
...
+  while (s < end && (n = clen(s)) > 0) s += n, hm->method.len += n;
+  while (s < end && s[0] == ' ') s++;  // Skip spaces
+  hm->uri.ptr = s;
+  while (s < end && (n = clen(s)) > 0) s += n, hm->uri.len += n;
+  while (s < end && s[0] == ' ') s++;  // Skip spaces
+  hm->proto.ptr = s;
+  while (s < end && s[0] != '\n' && s[0] != '\r') s++, hm->proto.len++;
+  if (s >= end || (s[0] == '\r' && s[1] != '\n')) return false;  // Stray \r
+  if (s < end && s[0] == '\r') s++;
+  if (s >= end || *s++ != '\n') return false;
```
scaprile, line 252: 251: Here `s < end`, otherwise it would have exit at line 250
252: We already checked for that, didn't we ? This looks like the right side of 250 to me. I think this would collapse to `if (s == end) return false;`
249 stops at first eol or buffer overflow. Then we check if it is an actual eol and exit on overflow or stray cr. I think 251 and 252 are redundant, that was already covered by 250.
This won't allow spaces after PROTO, hope that is correct.

src/http.c
```diff
@@ -199,22 +189,36 @@ struct mg_str *mg_http_get_header(struct mg_http_message *h, const char *name) {
...
+    k.ptr = s;
+    while (s < end && s[0] != ':' && (n = clen(s)) > 0) s += n, k.len += n;
+    if (k.len == 0) return false;               // Empty name
+    if (s >= end || *s++ != ':') return false;  // Invalid, not followed by :
+    while (s < end && s[0] == ' ') s++;         // Skip spaces
+    v.ptr = s;
+    while (s < end && s[0] != '\r' && s[0] != '\n') s++, v.len++;
+    if (s >= end || (s[0] == '\r' && s[1] != '\n')) return false;  // Stray \r
+    if (s < end && s[0] == '\r') s++;
+    if (s >= end || *s++ != '\n') return false;
```
scaprile, line 218: Same comment as 251 below.

scaprile (changes requested):
Minor doubts

src/http.c
scaprile: Well, "below" in code... GH posted my comments in the order I made them...

test/unit_test.c
```diff
@@ -1386,6 +1390,30 @@ static void test_http_parse(void) {
     ASSERT(req.body.len == 0);
   }
 
+  // #2292: fail on stray \n inside the headers
```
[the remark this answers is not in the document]
cpq: fixed

src/http.c
cpq: The only redundant check is 251, because 251 does `s++` and thus it can become `s == end`
We do allow spaces in hm->proto, cause we skip until newline

PR #2301, opened 2023-07-13
fix: Set new socket option required by Linux 3.9+

### LINKED ISSUE #2300: SO_REUSEADDR no longer working on newer Linux kernels
- My goal is: to start and stop an embedded web server in unit tests for another project
- My actions were: run Mongoose according to docs
- My expectation was: second and subsequent server instances could bind to the same port
- The result I saw: errno 98 (EADDRINUSE - Address already in use)

I found an explanation and a solution to this.  Since Linux kernel 3.9, both SO_REUSEADDR and SO_REUSEPORT have to be set as socket options.  I'm going to assume this will not be controversial and send a PR to address it.

# Environment

- mongoose version: 818f99b286df9668803085e9d86fcf7f140aa02e (`master` as of today)
- Compiler/IDE and SDK: gcc 11.3.0 on Ubuntu 22.04
- Target hardware/board: N/A
- Connectivity chip/module: N/A
- Target RTOS/OS (if applicable): N/A

joeyparrish:
See also https://stackoverflow.com/a/25193462/2673491

Fixes #2300

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -4682,6 +4682,11 @@ bool mg_open_listener(struct mg_connection *c, const char *url) {
                                 (char *) &on, sizeof(on))) != 0) {
       // "Using SO_REUSEADDR and SO_EXCLUSIVEADDRUSE"
       MG_ERROR(("setsockopt(SO_EXCLUSIVEADDRUSE): %d %d", on, MG_SOCK_ERR(rc)));
+#elif defined(SO_REUSEPORT)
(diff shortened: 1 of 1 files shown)


joeyparrish:
I was mistaken.  This is not needed.  I think I was leaking the listening socket somehow between instances.  See https://github.com/cesanta/mongoose/issues/2300#issuecomment-1635208389

PR #2343, opened 2023-08-15
Add http serve to fuzzer

cpq:


test/unit_test.c
```diff
@@ -134,21 +134,26 @@ static void test_commalist(void) {
...
   ASSERT(v.len == 0 && mg_vcmp(&k, "b") == 0);
   ASSERT(mg_commalist(&s3, &k, &v) == false);
 
+  v.len = k.len = 42;
   ASSERT(mg_commalist(&s4, &k, &v) == true);
   ASSERT(mg_vcmp(&k, "a") == 0 && mg_vcmp(&v, "123") == 0);
   ASSERT(mg_commalist(&s4, &k, &v) == false);
   ASSERT(mg_commalist(&s4, &k, &v) == false);
 
+  v.len = k.len = 42;
```
scaprile, line 156: Is this magic number explained somewhere ?

scaprile (changes requested):
LGTM, though after fixing the obvious c++ pedantic failure the fuzz test is still failing.
Is the way you are calling the test function OK ? No looping ? mgr_free right after test ?

scaprile (approved):

test/unit_test.c
cpq: No, it is just a random number (the famous answer), just to make sure we're not re-using the result of the previous test.

PR #2344, opened 2023-08-15
improved tcp closure

robertc2000:


scaprile (changes requested):

src/tcpip/tcpip.c
```diff
@@ -654,6 +673,7 @@ static void read_conn(struct mg_connection *c, struct pkt *pkt) {
 }
 
 static void rx_tcp(struct mg_tcpip_if *ifp, struct pkt *pkt) {
+  //mg_hexdump(pkt->raw.ptr, pkt->raw.len);
```
scaprile, line 676: leftover

test/mip_tap_test.c
```diff
@@ -1,6 +1,6 @@
 #define MG_ENABLE_TCPIP 1
 #define MG_ENABLE_SOCKET 0
-#define MG_USING_DHCP 1
+#define MG_USING_DHCP 0
```
scaprile, line 3: Reason ?

scaprile (approved):

test/mip_tap_test.c
```diff
@@ -234,8 +237,9 @@ static void test_http_fetch(void) {
   }
 
   // Clear
+  usleep(1000000);             // Waiting until all connections finally close
```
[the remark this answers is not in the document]
robertc2000: We call the manager before in the fetch function when we initiate the TCP closure and start the timeout for the connection to be deleted. Therefore, before calling the manager at the end we have to wait this maximum amount of time until we are sure all the connections are freed.

cpq (approved):

PR #2347, opened 2023-08-16
Fix #2322 - stricter Content-Length check, allow 1*DIGIT only

### LINKED ISSUE #2322: Improper parsing of `Content-Length` values
# Description
Section 8.6 of RFC 9110 defines the acceptable values in a `Content-Length` field value as follows:
```
Content-Length = 1*DIGIT
```
`DIGIT` is defined as follows in RFC 5234:
```
         DIGIT       =  %x30-39
```
In other words, a `Content-Length` header's value must be composed of one of more ASCII digits.

Mongoose does not enforce this rule. Instead, it interprets each `Content-Length` header value until it encounters a character that is not an ASCII digit, then gives up. This means that, for example, `Content-Length: 1_0` is interpreted as `Content-Length: 1`.

# Suggested Solution
Reject `Content-Length` headers that do not consist of one or more ASCII digits.

cpq:


scaprile (approved):

test/unit_test.c
```diff
@@ -1579,6 +1579,7 @@ static void test_http_range(void) {
 
   fetch(&mgr, buf, url, "%s", "GET /range.txt HTTP/1.0\nRange: bytes=5-10\n\n");
   ASSERT(mg_http_parse(buf, strlen(buf), &hm) > 0);
+  printf("%s", buf);
```
scaprile, line 1582: leftover

test/unit_test.c
```diff
@@ -2184,16 +2185,23 @@ static void test_util(void) {
...
+    mg_snprintf(buf, sizeof(buf), "%lu", SIZE_MAX);
+    MG_INFO(("--> %s", buf));
+    ASSERT(mg_to_size_t(mg_str(buf), &val) && val == SIZE_MAX);
+#if 0
     ASSERT(mg_to_size_t(mg_str("18446744073709551616"), &val) ==
            false);  // range +1
     ASSERT(mg_to_size_t(mg_str("18446744073709551610"), &val) == false);
     // TODO(): ASSERT(mg_to_size_t(mg_str("18446744073709551609"), &val) &&
     //         val == 18446744073709551609U);  // our max or SIZE_MAX
+#endif
```
scaprile, line 2204: These can go now
Yeah, SIZE_MAX and company cause headaches in some platforms...

PR #2351, opened 2023-08-18
mg_base64: check destination buffer size

Jasper-Ben:
Fix usage of base64 en/decode functions to include a destination size

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -82,16 +82,20 @@ int mg_base64_final(char *to, int n) {
   return n;
 }
 
-int mg_base64_encode(const unsigned char *p, int n, char *to) {
+int mg_base64_encode(const unsigned char *p, int n, char *to, int tolen) {
   int i, len = 0;
+  // check if *to is big enough (including null byte)
+  if (tolen < ((n/3) + (n%3?1:0)) * 4 + 1) return 0;
   for (i = 0; i < n; i++) len = mg_base64_update(p[i], to, len);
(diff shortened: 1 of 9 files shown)


Jasper-Ben:
:wave: @cpq,

 this is a security patch that prevents an easy to trigger buffer overflow when doing base64 operations (all that is required it sending a string that is too long). You might also want to request a CVE for this.

Unfortunately, this also means that the function call is somewhat more complex, as the dstlen needs to be calculated beforehand.

@iris-ersc can give you additional details, if required.

Jasper-Ben:
I just pushed an updated version that fixes the unit tests, however Fuzzing currently still breaks

Jasper-Ben:
With https://github.com/cesanta/mongoose/commit/b3abbd4c4d2af1ff61f07ee50588987b7613b319 fuzzing should work now as well.

cpq:
These function assumes that the large enough buffer is passed to them, because it is trivial to calculate the required buffer size. So I don't buy the "CVE" thing.
However, passing a destination buffer size is good, since it makes API more explicit.

Jasper-Ben:
@cpq that works for me :slightly_smiling_face:

cpq:
Thank you !
I pushed #2358 which also changes the API to use size_t, and fixes some more example and adds more unit tests.
Closing this.

Jasper-Ben:
@cpq nice! We were thinking about changing to size_t as well, but didn't want to create a massive diff :laughing:

PR #2380, opened 2023-09-08
IP fragmentation send path: added TCP/UDP packet splitting to fit within MTU

robertc2000:


scaprile (changes requested):
The approach seems to look fine for TCP.
However, for UDP, it will only work if the user calls mg_print*() functions that call mg_io_send(). Otherwise, calls to mg_send() will bypass your strategy and be sent without being fragmented. Please write a test that proves or refutes my point and, if I'm right, then solve it.
https://github.com/cesanta/mongoose/blob/d23664a4574caa6316dcb6d27df5fddb4f449ed7/src/net_builtin.c#L1075-L1088

In case we would need to mg_io_add for UDP, I wouldn't want to delay UDP packets that would fit the MTU, so we should not just buffer, as buffered datagrams would be sent on next poll, not now. For UDP, whatever fits the MTU we should try to send it now. (*)
If I'm right, this fragmentation stuff will have the side effect that we'll now be able to "send UDP datagrams larger than an IP datagram", which will cause them to the splitted into several datagrams and that IMO wouldn't be correct. Please check that, too.

(*): How would this interact with an ARP lookup ? Now our first UDP that requires an ARP lookup is lost, can we make use of this in advance and save the datagram for later ? (If the ISO police OSI enforcement unit comes for you, just tell them I told you to do it)

And now that I think of it again... We are not actually sending fragmented IP, are we ? So if a user requests a 1500-byte UDP datagram and our MTU is 750, we'll actually send two UDP datagrams breaking the very logic foundation of UDP datagrams. I think we should either reject this (force the users to break their messages) and (or at least) log it. Otherwise, we should send IP fragments but we will not receive our own data and that looks bad...
Our current logic is to follow down the path and the Ethernet layer will drop too-long packets.
I think this is turning into a long rambling, if that is the case, let's please follow up at the chat.

src/net_builtin.c
```diff
@@ -1018,11 +1020,13 @@ bool mg_open_listener(struct mg_connection *c, const char *url) {
 }
 
 static void write_conn(struct mg_connection *c) {
-  long len = c->is_tls ? mg_tls_send(c, c->send.buf, c->send.len)
-                       : mg_io_send(c, c->send.buf, c->send.len);
-  if (len > 0) {
-    mg_iobuf_del(&c->send, 0, (size_t) len);
-    mg_call(c, MG_EV_WRITE, &len);
+  while (c->send.len) {
```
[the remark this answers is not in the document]
robertc2000: You are right. Solved!

src/net_builtin.c
```diff
@@ -1018,12 +1020,12 @@ bool mg_open_listener(struct mg_connection *c, const char *url) {
...
-  if (len > 0) {
-    mg_iobuf_del(&c->send, 0, (size_t) len);
-    mg_call(c, MG_EV_WRITE, &len);
-  }
+    long len = c->is_tls ? mg_tls_send(c, c->send.buf, c->send.len)
+                      : mg_io_send(c, c->send.buf, c->send.len);
+    if (len > 0) {
+      mg_iobuf_del(&c->send, 0, (size_t) len);
+      mg_call(c, MG_EV_WRITE, &len);
+    }
```
scaprile, line 1028: Please run clang-format

src/net_builtin.c
```diff
@@ -552,17 +552,19 @@ long mg_io_send(struct mg_connection *c, const void *buf, size_t len) {
...
   struct connstate *s = (struct connstate *) (c + 1);
   uint32_t rem_ip;
+  size_t eth_h_len = 14, ip_max_h_len = 24, tcp_max_h_len = 60, udp_h_len = 8;
+  size_t max_headers_len = eth_h_len + ip_max_h_len +
+                          (c->is_udp ? udp_h_len : tcp_max_h_len);
   memcpy(&rem_ip, c->rem.ip, sizeof(uint32_t));
+  if (len + max_headers_len > ifp->tx.len) {
+    len = ifp->tx.len - max_headers_len;
+  }
+  if (len + max_headers_len - eth_h_len > c->mtu) {
```
cpq, line 562: Also, `mtu` must not be too small - it should be bigger that the size of headers. Shall we check that `mtu` is not less than a minimum possible here, or somewhere else?

src/net.h
```diff
@@ -54,6 +54,8 @@ struct mg_connection {
   void *pfn_data;              // Protocol-specific function parameter
   char data[MG_DATA_SIZE];     // Arbitrary connection data
   void *tls;                   // TLS specific data
+#define MTU_DEFAULT_VALUE 1500
+  uint16_t mtu;                // MTU for this connection
```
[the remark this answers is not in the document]
scaprile: Yeah, I second, missed that, that is ifp-> stuff not c-> ...

cpq (changes requested):
Please move `mtu` setting to the `struct connstate`

cpq (approved):

src/net_builtin.c
```diff
@@ -552,17 +554,24 @@ long mg_io_send(struct mg_connection *c, const void *buf, size_t len) {
...
+  if (len + max_headers_len > ifp->tx.len) {
+    len = ifp->tx.len - max_headers_len;
+  }
+  if (s->mtu < max_headers_len - eth_h_len) {
+    s->mtu = MIP_MTU_DEFAULT;
+  }
+  if (len + max_headers_len - eth_h_len > s->mtu) {
+    len = s->mtu - max_headers_len + eth_h_len;
+    if (c->is_udp)
+      MG_ERROR(("UDP datagram exceeds MTU"));
```
scaprile, line 570: Please tell the datagram has been truncated, something like
"Truncated UDP datagram exceeding MTU"

src/net_builtin.c
```diff
@@ -1081,6 +1090,14 @@ bool mg_send(struct mg_connection *c, const void *buf, size_t len) {
     mg_error(c, "net down");
   } else if (c->is_udp) {
     struct connstate *s = (struct connstate *) (c + 1);
+    size_t max_udp_header_len = 32;
+    if (s->mtu < max_udp_header_len) {
+      s->mtu = MIP_MTU_DEFAULT;
+    }
+    if (len + max_udp_header_len > s->mtu) {
+      len = s->mtu - max_udp_header_len;
+      MG_ERROR(("UDP datagram exceeds MTU"));
```
scaprile, line 1099: same comment as above (truncation)

scaprile (approved):
Please address those nitty comments of mine.

scaprile (commented):
Changes bring more changes...

PR #2386, opened 2023-09-17
Revert to the old TLS API, but keep certs as mg_str for DER

cpq:


src/tls_openssl.c
```diff
@@ -20,11 +20,10 @@ static int mg_tls_err(struct mg_tls *tls, int res) {
   return err;
 }
 
-static STACK_OF(X509_INFO) * load_ca_certs(const char *ca, int ca_len) {
-  BIO *ca_bio = BIO_new_mem_buf(ca, ca_len);
-  if (!ca_bio) return NULL;
```
scaprile, line 25: Is this safe ? No longer checking for NULL before calling (nor freeing later)

src/tls_openssl.c
```diff
@@ -38,100 +37,100 @@ static bool add_ca_certs(SSL_CTX *ctx, STACK_OF(X509_INFO) * certs) {
   return true;
 }
 
-static EVP_PKEY *load_key(const char *key, int key_len) {
-  BIO *key_bio = BIO_new_mem_buf(key, key_len);
-  if (!key_bio) return NULL;
```
scaprile, line 43: Is this safe ? No longer checking for NULL before calling (nor freeing later)

src/tls_openssl.c
```diff
@@ -38,100 +37,100 @@ static bool add_ca_certs(SSL_CTX *ctx, STACK_OF(X509_INFO) * certs) {
...
+static EVP_PKEY *load_key(struct mg_str s) {
+  BIO *bio = BIO_new_mem_buf(s.ptr, (int) (long) s.len);
+  EVP_PKEY *key = PEM_read_bio_PrivateKey(bio, NULL, 0, NULL);
+  BIO_free(bio);
+  return key;
 }
 
-static X509 *load_cert(const char *cert, int cert_len) {
-  BIO *cert_bio = BIO_new_mem_buf(cert, cert_len);
-  if (!cert_bio) return NULL;
```
scaprile, line 51: Is this safe ? No longer checking for NULL before calling (nor freeing later)

src/tls_openssl.c
```diff
@@ -38,100 +37,100 @@ static bool add_ca_certs(SSL_CTX *ctx, STACK_OF(X509_INFO) * certs) {
...
-
-  SSL_set_min_proto_version(tls->ssl, TLS1_2_VERSION);
-
+  SSL_set_session_id_context(tls->ssl, (const uint8_t *) id,
+                             (unsigned) strlen(id));
+  // Disable deprecated protocols
+  SSL_set_options(tls->ssl, SSL_OP_NO_SSLv2);
+  SSL_set_options(tls->ssl, SSL_OP_NO_SSLv3);
+  SSL_set_options(tls->ssl, SSL_OP_NO_TLSv1);
+  SSL_set_options(tls->ssl, SSL_OP_NO_TLSv1_1);
```
scaprile, line 84: I guess these are not the same ?

scaprile (changes requested):
Please see those suggestions as to where to split the PR (changes)
For the rest... I see you removed many redundancies, let's hope they weren't there for a reason. I fail to grasp how would tickets work now, I assume we'll find a way to let connections share context info later.

examples/tcp/main.c
```diff
@@ -54,10 +54,12 @@ static void cfn(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
...
-#if MG_TLS
-    struct mg_str host = mg_url_host(s_conn);
-    mg_tls_init(c, host);
-#endif
+    if (mg_url_is_ssl(s_conn)) {
+      struct mg_tls_opts opts = {.ca = mg_str(s_tls_ca),
+                                 .cert = mg_str(s_tls_cert),
+                                 .key = mg_str(s_tls_key)};
+      mg_tls_init(c, &opts);
+    }
```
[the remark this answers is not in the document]
cpq: There is a way to figure out TCP over TLS: `tls://`. See https://github.com/cesanta/mongoose/blob/7ecbbc517d0d61cb427747c768b2c08dfe0733bb/src/url.c#L10

We can even add `tcps://` there to be very exact.

examples/tcp/main.c
```diff
@@ -89,9 +91,12 @@ static void sfn(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
...
     MG_INFO(("SERVER accepted a connection"));
-#if MG_TLS
-    mg_tls_init(c, mg_str(""));
-#endif
+    if (mg_url_is_ssl(s_lsn)) {
+      struct mg_tls_opts opts = {.ca = mg_str(s_tls_ca),
+                                 .cert = mg_str(s_tls_cert),
+                                 .key = mg_str(s_tls_key)};
+      mg_tls_init(c, &opts);
+    }
```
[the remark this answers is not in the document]
cpq: `tcp://` is a valid URL for Mongoose :) And I think it makes sense to add `tcps://` to url.c

src/tls_mbed.c
```diff
@@ -54,40 +88,27 @@ void mg_tls_handshake(struct mg_connection *c) {
...
   struct mg_tls *tls = (struct mg_tls *) calloc(1, sizeof(*tls));
   int rc = 0;
-
   c->tls = tls;
   if (c->tls == NULL) {
     mg_error(c, "TLS OOM");
     goto fail;
   }
-
+  if (c->is_listening) goto fail;
```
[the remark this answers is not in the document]
cpq: Well we should not init TLS on a listening connection, only on accepted ones.

src/tls_openssl.c
cpq: Fixed

src/tls_openssl.c
cpq: ditto

src/tls_openssl.c
cpq: ditto

src/dns.c
```diff
@@ -16,17 +16,17 @@ struct dns_data {
 static void mg_sendnsreq(struct mg_connection *, struct mg_str *, int,
                          struct mg_dns *, bool);
 
-static void mg_dns_free(struct mg_connection *c, struct dns_data *d) {
-  LIST_DELETE(struct dns_data,
-              (struct dns_data **) &c->mgr->active_dns_requests, d);
+static void mg_dns_free(struct dns_data **head, struct dns_data *d) {
+  LIST_DELETE(struct dns_data, head, d);
```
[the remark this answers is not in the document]
cpq: Yes! Will send separately.

src/tls_openssl.c
cpq: Not the same

src/tls_mbed.c
```diff
@@ -1,20 +1,54 @@
-#include "fs.h"
-#include "printf.h"
+#include "log.h"
 #include "tls.h"
 
 #if MG_TLS == MG_TLS_MBED
 
 #if defined(MBEDTLS_VERSION_NUMBER) && MBEDTLS_VERSION_NUMBER >= 0x03000000
-#define MGRNG , rng_get, NULL
+#define MBEDTLS_RNG_GET , mg_mbed_rng, NULL
```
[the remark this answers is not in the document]
cpq: Added MG_!

test/unit_test.c
```diff
@@ -1734,12 +1748,43 @@ static void test_timer(void) {
...
+
   // Test proper timer deallocation, see #1539
   {
     struct mg_mgr mgr;
     mg_mgr_init(&mgr);
     mg_timer_add(&mgr, 1, MG_TIMER_REPEAT, f1, NULL);
+    ASSERT(mgr.timers != NULL);
     mg_mgr_free(&mgr);
+    ASSERT(mgr.timers == NULL);
     ASSERT(mgr.conns == NULL);
```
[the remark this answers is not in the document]
cpq: Well this is a genuine test. If it fails, then it should be fixed.

test/unit_test.c
```diff
@@ -1242,45 +1243,57 @@ static void f3(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
...
+    opts.ca = mg_str_n(data, size);
+    //opts.name = mg_url_host(url);
+    mg_tls_init(c, &opts);
+  }
+  for (i = 0; i < 1500 && ok <= 0; i++) mg_mgr_poll(&mgr, 1);
   ASSERT(ok == 200);
   c->is_closing = 1;
   mg_mgr_poll(&mgr, 1);
+
 #if 0
```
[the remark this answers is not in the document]
cpq: It should

src/tls_openssl.c
```diff
@@ -38,100 +38,101 @@ static bool add_ca_certs(SSL_CTX *ctx, STACK_OF(X509_INFO) * certs) {
...
```
scaprile, line 86: Sorry, inconsistent marking, I mean these

scaprile (approved):
I'll check if that disable test passes.
Please see that lone return() at the very top of a test

PR #2390, opened 2023-09-21
Restore tickets support

cpq:


src/tls_mbed.c
```diff
@@ -191,4 +186,33 @@ long mg_tls_send(struct mg_connection *c, const void *buf, size_t len) {
...
+    mbedtls_ssl_ticket_init(&ctx->tickets);
+    if ((rc = mbedtls_ssl_ticket_setup(&ctx->tickets, mg_mbed_rng, NULL,
+                                       MBEDTLS_CIPHER_AES_128_GCM, 86400)) !=
+        0) {
+      MG_ERROR((" mbedtls_ssl_ticket_setup %#x", -rc));
+    }
+#endif
+    mgr->tls_ctx = ctx;
+  }
+}
```
scaprile, line 206: Does it make sense to alloc memory and have a pointer when there is no support for that feature in MbedTLS ?

What about 
```
void mg_tls_ctx_init(struct mg_mgr *mgr) {
#ifdef MBEDTLS_SSL_SESSION_TICKETS
#else
  MG_DEBUG(("MBEDTLS_SSL_SESSION_TICKETS disabled"))
#endif
}
```

scaprile (approved):

src/tls_mbed.c
cpq: Meh. This is so minor. 1 allocation only. If it was per connection, then it would be worth the effort.

PR #2404, opened 2023-10-03
Add nucleo-g031 baremetal example

cpq:


scaprile (changes requested):
We can always make it prettier and commonize later.
2 things:
- if we are using the W5500, all other examples that do that have "w5500" in their name.
- why is mongoose.c file here  so different ? I wouldn't merge this file as is, did you change the Makefile ?

PR #2410, opened 2023-10-05
added OTA single bank support for H7

robertc2000:


src/device_stm32h5.c
```diff
@@ -12,6 +12,8 @@
 #define FLASH_NSCCR (FLASH_BASE + 0x30)
 #define FLASH_OPTSR_CUR (FLASH_BASE + 0x50)
 #define FLASH_OPTSR_PRG (FLASH_BASE + 0x54)
+#define SCB (0xE000E000UL +  0x0D00UL)
+#define SCB_VTOR  (SCB + 0x008)
```
scaprile, line 16: This is ARM Cortex-M, it is not H5 specific, we could rely on CMSIS or to keep the baremetal approach just undef SCB and VTOR and define a pointer and a simplified struct to use SCB->VTOR.
This could go in device.h, it won't harm non-ARM devices if we undef and define our own

src/device_stm32h7.c
```diff
@@ -13,12 +13,18 @@
 #define FLASH_CCR 0x14
 #define FLASH_OPTSR_CUR 0x1c
 #define FLASH_OPTSR_PRG 0x20
+#define SCB (0xE000E000UL +  0x0D00UL)
+#define SCB_VTOR  (SCB + 0x008)
```
scaprile, line 17: Same comment as for H5

examples/stm32/nucleo-h723zg-make-baremetal-builtin/mongoose_custom.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#define MG_ARCH MG_ARCH_NEWLIB
+#define MG_OTA MG_OTA_FLASH
+#define MG_DEVICE MG_DEVICE_STM32H7
+#define MG_DEVICE_SINGLE_BANK
```
cpq, line 6: All our defines are numeric - please follow the tradition. Maybe it is better to use DUAL_BANK? Cause single bank is kind of a default ? I'd add this to the config.h:

```c
#ifndef MG_DEVICE_DUAL_BANK
#define MG_DEVICE_DUAL_BANK 0
#endif
```

Then for H743 and H563 we will add this:

```c
#define MG_DEVICE_DUAL_BANK 1
```

src/device_stm32h5.c
```diff
@@ -12,6 +12,8 @@
 #define FLASH_NSCCR (FLASH_BASE + 0x30)
 #define FLASH_OPTSR_CUR (FLASH_BASE + 0x50)
 #define FLASH_OPTSR_PRG (FLASH_BASE + 0x54)
+#define SCB (0xE000E000UL +  0x0D00UL)
```
cpq, line 15: That's a constant, why not this:
```c
#define SCB 0xe000ed00   // Base address of a System Control Block
```

src/ota.h
```diff
@@ -19,10 +19,10 @@ bool mg_ota_write(const void *buf, size_t len);  // Write chunk, aligned to 1k
...
 
 enum {
-  MG_OTA_UNAVAILABLE = 0,  // No OTA information is present
-  MG_OTA_FIRST_BOOT = 1,   // Device booting the first time after the OTA
-  MG_OTA_UNCOMMITTED = 2,  // Ditto, but marking us for the rollback
-  MG_OTA_COMMITTED = 3,    // The firmware is good
+  MG_OTA_UNAVAILABLE = 0,    // No OTA information is present
+  MG_OTA_FIRST_BOOT = 1,     // Device booting the first time after the OTA
+  MG_OTA_COMMITTED = 2,      // The firmware is good
+  MG_OTA_ROLLBACK = 4        // Rolling back to this firmware
```
scaprile, line 25: What happened to 3 ?
If those are bits I would use (1<<n) or MG_BIT()

robertc2000: Yes, these are bits. I initially used MG_BIT but it didn't work since the compiler could not find the reference to MG_BIT in a certain test, I will use (1 << n) instead

src/device_stm32h7.c
```diff
@@ -13,12 +13,18 @@
...
 void *mg_flash_start(void) {
   return (void *) 0x08000000;
 }
 size_t mg_flash_size(void) {
-  return 2 * 1024 * 1024;  // 2Mb
+#ifdef MG_DEVICE_SINGLE_BANK
+  return 1024 * 1024;  // 1Mb
+#else
+  return 2 * 1024 * 1024; // 2Mb;
+#endif
```
cpq, line 27: Instead, read flash size from a device.
For example, [RM0433](https://www.st.com/resource/en/reference_manual/rm0433-stm32h742-stm32h743753-and-stm32h750-value-line-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) for H743 in section 61.1 describes the Unique device ID register. It is a unique 96-bit STM32 identifier.
What follows in section 61.2 is the flash size register: address 0x1FF1E880 which contains the size in kB. Let's read it instead of hardcoding the values.

examples/stm32/nucleo-h723zg-make-baremetal-builtin/link.ld
```diff
@@ -0,0 +1,34 @@
...
+SECTIONS {
+  .vectors  : { KEEP(*(.isr_vector)) }  > flash
+  .text     : { *(.text* .text.*) }     > flash
+  .rodata   : { *(.rodata*) }           > flash
+
+  .data : {
+    _sdata = .;   /* for init_ram() */
+    *(.first_data)
+    *(.data SORT(.data.*))
+    *(.bootloader)        /* .bootloader function sections */
```
cpq, line 20: The common practice is to call this section "iram" (Instruction RAM) - let's use `.iram` instead

src/device.h
```diff
@@ -12,6 +12,10 @@
 
 #ifndef MG_DEVICE
 #define MG_DEVICE MG_DEVICE_NONE
+#elif MG_DEVICE == MG_DEVICE_STM32H7
+#define MG_BOOTLOADER_SECTION __attribute__((section(".bootloader")))
```
cpq, line 16:
```c
#define MG_IRAM ...
```

src/device_stm32h5.c
scaprile: CMSIS stuff, SCB->VTOR, present in all ARM cores except M0 and M0+ where it is optional (IIRC) and there is an SCB_VTOR_PRESENT constant or something like that to check for it

src/ota.h
scaprile: Perhaps util.h has to be included, there is one of the tests (unamalgamated ?) that honors its name

src/ota_flash.c
```diff
@@ -19,14 +19,29 @@ struct mg_otadata {
   uint32_t crc32, size, timestamp, status;
 };
 
+static uint32_t mg_ota_get_crt_key(void) {
```
cpq, line 22: mg_ota_key_current() and mg_ota_get_key_previous() please

src/device_stm32h5.c
cpq: @scaprile this code, like our ethernet drivers, does not pull CMSIS

src/device.h
scaprile: That won't work on non gcc-based compilers... see my comment regarding reinventing CMSIS

examples/stm32/nucleo-h723zg-make-baremetal-builtin/link.ld
scaprile: Yes, "bootloader" I though of a specific flash partition

PR #2420, opened 2023-10-10
OTA single bank firmware swapping

robertc2000:


examples/stm32/nucleo-h723zg-make-baremetal-builtin/README.md
```diff
@@ -0,0 +1,3 @@
+# Baremetal web device dashboard on NUCLEO-H743ZI
```
scaprile, line 1: 723ZG

examples/stm32/nucleo-h723zg-make-baremetal-builtin/main.c
```diff
@@ -0,0 +1,80 @@
...
+  mg_log_set(MG_LL_DEBUG);        // Set log level
+  mg_ota_bootloader();
+
+  MG_INFO(("Chip revision: %c, max cpu clock: %u MHz", chiprev(),
+           (chiprev() == 'V') ? 480 : 400));
+  MG_INFO(("Starting, CPU freq %g MHz", (double) SystemCoreClock / 1000000));
+
+  struct mg_mgr mgr;        // Initialise
+  mg_mgr_init(&mgr);        // Mongoose event manager
+  mg_log_set(MG_LL_DEBUG);  // Set log level
```
scaprile, line 48: mg_log_set is called twice with the same argument
Maybe first one is not needed ?

src/config.h
```diff
@@ -4,10 +4,6 @@
 #define MG_ENABLE_LOG 1
 #endif
 
-#ifndef MG_ENABLE_CUSTOM_LOG
-#define MG_ENABLE_CUSTOM_LOG 0  // Let user define their own MG_LOG
-#endif
-
```
scaprile, line 10: `git rebase master`

src/device_stm32h7.c
```diff
@@ -13,69 +13,88 @@
 #define FLASH_CCR 0x14
 #define FLASH_OPTSR_CUR 0x1c
 #define FLASH_OPTSR_PRG 0x20
+#define FLASH_SIZE_REG 0x1FF1E880
```
scaprile, line 16: I still prefer FLASH->CCR and so, though I understand it complicates things...

src/log.c
```diff
@@ -26,10 +26,8 @@ void mg_log_set(int log_level) {
   s_level = log_level;
 }
 
-#if MG_ENABLE_CUSTOM_LOG
-// Let user define their own mg_log_prefix() and mg_log()
-#else
-bool mg_log_prefix(int level, const char *file, int line, const char *fname) {
```
scaprile, line 32: rebase

src/ota.h
```diff
@@ -13,6 +14,11 @@
 #define MG_OTA MG_OTA_NONE
 #endif
 
+#if defined(__GNUC__) && !defined(__APPLE__)
+#undef MG_IRAM
+#define MG_IRAM __attribute__((section(".iram")))
+#endif
+
```
scaprile, line 21: I see all ota functions (most ?) are prefixed with this.
Do all of them need to run in RAM ? Maybe not the case for dual bank devices, or are those not called/used with dual bank devices ?

robertc2000: Only if the device is single bank do they need to run in RAM. Perhaps we could also add a condition on line 17 to check if the device is single bank?

PR #2446, opened 2023-11-02
Single bank OTA

cpq:


examples/stm32/nucleo-h723zg-make-baremetal-builtin/hal.h
```diff
@@ -8,6 +8,16 @@
...
+#define BTN PIN('C', 13)   // User push button
+#define LED1 PIN('B', 0)   // On-board LED pin (green)
+#define LED2 PIN('E', 1)   // On-board LED pin (yellow)
+#define LED3 PIN('B', 14)  // On-board LED pin (red)
+#define LED LED2           // Use yellow LED for blinking
+
+#ifndef UART_DEBUG
+#define UART_DEBUG USART3
+#endif
+
```
scaprile, line 20: I like defines after includes, as it was before... just a matter of taste and having all necessary stuff at top
Plus, all hal.h have UART_DEBUG near the UART code

src/log.h
```diff
@@ -5,16 +5,23 @@
...
 #if MG_ENABLE_LOG
-#define MG_LOG(level, args)                                                \
-  do {                                                                     \
-    if (mg_log_prefix((level), __FILE__, __LINE__, __func__)) mg_log args; \
+#define MG_LOG(level, args)                                 \
+  do {                                                      \
+    if ((level) <= mg_log_level) {                          \
+      mg_log_prefix((level), __FILE__, __LINE__, __func__); \
+      mg_log args;                                          \
+    }                                                       \
```
scaprile, line 24: Cool.
I guess you couldn't resist temptation, a separate PR could have been nice

PR #2468, opened 2023-11-08
Add teensy41 example

cpq:


src/drivers/rt1020.c
```diff
@@ -52,16 +52,16 @@ uint8_t s_txbuf[ETH_DESC_CNT][ETH_PKT_SIZE] __attribute__((aligned((64U))));
 
 static struct mg_tcpip_if *s_ifp;  // MIP interface
 
-enum { PHY_ADDR = 2, PHY_BCR = 0, PHY_BSR = 1, PHY_PC1R = 30, PHY_PC2R = 31 };
+enum { PHY_BCR = 0, PHY_BSR = 1, PHY_ID2 = 2, PHY_ID3 = 3 };
 
-static uint32_t eth_read_phy(uint8_t addr, uint8_t reg) {
+static uint16_t rt10xx_phy_read(uint8_t addr, uint8_t reg) {
```
scaprile, line 57: Most drivers have an "eth_write_phy"
Yes, the new name is taxonomy friendlier, I like it

src/drivers/rt1020.c
```diff
@@ -52,16 +52,16 @@ uint8_t s_txbuf[ETH_DESC_CNT][ETH_PKT_SIZE] __attribute__((aligned((64U))));
 
 static struct mg_tcpip_if *s_ifp;  // MIP interface
 
-enum { PHY_ADDR = 2, PHY_BCR = 0, PHY_BSR = 1, PHY_PC1R = 30, PHY_PC2R = 31 };
+enum { PHY_BCR = 0, PHY_BSR = 1, PHY_ID2 = 2, PHY_ID3 = 3 };
```
scaprile, line 55: proper register names are ID1 and ID2

src/drivers/rt1020.c
```diff
@@ -145,14 +167,18 @@ static size_t mg_tcpip_driver_rt1020_tx(const void *buf, size_t len,
...
-    uint32_t rcr = ENET->RCR &= ~BIT(9);       // 100M
-    if ((pc1r & 3) == 1) rcr |= BIT(9);        // 10M
-    if ((pc1r & BIT(2)) == 0) tcr &= ~BIT(2);  // Half-duplex
+    uint32_t tcr = ENET->TCR |= BIT(2);             // Full-duplex
+    uint32_t rcr = ENET->RCR &= ~BIT(9);            // 100M
+    if (rt10xx_phy_read(d->phy_addr, PHY_ID2)) {    // KSZ8081RNB ?
+      uint32_t pc1r = rt10xx_phy_read(d->phy_addr, 30);  // Read PC2R
+      if ((pc1r & 3) == 1) rcr |= BIT(9);                // 10M
+      if ((pc1r & BIT(2)) == 0) tcr &= ~BIT(2);          // Half-duplex
+    }
```
scaprile, line 181: Here we must check PHY ID again. In fact, this code, as is, will not run at 10baseT nor half-duplex on the Teensy.
For the DP83825, the register to read in order to get this info is PHYSTS (Offset = 0x10)

Actions here, are "up" and will be done by all drivers (with non built-in PHY), though each one will have its own way to access the PHY.

scaprile (approved):
So, in the future, we can move PHY detection and "init" + "up" functions to one or more PHY drivers, and these MAC drivers will call a function here, passing a pointer to the actual function that will interact with the MAC controller to get the PHY data.
This will avoid a lot of similar code with same function in most drivers.

examples/stm32/nucleo-h743zi-make-baremetal-builtin/main.c
```diff
@@ -27,7 +27,7 @@ static void timer_fn(void *arg) {
   gpio_toggle(LED);                                      // Blink LED
   struct mg_tcpip_if *ifp = arg;                         // And show
   const char *names[] = {"down", "up", "req", "ready"};  // network stats
-  MG_INFO(("Ethernet: %s, IP: %M, rx:%u, tx:%u, dr:%u, er:%u",
+  MG_INFO(("Ethernet 2: %s, IP: %M, rx:%u, tx:%u, dr:%u, er:%u",
```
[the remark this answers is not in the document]
cpq: Ah, that is testing leftover, removed

src/drivers/rt1020.c
cpq: I was thinking that in (very unlikely situation) we have 2 drivers enabled - they  would clash then.

src/drivers/rt1020.c
cpq: Oh. Fixed.

src/drivers/rt1020.c
```diff
@@ -98,10 +98,32 @@ static bool mg_tcpip_driver_rt1020_init(struct mg_tcpip_if *ifp) {
...
+
+  if (phy_id2 == 0x22) {  // KSZ8081RNB, like EVK-RTxxxx boards
+    rt10xx_phy_write(d->phy_addr, 31, BIT(15) | BIT(8) | BIT(7));  // PC2R
+    // PHY_PC1R = 30, PHY_PC2R = 31
+  } else if (phy_id2 == 0x2000) {              // DP83825I, like Teensy4.1
+    rt10xx_phy_write(d->phy_addr, 23, 0x81);   // 50Mhz clock input
+    rt10xx_phy_write(d->phy_addr, 24, 0x280);  // LED status, active high
+  } else {                                     // Default to LAN8720
+    MG_INFO(("Defauling to LAN8720 PHY..."));
+  }
```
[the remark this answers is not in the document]
cpq: Specifically here, the "up" function is called infrequently so it's not a big deal to query the ID.
Also, PHY is kinda Ethernet specific..
Let's for time being keep it this way, to keep things as loosely coupled as possible.
If we have to optimise in the future - let's think then..

src/drivers/rt1020.c
```diff
@@ -98,10 +98,32 @@ static bool mg_tcpip_driver_rt1020_init(struct mg_tcpip_if *ifp) {
...
+
+  // uint8_t pc2r = d->phy_pc2r;
+  // if (pc2r == 0) pc2r = PHY_PC2R;
+  // rt10xx_phy_write(d->phy_addr, pc2r, BIT(15) | BIT(8) | BIT(7));
+  // if (d->phy_ledcr) rt10xx_phy_write(d->phy_addr, d->phy_ledcr, 0x280);
+
```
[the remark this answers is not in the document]
cpq: Gone!

src/drivers/rt1020.c
cpq: Ack. I'll leave this for you!

PR #2471, opened 2023-11-12
close_conn elegantly

mudori:
shutdown socket before close


scaprile:
Can you please explain what is your problem, why you need this, and your use case ? What platform are you using ?

scaprile:
From [the docs](https://mongoose.ws/documentation/#2-minute-integration-guide)
> NOTE: Since Mongoose's core is not protected against concurrent accesses, make sure that all mg_* API functions are called from the same thread or RTOS task.

If you absolutely need multi-threading, take the time to go through the user guide (linked above) and [tutorials](https://mongoose.ws/tutorials/), there is one that shows you how to do [multi-threading](https://mongoose.ws/tutorials/multi-threaded/).

PR #2485, opened 2023-11-22
Add example for ch32v307

cpq:


src/drivers/stm32.c
```diff
@@ -192,13 +195,16 @@ static bool mg_tcpip_driver_stm32_up(struct mg_tcpip_if *ifp) {
   return up;
 }
 
+#ifdef __riscv
+__attribute__((interrupt()))  // For RISCV CH32V307, which share the same MAC
+#endif
```
scaprile, line 200: This compiler dependent, not only chip dependent

src/drivers/stm32.c
```diff
@@ -209,8 +215,8 @@ void ETH_IRQHandler(void) {
       if (++s_rxno >= ETH_DESC_CNT) s_rxno = 0;
     }
   }
-  ETH->DMASR = MG_BIT(7);  // Clear possible RBUS while processing
-  ETH->DMARPDR = 0;     // and resume RX
+  ETH->DMASR = ~0;   // Clear flags
```
scaprile, line 218: Please hide this q&d hack under some #ifdef

PR #2486, opened 2023-11-22
Post-IRQ DMASR flags cleanup

cpq:


mongoose.c
```diff
@@ -9308,12 +9308,12 @@ void ETH_IRQHandler(void) {
...
     }
   }
-#ifdef __riscv
-  ETH->DMASR = ~0;   // TODO: do more fine-grained flag cleanup
-#else
-  ETH->DMASR = MG_BIT(7);  // Clear possible RBUS while processing
-#endif
-  ETH->DMARPDR = 0;  // and resume RX
+  // Cleanup flags
+  ETH->DMASR = (7UL < 20)    // TPS - Transmit process status
```
scaprile, line 9312: "These bits indicate the Transmit DMA FSM state. This field does not generate an interrupt."
The case where this may stop (tx stops when no more data is sent) is handled at the TX function call. This should not be necessary and definitely not here; furthermore, these bits are read-only, so this has no effect.

mongoose.c
```diff
@@ -9308,12 +9308,12 @@ void ETH_IRQHandler(void) {
...
+               | (7UL < 17)  // RPS - Receive process status
```
scaprile, line 9313: "These bits indicate the Receive DMA FSM state. This field does not generate an interrupt."
The case where this may stop (a burst arrives and no more descriptors are left) is handled at the end of the handler with `ETH->DMARPDR = 0`. This should not be necessary; furthermore, these bits are read-only, so this has no effect.

mongoose.c
```diff
@@ -9308,12 +9308,12 @@ void ETH_IRQHandler(void) {
...
+               | MG_BIT(16)  // NIS, normal interrupt summary
```
scaprile, line 9314: > Bit 16 NIS: Normal interrupt summary
The normal interrupt summary bit value is the logical OR of the following when thecorresponding interrupt bits are enabled in the ETH_DMAIER register:
– ETH_DMASR [0]: Transmit interrupt
– ETH_DMASR [2]: Transmit buffer unavailable
– ETH_DMASR [6]: Receive interrupt
– ETH_DMASR [14]: Early receive interrupt
Only unmasked bits affect the normal interrupt summary bit.
This is a sticky bit and it must be cleared (by writing a 1 to this bit) each time a corresponding bit that causes NIS to be set is cleared.

This bit has been cleared above when this comes as part of the RX function.
It will be set because of TX, and that will be handled in TX, if it is set because of RX, then it will be handled on the next interrupt. The only condition that is not handled is "Early receive interrupt", which I think is not enabled, but this chip may be different.
If setting this bit solves your problem, please keep it wrapped, there is no reason to do this with the STM

mongoose.c
```diff
@@ -9308,12 +9308,12 @@ void ETH_IRQHandler(void) {
...
+               | MG_BIT(7);  // RBUS - Receive buffer unavailable status
```
scaprile, line 9315: Please keep prior comment, it tells why we reset this flag

scaprile (changes requested):

scaprile (approved):

PR #2493, opened 2023-11-25
Fix #2456 - wake up mongoose thread

### LINKED ISSUE #2456: Realtime Data with Web Sockets
Now that pipe functionality is depreciated when sending data using web sockets, I was wondering if there is any alternative? I looked through the documentation, and all the solutions there appear to be polling only, with no other way to wake up the main event thread to actually send or receive the data. I am working on an application that sends realtime audio and video data over a web socket connection, and previously the pipe feature worked great for this type of application.

cpq:


scaprile (approved):
Yes, elegant enough, as always.
Signals stuff seems to bring more inconvenience than solutions so this is the proper way.
I'll add it to the tutorial with some comments on this being optional and just for the sake of saving some sleepy time.

PR #2495, opened 2023-11-27
Add modbus dashboard

cpq:


scaprile (commented):
I hope my copy is not very old and you based on a newer one

Slave URL:
> All MODBUS/TCP ADU are sent via TCP to registered port 502.
> The listening TCP port 502 is reserved for MODBUS communications. It is
mandatory to listen by default on that port.

Slave ID:
> Unit Identifier
This field is used for routing purpose when addressing a device on a MODBUS+
or MODBUS serial line sub-network. In that case, the “Unit Identifier” carries the
MODBUS slave address of the remote device:
The MODBUS slave device addresses on serial line are assigned from 1 to
247 (decimal). Address 0 is used as broadcast address.
On TCP/IP, the MODBUS server is addressed using its IP address; therefore, the
MODBUS Unit Identifier is useless. The value 0xFF has to be used.

Starting register:
This is weird because old PLC dudes number then from 1 to 65536, the reail thing is that content in the wire is 0-65535, and masters vary in the way they present them, be it 0-65535 or 1-65536

Register count:
This depends on the actual Modbus function in use.
If you use Read or Write Register, there is only one register accessed.
If you use Read Multiple or Write Multiple, then this field is used, and it can be 1.
In both cases, these are different Modbus Function numbers, the usual master lets the user choose the function (Read Single Input, Read Multiple, Read Single Holding, Read Multiple Holding, Write Single Holding, Write Multiple Holding, Slave ID)
A device may not support Read Multiple or Write Multiple, that is the reason.

examples/modbus-dashboard/net.c
```diff
@@ -0,0 +1,256 @@
...
+                                                  unsigned long cid) {
+  struct mg_connection *c = NULL;
+  char *url = mg_json_get_str(json, "$.url");
+  long timeout = mg_json_get_long(json, "$.timeout", 750);
+  uint8_t id = (uint8_t) mg_json_get_long(json, "$.id", 1);
+  uint16_t reg = mg_htons((uint16_t) (mg_json_get_long(json, "$.reg", 1) - 1));
+  // uint16_t val = (uint16_t) mg_json_get_long(json, "$.val", 0);
+  uint8_t func = (uint8_t) mg_json_get_long(json, "$.func", 0);
+  uint16_t nregs = mg_htons((uint16_t) mg_json_get_long(json, "$.nregs", 1));
+  MG_INFO(("%lu REQUEST: %.*s", cid, json.len, json.ptr));
```
scaprile, line 122: I think it is clearer to have 16-bit quantities in host order and only convert them to network order when sending.

PR #2508, opened 2023-11-30
Use V307 flash

cpq:


examples/stm32/nucleo-h563zi-make-baremetal-builtin/main.c
```diff
@@ -77,7 +77,7 @@ int main(void) {
   }
 
   MG_INFO(("Initialising application..."));
-  web_init(&mgr);
+  //web_init(&mgr);
```
robertc2000, line 80: Is there a reason for this change?

cpq: Oh, good catch. I did that for testing and forgot to re-enable. Let me cleanup..

PR #2516, opened 2023-12-07
mqtt dashboard compatibility with hal.h functions

robertc2000:

PR #2519, opened 2023-12-08
Add profiling capability

cpq:
This PR allows to tag certain events with their timestamps in the `c->prof` IO buffer.
When a connection is closed, we dump the timestamps of the events.
This way we can profile bottlenecks, especially in the TLS code

To test,

```sh
make -C examples/http-server/ CFLAGS_EXTRA=-DMG_ENABLE_PROFILE=1
```

In other terminal,
```sh
curl localhost:8000
```

output in the 1st terminal:

```
1ea1f1e7 2 net.c:150:mg_close_conn      3 profile:                                                            
1ea1f1e8 2 net.c:150:mg_close_conn      1ea1f1e6 init                                                         
1ea1f1e8 2 net.c:150:mg_close_conn          0 EV_OPEN
1ea1f1e8 2 net.c:150:mg_close_conn          0 EV_ACCEPT 
1ea1f1e8 2 net.c:150:mg_close_conn          0 EV_READ
1ea1f1e8 2 net.c:150:mg_close_conn          0 EV_HTTP_MSG
1ea1f1e8 2 net.c:150:mg_close_conn          0 EV_WRITE
1ea1f1e8 2 net.c:150:mg_close_conn          1 EV_CLOSE
```


examples/mqtt-dashboard/device/main.c
```diff
@@ -12,13 +12,13 @@ static void signal_handler(int signo) {
 // Mocked device pins
 static bool s_pins[100];
 
-void gpio_write(int pin, bool status) {
+void hal_gpio_write(int pin, bool status) {
```
scaprile, line 15: unrelated ground here.
We are removing these, they will be in hal.h, and there are some fixes too. #2516
Hal functions already have the name we are using.
Do you propose to have these in main and they call the hal ?
I'd rather have also a hal here, this code might run on a Raspberry Pi and light an actual LED or flip a relay.
We'll also keep main.c clean

examples/stm32/nucleo-h723zg-make-baremetal-builtin/main.c
```diff
@@ -23,6 +23,16 @@ void mg_random(void *buf, size_t len) {  // Use on-board RNG
...
+#ifdef MQTT_DASHBOARD
+void hal_gpio_write(int pin, bool status) {  // For MQTT dashboard HAL
+  gpio_write((uint16_t) pin, status);
+}
+
+bool hal_gpio_read(int pin) {  // For MQTT dashboard HAL
+  return gpio_read((uint16_t) pin);
+}
+#endif
+
```
scaprile, line 35: This is not necessary if we do as in #2516 
I would also prefer to keep this away from the profiling stuff

src/arch_zephyr.h
```diff
@@ -22,8 +22,15 @@
...
 #define strerror(x) zsock_gai_strerror(x)
+
+#ifndef FD_CLOEXEC
 #define FD_CLOEXEC 0
+#endif
+
+#ifndef F_SETFD
 #define F_SETFD 0
+#endif
+
```
scaprile, line 33: I see you also fixed the Zephyr tests...

src/event.c
```diff
@@ -4,6 +4,17 @@
...
 void mg_call(struct mg_connection *c, int ev, void *ev_data) {
+#if MG_ENABLE_PROFILE
+  const char *names[] = {
+      "EV_ERROR",    "EV_OPEN",      "EV_POLL",      "EV_RESOLVE",
+      "EV_CONNECT",  "EV_ACCEPT",    "EV_TLS_HS",    "EV_READ",
+      "EV_WRITE",    "EV_CLOSE",     "EV_HTTP_MSG",  "EV_HTTP_CHUNK",
+      "EV_WS_OPEN",  "EV_WS_MSG",    "EV_WS_CTL",    "EV_MQTT_CMD",
+      "EV_MQTT_MSG", "EV_MQTT_OPEN", "EV_SNTP_TIME", "EV_USER"};
+  if (ev != MG_EV_POLL && ev < (int) (sizeof(names) / sizeof(names[0]))) {
+    MG_PROF_ADD(c, names[ev]);
```
scaprile, line 15: Invoking this macro but not including profile.h

src/profile.h
```diff
@@ -0,0 +1,44 @@
...
+    struct mg_profitem *first = (struct mg_profitem *) io->buf;        \
+    struct mg_profitem item = {name_, (uint32_t) mg_millis()};         \
+    if (io->len >= sizeof(*first)) item.timestamp -= first->timestamp; \
+    mg_iobuf_add(io, io->len, &item, sizeof(item));                    \
+  } while (0)
+
+#define MG_PROF_INIT(c)                \
+  do {                                 \
+    mg_iobuf_init(&(c)->prof, 0, 256); \
+    MG_PROF_ADD((c), "init");          \
```
scaprile, line 22: Perhaps we could have a special treatment for the "init" case and save that check that is always true for the others ?
Just trying to spare a picosecond... but maybe you have a bright idea

src/profile.h
```diff
@@ -0,0 +1,44 @@
...
+
+#define MG_PROF_DUMP(c)                                               \
+  do {                                                                \
+    struct mg_iobuf *io = &c->prof;                                   \
+    struct mg_profitem *p = (struct mg_profitem *) io->buf;           \
+    struct mg_profitem *e = &p[io->len / sizeof(*p)];                 \
+    MG_INFO(("%lu profile:", c->id));                                 \
+    while (p < e) {                                                   \
+      MG_INFO(("%5lx %s", (unsigned long) p->timestamp, p->name)); \
+      p++;                                                            \
```
scaprile, line 35: These will most likely have the same timestamp, we could be less intrusive by combining everything in one log. Or would alloc+free be worse ? I think printing 5 extra timestamps for 5 events is quite more time consuming.

scaprile (commented):
Cool.
Please see unrelated stuff

src/profile.h
cpq: Done!

src/profile.h
cpq: Printing time is not a worry at all.
Added comments!

examples/stm32/nucleo-h723zg-make-baremetal-builtin/main.c
scaprile: OK, there are additional benefits doing it your way, it is more flexible. There are some issues with "read" that this way solves.
I'll migrate #2516 stuff to this way.

src/profile.h
scaprile: Not here, yes, we are not queing. Maybe for huge file serving it may be a concern

PR #2523, opened 2023-12-09
TLS layer: c->rtls to optimise recvd TLS data

cpq:
The reason for this PR is to minimise the amount of data copies for the TLS case. TLDR: raw data is now received in the `c->rtls` IO buffer, and can be accessed directly. 

So, receive path:
`stack receives new raw data -> c->rtls -> (TLS decrypt) -> c->recv`

Note that the send path does not have the same. Apparently, TLS implementation can just encrypt data (`mg_tls_send()`) data on-the-fly.

Right now, the builtin receive path looks like this:

1. Stack. Packet data gets copied to the s->raw IO buffer
2. Stack. Call mg_tls_recv
3. TLS. mg_tls_recv copies data in its own recv buffer
4. TLS. mg_tls_recv decrypts the data and returns
5. Stack gets decrypted data and copies into c->recv

This PR unifies receive path to all TLS implementations by creating `c->rtls` IO buffer for the raw, encrypted data.

This change required OpenSSL implementation be changed, because OpenSSL accessed an underlying socket descriptor directly. Instead, OpenSSL must use our own low-level send/recv primitives for the low level IO - just like mbedTLS does. So, OpenSSL code was refactored: a custom BIO is created that implements the IO rather that relying on the socket. Theoretically, it makes it possible to use OpenSSL with the builtin stack - which is unlikely anyways.

### DIFF
--- a/src/tls_builtin.c
+++ b/src/tls_builtin.c
@@ -2,8 +2,12 @@
 
 #if MG_TLS == MG_TLS_BUILTIN
 struct tls_data {
-  struct mg_iobuf send;
-  struct mg_iobuf recv;
+  uint8_t client_random[32];  // From client hello
+  uint8_t client_pub[32];     // From client hello
+};
+struct tls_ctx {
+  struct mg_iobuf server_cert;  // Decoded server certificate
+  struct mg_iobuf server_key;   // Decoded server private key
 };
 
 #define MG_LOAD_BE16(p) ((uint16_t) ((MG_U8P(p)[0] << 8U) | MG_U8P(p)[1]))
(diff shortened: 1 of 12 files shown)


src/tls_builtin.c
```diff
@@ -56,8 +62,8 @@ void mg_tls_init(struct mg_connection *c, const struct mg_tls_opts *opts) {
 void mg_tls_free(struct mg_connection *c) {
   struct tls_data *tls = c->tls;
   if (tls != NULL) {
-    mg_iobuf_free(&tls->send);
-    mg_iobuf_free(&tls->recv);
+    // mg_iobuf_free(&tls->send);
+    // mg_iobuf_free(&tls->recv);
   }
```
scaprile, line 67: What is the reason for this ?

scaprile (approved):
AFAICU, LGTM

src/tls_builtin.c
cpq: Ah, just a leftover. tls_builtin.c will be totally replaced soon, so does not matter, really.

PR #2535, opened 2023-12-17
Move clock_init() to hal.h

cpq:
Some other changes apart of moving:

1. Changed non-portable asm("DSB") to the portable __DSB()
2. Moved syscfg init to the ethernet_init() - cause ethernet init uses syscfg
3. Moved SystemInit() to  main.c - cause it became really tiny

### DIFF
--- a/examples/stm32/nucleo-h743zi-make-baremetal-builtin/main.c
+++ b/examples/stm32/nucleo-h743zi-make-baremetal-builtin/main.c
@@ -32,6 +32,8 @@ static void timer_fn(void *arg) {
            ifp->ndrop, ifp->nerr));
 }
 
+// Device initialisation flow is in startup_XXX.s, provided by device vendor.
+// It inits data/bss, IRQ table, then calls SystemInit(), then calls main()
 int main(void) {
   gpio_output(LED);               // Setup green LED
   uart_init(UART_DEBUG, 115200);  // Initialise debug printf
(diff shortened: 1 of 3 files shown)


examples/stm32/nucleo-h743zi-make-baremetal-builtin/main.c
```diff
@@ -7,6 +7,15 @@
...
 
+uint32_t SystemCoreClock;  // Holds system core clock in Hz. Required by CMSIS
+
+void SystemInit(void) {             // Called automatically by startup code
+  clock_init();                     // Core clock to 480 MHz
+  rng_init();                       // Initialise random number generator
+  SystemCoreClock = CPU_FREQUENCY;  // defined in hal.h
+  SysTick_Config(SystemCoreClock / 1000);  // Sys tick every 1ms
+}
+
```
scaprile, line 18: Didn't we agree to keep this in sysinit.c ?

scaprile (changes requested):

scaprile:
> 1. Changed non-portable asm("DSB") to the portable __DSB()

Yes! We started with no CMSIS and since baremetal uses newlib and armgcc I didn't care, but we should use CMSIS macros in hal.h as a good practice.

> 2\. Moved syscfg init to the ethernet_init() - cause ethernet init uses syscfg

In fact I'm not sure if that is not something global that is better served in SystemInit. I'm OK with also having it at _init() functions as it is a dependency, in fact I think we do that with the UART to avoid depending on the order of calling. I'll check

scaprile (approved):

PR #2543, opened 2023-12-21
Remove gettimeofday implementation, to avoid clashes

cpq:


scaprile:
#### Please rebase, this branch stemmed from an old trunk and that is the reason why pico-rndis-device and h723 mqtt device tests are failing

scaprile (approved):

PR #2550, opened 2023-12-29
FIX empty payload MQTT message over UDP results in unnecessary closure

Ercsion:


Ercsion:
If the MQTT UDP client publishes a message without payload, `mg_io_send()` will call `sendto()` and return `-1` 
because the 2nd and 3rd arguments are `0`. At this point, the `errno` is `0`. Subsequently, the return value of 
`mg_io_send()`, which is `-1`, is passed to `iolog()`. As a result, `is_closing` will be set to `1`, triggering the `MG_EV_CLOSE` event.

Just ignore the empty payload to avoid unnecessary close.

scaprile:
MQTT does not work over UDP,

Ercsion:
> MQTT does not work over UDP,

take a look at this link [Welcome to MQTT/UDP]( https://mqtt-udp.readthedocs.io/en/latest/)

scaprile:
There are a lot of issues when the transport is not TCP that we do not took into consideration when designing and writing our MQTT implementation.

Ercsion:
mongoose works fine with MQTT/ UDP in our embedded project for now.

After all, you are the BOSS.😃

scaprile:
I fwd this to internal discussion.
It _may_ work, but we can't assure it _will_ work. You can always keep your branch if we decide not to follow this path.

PR #2562, opened 2024-01-09
RT1020 and RT1060 OTA

robertc2000:


examples/nxp/rt1020-evk-make-baremetal-builtin/Makefile
```diff
@@ -4,7 +4,14 @@ CFLAGS += -g3 -Os -ffunction-sections -fdata-sections
 CFLAGS += -I. -Icmsis_core/CMSIS/Core/Include -Icmsis_mcu/devices/MIMXRT1021 #-DCPU_MIMXRT1021DAG5A
 CFLAGS += -mcpu=cortex-m7 -mthumb -mfloat-abi=hard -mfpu=fpv5-d16 $(CFLAGS_EXTRA)
 LDSCRIPT = link_ram.ld
-LDFLAGS ?= -T$(LDSCRIPT) -nostdlib -nostartfiles --specs nano.specs -lc -lgcc -Wl,--gc-sections -Wl,-Map=$@.map 
+LDFLAGS ?= -T$(LDSCRIPT) -nostdlib -nostartfiles --specs nano.specs -lc -lgcc -Wl,--gc-sections -Wl,-Map=$@.map
+
+PACK = ./pack             # Utility to pack files into a packed filesystem
```
cpq, line 9: Don't repack the UI! None of the embedded projects do. We simply symlink it from the device dashboard, and use a pre-built packed_fs.c. So, ditch this whole section please.

scaprile (commented):
RT1060 examples are now running from flash, please rebase and update your code

scaprile (commented):
RT1020 examples moved to flash, please rebase and update

examples/nxp/rt1060-evk-make-baremetal-builtin/main.c
```diff
@@ -31,16 +32,57 @@ static void timer_fn(void *arg) {
...
+  return mg_millis();
+}
+#endif
+
 int main(void) {
-  gpio_output(LED);               // Setup blue LED
+  gpio_output(LED);               // Setup green LED
   uart_init(UART_DEBUG, 115200);  // Initialise debug printf
   ethernet_init();                // Initialise ethernet pins
-  MG_INFO(("Starting, CPU freq %g MHz", (double) SystemCoreClock / 1000000));
```
scaprile, line 38: Why move it ? All examples have it there

examples/nxp/rt1060-evk-make-baremetal-builtin/main.c
```diff
@@ -31,16 +32,57 @@ static void timer_fn(void *arg) {
...
 
   struct mg_mgr mgr;        // Initialise
   mg_mgr_init(&mgr);        // Mongoose event manager
   mg_log_set(MG_LL_DEBUG);  // Set log level
 
+  MG_INFO(("Starting, CPU freq %g MHz", (double) SystemCoreClock / 1000000));
```
scaprile, line 59: see above

examples/nxp/rt1060-evk-make-baremetal-builtin/main.c
```diff
@@ -65,6 +107,6 @@ int main(void) {
   for (;;) {
     mg_mgr_poll(&mgr, 0);
   }
-
```
scaprile, line 68: Why ? All examples have this line, it separates the loop from the (useless) rest

src/device_imxrt.c
```diff
@@ -0,0 +1,278 @@
+#include "device.h"
+#include "log.h"
+
+#if MG_DEVICE == MG_DEVICE_RT1020 || MG_DEVICE == MG_DEVICE_RT1060
+
+#include <flexspi.h>  /**************************************************************************************************************************/
```
scaprile, line 6: This is going to be a heck of a problem...
There is quite likely no file with that name in Xpresso, and probably file contents I got from several files and built just one...
This is one of the reasons why I think this has to be part of Mongoose.
@cpq ?

Some clashing may also occur with Xpresso, for the same reasons in the comment just bellow this line

src/device_imxrt.c
```diff
@@ -0,0 +1,278 @@
...
+// identical for both 1020 and 1060, they have the same controller. This should be part of Mongoose, not read from the example
+
+
+// these are defined in the device CMSIS .h file, so examples will have their own, we should use a different name or #undef
+// hence, the example flexspi.h does not have these macros, that is NXP code carrying their license
+// Mongoose does not include any CMSIS file but these would go into mongoose.h and the user including mongoose.h and the device .h both definitions would clash
```
scaprile, line 12: This is to meditate on...
@cpq ?

src/device_imxrt.c
```diff
@@ -0,0 +1,278 @@
...
+typedef struct {
+  uint32_t version;
+  int (*init)(uint32_t instance, flexspi_nor_config_t *config);
+  int (*program)(uint32_t instance, flexspi_nor_config_t *config,
+                      uint32_t dst_addr, const uint32_t *src);
+  int (*erase_all)(uint32_t instance, flexspi_nor_config_t *config);
+  int (*erase)(uint32_t instance, flexspi_nor_config_t *config,
+                    uint32_t start, uint32_t lengthInBytes);
+  int (*read)(uint32_t instance, flexspi_nor_config_t *config,
+                   uint32_t *dst, uint32_t addr, uint32_t lengthInBytes);
```
scaprile, line 54: These will also probably clash with the xip.c code in Xpresso, we should come up with our own naming or adding the usual prefix combo

src/device_imxrt.c
```diff
@@ -0,0 +1,278 @@
...
+typedef struct {
+  const uint32_t version;  // Bootloader version number
+  const char *copyright;   // Bootloader Copyright
+  void (*runBootloader)(
+      void *arg);             // Function to start the bootloader executing
+  const uint32_t *reserved0;  // Reserved
+  const flexspi_nor_driver_interface_t
+      *flexSpiNorDriver;         // FlexSPI NOR Flash API
+  const uint32_t *reserved1[2];  // Reserved
+} bootloader_api_entry_t;
```
scaprile, line 73: This is less likely to cause trouble

src/device_imxrt.c
```diff
@@ -0,0 +1,278 @@
...
+MG_IRAM bool mg_flash_erase(void *addr) {
+  flexspi_nor_config_t config;
+  if (flexspi_nor_get_config(&config) != 0) {
+    return false;
+  }
+  if (flash_page_start(addr) == false) {
+    MG_ERROR(("%p is not on a sector boundary", addr));
+    return false;
+  }
+  // Note: Interrupts must be disabled before any call to the ROM API on RT1020
```
scaprile, line 162: On 1060 too

src/device_imxrt.c
```diff
@@ -0,0 +1,278 @@
...
+  if (!s_flash_irq_disabled) {
+    asm("cpsid i");
+  }
+  bool ok = (flexspi_nor->erase(FLEXSPI_NOR_INSTANCE, &config, (uint32_t) addr,
+                                mg_flash_sector_size()) == 0);
+  if (!s_flash_irq_disabled) {
+    asm("cpsie i");  // // Reenable them after the call
+  }
+  MG_DEBUG(("Sector starting at %p erasure: %s", addr, ok ? "ok" : "fail"));
```
scaprile, line 171: LOG from IRAM

src/device_imxrt.c
```diff
@@ -0,0 +1,278 @@
...
+  return true;
+}
+
+static inline void spin(volatile uint32_t count) {
+  while (count--) (void) 0;
+}
+
+static inline void flash_wait(void) {
+  while ((*((volatile uint32_t *)(0x402A8000 + 0xE0)) & MG_BIT(1)) == 0) spin(1);
+}
```
scaprile, line 185: Did this work fine ?

src/device_imxrt.c
```diff
@@ -0,0 +1,278 @@
...
+  uint32_t status =
+      flexspi_nor->get_config(FLEXSPI_NOR_INSTANCE, config, options);
+  if (!s_flash_irq_disabled) {
+    asm("cpsie i");
+  }
+  if (status) {
+    MG_ERROR(("Failed to extract flash configuration: status %u", status));
+  }
+  return status;
+}
```
scaprile, line 142: too many checks. It does no harm to disable already disabled IRQs, if you need them to be disabled just do it. These checks only make sense when for some reason you need to keep track of the IRQ state before entering your code and restore it at exit. I guess you do this because sometimes you don't want IRQs to be reenabled.

This has to be written in a compiler agnostic way... plus... we already have a macro for these:
```
		#define MG_ARM_DISABLE_IRQ() asm volatile("cpsid i" : : : "memory")
		#define MG_ARM_ENABLE_IRQ() asm volatile("cpsie i" : : : "memory")
```
These might not be compiler agnostic, are they ?

We also have these:
```
	#if defined(__CC_ARM)
		#define MG_DSB() __dsb(0xf)
	#elif defined(__ARMCC_VERSION)
		#define MG_DSB() __builtin_arm_dsb(0xf)
	#elif defined(__GNUC__) && defined(__arm__) && defined(__thumb__)
		#define MG_DSB() asm("DSB 0xf")
	#elif defined(__ICCARM__)
		#define MG_DSB() __iar_builtin_DSB()
	#else
		#define MG_DSB()
	#endif
```
and behold our own CMSIS_compiler, a reduced one

(And this calls for a revision on what we are doing in other modules...)

src/device_imxrt.c
```diff
@@ -0,0 +1,278 @@
...
+    asm("cpsid i");
+  }
+  bool ok = (flexspi_nor->erase(FLEXSPI_NOR_INSTANCE, &config, (uint32_t) addr,
+                                mg_flash_sector_size()) == 0);
+  if (!s_flash_irq_disabled) {
+    asm("cpsie i");  // // Reenable them after the call
+  }
+  MG_DEBUG(("Sector starting at %p erasure: %s", addr, ok ? "ok" : "fail"));
+  return ok;
+}
```
scaprile, line 173: Regarding this function:
- enabling interrupts there is risky if you are erasing sector 0, thing you take care of in other function. I think it is better to handle IRQs at the calling function
- that MG_LOG is a) calling for trouble, and b) a nice delay. Like it or not, you are running flash code, the log function is not in RAM. It takes tens of milliseconds.

src/device.h
```diff
@@ -8,13 +8,22 @@
...
 #define MG_DEVICE_CH32V307 100  // WCH CH32V307
 #define MG_DEVICE_CUSTOM 1000   // Custom implementation
 
 #ifndef MG_DEVICE
 #define MG_DEVICE MG_DEVICE_NONE
 #endif
 
+#if MG_DEVICE == MG_DEVICE_RT1020 || MG_DEVICE == MG_DEVICE_RT1060
+#define MG_FLASH_OFFSET 0x60000000  // Offset to memory mapped flash
+#define MG_FLASH_CODE_OFFSET 0x2000 // Offset to code from start of flash
```
scaprile, line 22: Why not use mg_flash_start() to handle this ? (same in device_imxrt.c)

src/device_imxrt.c
```diff
@@ -0,0 +1,278 @@
...
+
+#define bootloader (*(bootloader_api_entry_t **) (0x0020001c))
+#define flexspi_nor bootloader->flexSpiNorDriver
+
+
+static bool s_flash_irq_disabled;
+
+MG_IRAM void *mg_flash_start(void) {
+  return (void *) 0x0;
+}
```
scaprile, line 83: What if mg_flash_start() returns 0x60000000 ?

src/ota_flash.c
```diff
@@ -74,7 +74,7 @@ bool mg_ota_end(void) {
   bool ok = false;
   if (s_size) {
     size_t size = s_addr - base;
-    uint32_t crc32 = mg_crc32(0, base, s_size);
+    uint32_t crc32 = mg_crc32(0, base + MG_FLASH_OFFSET, s_size);
```
scaprile, line 77: Is this macro really needed ? (see other files, I don't in which order GH will present this to you...)

src/ota_flash.c
```diff
@@ -187,11 +187,11 @@ MG_IRAM void mg_ota_boot(void) {
...
+      mg_flash_write(tmpsector, partition1 + ofs + MG_FLASH_OFFSET, ss);
       // mg_flash_erase(partition1 + ofs);
-      mg_flash_write(partition1 + ofs, partition2 + ofs, ss);
+      mg_flash_write(partition1 + ofs, partition2 + ofs + MG_FLASH_OFFSET, ss);
       // mg_flash_erase(partition2 + ofs);
-      mg_flash_write(partition2 + ofs, tmpsector, ss);
+      mg_flash_write(partition2 + ofs, tmpsector + MG_FLASH_OFFSET, ss);
     }
     mg_device_reset();
   }
```
scaprile, line 197: Did you actually try to swap partitions and succeed ?
Log calls in mg_flash_erase() will eventually mean you are calling flash that has been erased...

examples/nxp/rt1020-evk-make-baremetal-builtin/main.c
```diff
@@ -31,41 +31,57 @@ static void timer_fn(void *arg) {
...
-          c, 200, "", "%s",
-          "<html><head><link rel='icon' href='data:;base64,='></head><body>"
-          "<h1>Welcome to Mongoose</h1>"
-          "See <a href=/api/hello>/api/hello</a> for REST example"
-          "</body></html>");
-    } else {  // All other URIs
-      mg_http_reply(c, 404, "", "Not Found\n");
-    }
-  }
+#ifdef MQTT_DASHBOARD
```
cpq, line 34: Not in this PR:
We have those no-op stubs defined for may projects that don't have / don't use real hardware.
I suggest to move these no-op stubs to the mqtt-dashboard/net.c itself - that would be a default implementation, no-op. Wrap it into conditionals to provide an ability  to redefine.

mqtt-dashboard/net.c:

```c
#ifndef ENABLE_CUSTOM_HAL
bool hal_gpio_write(....) {
  MG_DEBUG(("pin: .., val:...", ...));
}
...
#endif
```

src/device_imxrt.c
cpq: This include looks dangerous to me.
Generally, we do not include any device-specific files from an SDK. For two reasons:
1. To be independent from any 3rd party source
2. To protect from any possible clashes

src/device_imxrt.c
cpq: Ditto - I do not feel easy about that include.

src/device_imxrt.c
cpq: Yes.
We should redefine our own structure.
The same way we do e.g. for ST Ethernet driver, or ST H5/H7 devices for flashing - we do not rely on any SDK sources, but define addresses/structures ourselves.

mongoose.h
```diff
@@ -2677,13 +2677,22 @@ MG_IRAM void mg_ota_boot(void);  // Bootloader function
...
+#define MG_DEVICE_RT1020 3      // IMXRT1020
+#define MG_DEVICE_RT1060 4      // IMXRT1060
 #define MG_DEVICE_CH32V307 100  // WCH CH32V307
 #define MG_DEVICE_CUSTOM 1000   // Custom implementation
 
 #ifndef MG_DEVICE
 #define MG_DEVICE MG_DEVICE_NONE
 #endif
 
+#if MG_DEVICE == MG_DEVICE_RT1020 || MG_DEVICE == MG_DEVICE_RT1060
```
cpq, line 2689: Why do we need that?

Flash layout is supposed to be described by the mg_flash_* API. Having a define which kinda does the same is confusing. I am not sure we should have this.

src/device.h
cpq: Ditto. mg_flash_* API was added to fully describe flash layout. We must not have this extra stuff.

cpq (changes requested):

examples/nxp/rt1060-evk-make-baremetal-builtin/link.ld
```diff
@@ -2,7 +2,7 @@ ENTRY(Reset_Handler);
 MEMORY {
   flash_hdr(rx)  : ORIGIN = 0x60000000, LENGTH = 8k
   flash_irq(rx)  : ORIGIN = 0x60002000, LENGTH = 1k
-  flash_code(rx) : ORIGIN = 0x60002400, LENGTH = 8183k
+  flash_code(rx) : ORIGIN = 0x60002400, LENGTH = 8179k
```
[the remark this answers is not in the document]
robertc2000: We use the last sector (4k) of the first half of the flash as a temporary sector in order to do the firmware swap. This is why, the firmware should be 4k less in size (from 8183k to 8179k).

examples/nxp/rt1060-evk-make-baremetal-builtin/main.c
robertc2000: My bad, I didn't notice... I'll put it back

src/device_imxrt.c
robertc2000: Yes, you are right. Initially, there was the problem that the flash is offset by this value (0x60000000). So, there appeared two cases: when you wanted to write (using the ROM API), you had to translate this address by subtracting 0x60000000 from it. The other case was when you wanted to read and it this case you just accessed it at 0x60000000 (no subtraction required).

What I wanted to do by using that macro was to avoid over-complicating myself in the mg_flash_write / mg_flash_erase functions with having to worry about translating the write address when using the ROM API, so I added that macro which worked for any case (for flashes that require no translation, like ST, it's value is 0 and changes nothing)

Anyway, I realise now that it can work even without adding the macro. We'll stick to the original mg_flash_* API like Sergey commented above. I'll make the address translations (for ROM API calls to write/erase) in the mg_flash_write/erase functions and leave the rest of the code untouched.

src/ota_flash.c
robertc2000: Solved, see above

src/ota_flash.c
robertc2000: Before we start swapping the partitions, we disable the logs, so the macro has no effect.

mongoose.h
robertc2000: Solved

src/device_imxrt.c
robertc2000: Normally, before the firmware swap, we make sure to disable the logs in order to make sure we don't call any log functions after erasing them. If logging is disabled, this macro has no effect.

examples/nxp/rt1060-evk-make-baremetal-builtin/link.ld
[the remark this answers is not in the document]
scaprile: YALMT (Yet Another Linker Magic Trick), this is not portable, what happens if Mongoose is built with MCUXpresso or Keil ? --> OTA does not work. (same with IAR and other tools we do not actively support)
Are we "reserving" room in a similar way in other architectures ? We should not to that.
The cleanest portable way I can think of is to define a constant array of 0xFF when OTA is enabled, even though that will extend Mongoose footprint by 4KB (though that is actually true, it affects flashing time...)

src/ota_flash.c
scaprile: The macro always has an effect, code has been compiled in with log enabled so that macro generated some code.
Does that RAM code actually check for the log level and then skips jumping to the log function, whatever it is ? (The user can change it)
Then that is cool and quite clever.
Bear in mind that all those strings are quite likely using valuable RAM space, nevetheless. (for the same reason)

examples/nxp/rt1020-evk-make-baremetal-builtin/flash_image.c
scaprile, line 1: Please go back to the original file and don't add unnecessary changes, that will be reverted back when someone runs clang-format and f-starting goes before h-starting again...

src/ota_flash.c
robertc2000: Yes, the macro will first check the log level value and then decides if it the logging function will be executed. In the case of MG_LL_NONE (we set logging to this value in mg_ota_boot, right before swapping), no logging function will be executed.

src/device_imxrt.c
scaprile: If they did not want to wait inside the ROM code to return quickly and be able to do something else, they could have just documented that...

src/device_imxrt.c
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  uint32_t *dst = (uint32_t *) addr;
+  uint32_t *src = (uint32_t *) buf;
+  uint32_t *end = (uint32_t *) ((char *) buf + len);
+  bool ok = true;
+
+  // Note: If we overwrite the flash irq section of the image, we must also
+  // make sure interrupts are disabled and are not reenabled until we write
+  // this sector with another irq table.
```
scaprile, line 147: In fact, once we change this, new IRQs will be pointing to somewhere in the new image, so actually we can't reenable IRQs until the new image is able to run. I guess that means booting again, so that crude reality might simplify your handling of interrupts in the code.
Once this sector has been erased, there is no going back, bridges are burned, we either flash the whole thing and reboot or we are bricked...

src/device_imxrt_flexspi.h
scaprile, line 1: We need to either de-vendor-ize this file, and cesanta-ize / mongoose-ize it; make it our own. Most comments are plainly useless and that comment formatting is not our own
Should we also add `IMXRT` to the `MG`prefix ? @cpq ?

Now, with this file in mongoose.h, our baremetal examples don't need their flexspi.h file, flash_image.c can be reworked to use these built-in flexspi settings.

scaprile:
Managed to get rid of flexspi.h, please rebase

cpq (approved):

PR #2566, opened 2024-01-10
draft nucleo h755 dual core

lonejack:
This version works on a dual core uC.
I know that you use a different environment based with a smart makefile...
Anyway it works!
Please note that:

- the mongoose files are not symlinks;

After that you imported the project with STM32CubeIDE:
insert on CM4 project the sources and include about WebServer directory under CM4 project;

Furthermore, unfortunately the CM4 can't access to the unique ID. So I set some defines in the project(ioc) instead of UUID.
In the next days (depending on the workload) I'll make a version built with a simple makefile as you like.


scaprile:
> I know that you use a different environment based with a smart makefile...

No, we don't. All nucleo-*-cube-* examples are just Cube examples to be imported with Cube and run in Cube


> insert on CM4 project the sources and include about WebServer directory under CM4 project;

So how is this different (for Mongoose) from an F4 project (Cortex-M4) ? Like https://github.com/cesanta/mongoose/tree/master/examples/stm32/nucleo-f429zi-cube-freertos-builtin (besides using the STM32H driver instead of the STM32F driver...)

> * the mongoose files are not symlinks;

We use symlinks for a reason.

lonejack:
Final version ready with symlinks...
Ciao

lonejack:
Don't integrate this version. Some problem present.

PR #2581, opened 2024-01-20
Fix #2559: order of handlers: protocol first, user second

### LINKED ISSUE #2559: MG_EV_CLOSE is fired before MG_EV_HTTP_MSG
- My goal is: use mongoose as http client
- My actions were: used the http-client example with enabled ssl
- My expectation was: The MG_EV_HTTP_MSG fires before the MG_EV_CLOSE event
- The result I saw: The MG_EV_CLOSE fires before the MG_EV_HTTP_MSG event.

# Environment

- mongoose version: current master
- Compiler/IDE and SDK: gcc linux (ubuntu)
- Target hardware/board:  x86_64
- Connectivity chip/module: _?_
- Target RTOS/OS (if applicable): linux

I use mongoose in myMPD as reverse proxy. Since commit 144c2f443e8f1af4bf72229475cc5bd48d0c5856 the MG_EV_CLOSE event fires before the MG_EV_HTTP_MSG event when compiled with openssl. Reverting to a previous commit fixes that. I access some connection specific fn_data in the MG_EV_HTTP_MSG and it SEGFAULTS because I free it in the MG_EV_CLOSE event.

This issue can be reproduced with the http-client example.

Log
```
9a8168 4 sock.c:697:mg_mgr_poll         1 r- Tchrc
9a8169 3 sock.c:299:read_conn           1 0x5 snd 0/2048 rcv 36864/38912 n=1053 err=0
9a8169 4 sock.c:697:mg_mgr_poll         2 -- tchrc
9a8169 4 sock.c:697:mg_mgr_poll         1 r- Tchrc
9a8169 3 sock.c:299:read_conn           1 0x5 snd 0/2048 rcv 37917/38912 n=-1 err=0
CLOSED
HTTP MSG
9a8169 3 net.c:151:mg_close_conn        1 5 closed
9a816b 4 sock.c:697:mg_mgr_poll         2 -- tchrC
9a816b 3 net.c:151:mg_close_conn        2 4 closed
9a816b 3 net.c:249:mg_mgr_free          All connections closed
```

Source code
```
// Copyright (c) 2021 Cesanta Software Limited
// All rights reserved
//
// Example HTTP client. Connect to `s_url`, send request, wait for a response,
// print the response and exit.
// You can change `s_url` from the command line by executing: ./example YOUR_URL
//
// To enable SSL/TLS, , see https://mongoose.ws/tutorials/tls/#how-to-build

#include "mongoose.h"

// The very first web page in history. You can replace it from command line
static const char *s_url = "https://www.google.com/";
static const char *s_post_data = NULL;      // POST data
static const uint64_t s_timeout_ms = 1500;  // Connect timeout in milliseconds

// Print HTTP response and signal that we're done
static void fn(struct mg_connection *c, int ev, void *ev_data, void *fn_data) {
  if (ev == MG_EV_OPEN) {
    // Connection created. Store connect expiration time in c->data
    *(uint64_t *) c->data = mg_millis() + s_timeout_ms;
  } else if (ev == MG_EV_POLL) {
    if (mg_millis() > *(uint64_t *) c->data &&
        (c->is_connecting || c->is_resolving)) {
      mg_error(c, "Connect timeout");
    }
  } else if (ev == MG_EV_CONNECT) {
    // Connected to server. Extract host name from URL
    struct mg_str host = mg_url_host(s_url);

    if (mg_url_is_ssl(s_url)) {
      struct mg_tls_opts opts = {.ca = mg_unpacked("/certs/ca.pem"),
                                 .name = mg_url_host(s_url)};
      mg_tls_init(c, &opts);
    }

    // Send request
    int content_length = s_post_data ? strlen(s_post_data) : 0;
    mg_printf(c,
              "%s %s HTTP/1.0\r\n"
              "Host: %.*s\r\n"
              "Content-Type: octet-stream\r\n"
              "Content-Length: %d\r\n"
              "\r\n",
              s_post_data ? "POST" : "GET", mg_url_uri(s_url), (int) host.len,
              host.ptr, content_length);
    mg_send(c, s_post_data, content_length);
  } else if (ev == MG_EV_HTTP_MSG) {
    // Response is received. Print it
    //printf("%.*s", (int) hm->message.len, hm->message.ptr);
    printf("HTTP MSG\n");
    c->is_draining = 1;        // Tell mongoose to close this connection
    *(bool *) fn_data = true;  // Tell event loop to stop
  } else if (ev == MG_EV_ERROR) {
    *(bool *) fn_data = true;  // Error, tell event loop to stop
  } else if (ev == MG_EV_CLOSE) {
    printf("CLOSED\n");
  }
}

int main(int argc, char *argv[]) {
  const char *log_level = getenv("LOG_LEVEL");  // Allow user to set log level
  if (log_level == NULL) log_level = "4";       // Default is verbose

  struct mg_mgr mgr;              // Event manager
  bool done = false;              // Event handler flips it to true
  if (argc > 1) s_url = argv[1];  // Use URL provided in the command line
  mg_log_set(atoi(log_level));    // Set to 0 to disable debug
  mg_mgr_init(&mgr);              // Initialise event manager
  mg_http_connect(&mgr, s_url, fn, &done);  // Create client connection
  while (!done) mg_mgr_poll(&mgr, 50);      // Event manager loops until 'done'
  mg_mgr_free(&mgr);                        // Free resources
  return 0;
}
```

cpq:


scaprile:
This breaks those examples where we want to process incoming messages ourselves. The protocol handler has a higher priority and deletes the data from the buffer before we get a chance to even see it.

For example:
```
[http-streaming-client]$ make
cc main.c mongoose.c  packed_fs.c       -W -Wall -Wextra -g -I.   -DMG_ENABLE_LINES=1 -DMG_ENABLE_PACKED_FS=1  -o example                                    
./example                    
<html><head></head><body><header>
...
</body></html>
[http-streaming-client]$ git checkout 2559 && make clean
Switched to branch '2559'
Your branch is up to date with 'origin/2559'.
rm -rf                    example                    *.o *.obj *.exe *.dSYM mbedtls
[http-streaming-client]$ make
cc main.c mongoose.c  packed_fs.c       -W -Wall -Wextra -g -I.   -DMG_ENABLE_LINES=1 -DMG_ENABLE_PACKED_FS=1  -o example                                    
./example                    
[http-streaming-client]$ 
```
No output !

scaprile (changes requested):
Changing the order of handlers can have many unexpected and undesired effects. We'd have to probably re-write a lot of stuff.
_I'd try to prevent sending EV_CLOSE when there is outstanding data, fire an EV_READ that will in this case end up being an EV_HTTP_MSG and then fire the close event._

Please see other comments
file-upload-single-post doesn't work, either, it doesn't even serve the info page.

cpq:
Fixed the example.

I believe that the order of handler "proto first, user second" is semantically correct.
I has been like that all the time! It was changed, I believe, to make this streaming client work. But it was the wrong decision, IMO

scaprile:
file-upload-single-post doesn't work, either, it doesn't even serve the info page.
Those are only two I tested.

cpq:
Doh... File uploads is kind of basics! We must test those automatically.
Anyhow, will look at file uploads.

scaprile:
The rest of the examples tested OK.
The recently added file-transfer example will quite likely have the same issues.

cpq:
Fixed!

The idea is to use `mg_listen()` instead of `mg_http_listen()`. This way we just get rid of the HTTP protocol handler!
I assume that to be a rare case. The uploader could run on a separate port, to avoid clashes with the rest of the UI.

scaprile:
Having a separate handler because we are bypassing the default handler to shape buffering in embedded is understandable

scaprile (approved):

PR #2587, opened 2024-01-23
Fix #2570: close only when TLS buffers drained

### LINKED ISSUE #2570: POST and TLS client
- My goal is:  To retrieve data from an https server (using openssl).
- My actions were:  Send a POST to https://httpbin.org/post,  20 times in a row, and verify response.
- My expectation was:  Each post responds correctly within 1 sec.
- The result I saw:    2 posts respond correctly, then 1 post times-out, then 1 post is correct, then 2 fail, etc.
- My question is:   Am I doing something wrong, or can this be fixed?
Public reproducible build:     https://github.com/jimellert/mongooseDemo1.git
Run test number 3 to reproduce the problem.

# Environment

- mongoose version:  7.12
- Compiler/IDE and SDK:   Visual Studio Community 2022 ver 17.8.3,   Visual C++
- Target hardware/board:   Dell Precision 7730,   Raspberry Pi 4 Model B Rev 1.4
- Connectivity chip/module: 
- Target RTOS/OS (if applicable):   Windows 10,   Ubuntu 22.04.3 LTS

cpq:
Mongoose reads raw encrypted data into the c->rtls buffer. Then, it feeds that data to the TLS library.
Currently, if server finished sending the data, it closes the connection. Mongoose reads data into c->rtls, so c->rtls has some data to be decrypted, then Mongoose reads 0 from the socket, indicating that server has closed the connection.
And Mongoose closed the connection too, despite it has some data in c->rtls yet to be decrypted!

This PR fixes that issue and decrypts everything before closing the connection.

### DIFF
--- a/src/sock.c
+++ b/src/sock.c
@@ -282,14 +282,13 @@ static void read_conn(struct mg_connection *c) {
       n = recv_raw(c, (char *) &c->rtls.buf[c->rtls.len],
                    c->rtls.size - c->rtls.len);
       // MG_DEBUG(("%lu %ld", c->id, n));
-      if (n == MG_IO_ERR) {
+      if (n == MG_IO_ERR && mg_tls_pending(c) == 0 && c->rtls.len == 0) {
+        // Close only if we have fully drained both raw (rtls) and TLS buffers
         c->is_closing = 1;
-      } else if (n > 0) {
-        c->rtls.len += (size_t) n;
+      } else {
+        if (n > 0) c->rtls.len += (size_t) n;
         if (c->is_tls_hs) mg_tls_handshake(c);
         if (c->is_tls_hs) return;
-        n = mg_tls_recv(c, buf, len);
-      } else if (n == MG_IO_WAIT) {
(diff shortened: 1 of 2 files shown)


src/sock.c
```diff
@@ -282,15 +282,14 @@ static void read_conn(struct mg_connection *c) {
...
-      } else if (n > 0) {
-        c->rtls.len += (size_t) n;
+      } else {
+        if (n > 0) c->rtls.len += (size_t) n;
         if (c->is_tls_hs) mg_tls_handshake(c);
         if (c->is_tls_hs) return;
         n = mg_tls_recv(c, buf, len);
-      } else if (n == MG_IO_WAIT) {
-        n = mg_tls_recv(c, buf, len);
       }
```
scaprile, line 293: mg_tls_handshake() used to execute on `n > 0`, now that path is taken as long as there is no IO_ERR; even on IO_WAIT, overriding the call to mg_tls_recv().
Looks cleaner, as long as that is what's intended.

scaprile (approved):

PR #2588, opened 2024-01-24
Add file transfer example

scaprile:
Known caveats:
- upload dir must exist, default dir is included in this PR
- If file exists, contents are appended (might be helpful to append logs)


examples/file-transfer/README.md
```diff
@@ -0,0 +1,47 @@
+# File Transfer
+
+This example contains minimal HTTP client and server.
+
+The client uploads a file to the server in a single POST, shaping traffic to send small data chunks.
+
+The server manually processes requests in order to be able to write as soon as data arrives, to avoid buffering a whole (possibly huge) file not fitting in RAM.
```
cpq, line 7: Please mention authentication, e.g. "Only an authenticated users can perform an upload."

cpq:
Looks great, thank you. 
Please merge.

examples/file-transfer/README.md
scaprile: Two paragraphs below that

PR #2596, opened 2024-02-01
Add helper mg_fs_ls()

cpq:


src/fs.c
```diff
@@ -72,3 +72,22 @@ bool mg_file_printf(struct mg_fs *fs, const char *path, const char *fmt, ...) {
   free(data);
   return result;
 }
+
+// This helper function allows to scan a filesystem in a sequential way,
+// without using callback function:
+//      char buf[100] = "";
+//      while (mg_fs_ls(fs, "./my_data/", buf, sizeof(buf()) { ... }
```
scaprile, line 79: `while (mg_fs_ls(&mg_fs_posix, "./", buf, sizeof(buf))) {`

PR #2598, opened 2024-02-03
Make enum's error code explicit

ryuukk:
This is easier to read when debugging code, log message display number, but documentation doesn't

### DIFF
--- a/src/event.h
+++ b/src/event.h
@@ -7,24 +7,24 @@ void mg_call(struct mg_connection *c, int ev, void *ev_data);
 void mg_error(struct mg_connection *c, const char *fmt, ...);
 
 enum {
-  MG_EV_ERROR,      // Error                        char *error_message
-  MG_EV_OPEN,       // Connection created           NULL
-  MG_EV_POLL,       // mg_mgr_poll iteration        uint64_t *uptime_millis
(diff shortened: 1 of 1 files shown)


scaprile:
enum's start at 0 except otherwise indicated, this is no doc flaw

scaprile:
We don't see a need for something like what you suggest. Thank you very much for your collaboration, we just don't feel inclined to include something like that.
I'm not saying you changed it. I'm saying it is that way and there is no point in putting those numbers there.

> This is easier to read when debugging code, log message display number, but documentation doesn't

You can write a function that translates event to string and logs that, if that is what you want. We haven't seem a need for that so far.
You shouldn't have a need to debug Mongoose. If you just happen to fall inside Mongoose, you can skip out of it.
You shouldn't have a need to do a manual conversion from number to event. We add new features to Mongoose every day and we don't need that.
Your event handler will only receive those events that correspond to its function, plus POLL, READ, WRITE.
To debug your event handler, you usually handle some events and if you need to know one of those happens, you either put a breakpoint and find yourself there; or printf some notice on that part of the handler. That is usually what some of us do; I personally would rather stop at a known place using breakpoints.

scaprile:
If you open an issue and provide all necessary information for us to reproduce it, we'll gladly fix it.
Thank you

PR #2599, opened 2024-02-04
Allow user specify file in mg_http_upload

cpq:
This PR also fixes a bug in `mg_path_is_sane()`, and adds unit tests for it

### DIFF
--- a/src/str.c
+++ b/src/str.c
@@ -189,8 +189,9 @@ void mg_unhex(const char *buf, size_t len, unsigned char *to) {
 
 bool mg_path_is_sane(const char *path) {
   const char *s = path;
+  if (path[0] == '.' && path[1] == '.') return false;  // Starts with ..
   for (; s[0] != '\0'; s++) {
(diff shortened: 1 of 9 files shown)


src/str.c
```diff
@@ -189,8 +189,9 @@ void mg_unhex(const char *buf, size_t len, unsigned char *to) {
 
 bool mg_path_is_sane(const char *path) {
   const char *s = path;
+  if (path[0] == '.' && path[1] == '.') return false;  // Starts with ..
   for (; s[0] != '\0'; s++) {
-    if (s == path || s[0] == '/' || s[0] == '\\') {  // Subdir?
+    if (s[0] == '/' || s[0] == '\\') {               // Subdir?
       if (s[1] == '.' && s[2] == '.') return false;  // Starts with ..
```
scaprile, line 195:
```
if (s[0] == MG_DIRSEP && s[1] == '.' && s[2] == '.') return false;  // Subdir with ..
```
? Bonus: the other slash is allowed

PR #2607, opened 2024-02-13
Fix #2603: TLS read path

### LINKED ISSUE #2603: Reading post data from https connection fails now and then
- My goal is: Read post requests from a https socket issued by javascript
- My actions were: read the data in `MG_EV_HTTP_MSG` with `struct mg_http_message *hm = (struct mg_http_message *) ev_data;` as used in your examples
- My expectation was: mongoose reads the complete post data
- The result I saw: 

```
694221 3 sock.c:441:accept_conn         16 6 accepted 192.168.88.144:53682 -> 0.0.0.0:8443
694221 3 tls_openssl.c:111:mg_tls_init  16 Setting TLS
694223 3 tls_openssl.c:190:mg_tls_init  16 SSL accept OK
694259 3 tls_openssl.c:200:mg_tls_hands 16 success
694259 3 sock.c:298:read_conn           16 0x6 snd 0/0 rcv 0/2048 n=-2 err=0
694259 3 sock.c:298:read_conn           16 0x6 snd 0/0 rcv 0/2048 n=557 err=0
```

Mongoose does not read the post data from the socket.

The problem occurs only:
- with latest master
- with an iPhone as client (tested with safari and firefox mobile)
- only with enabled ssl (openssl in my case)

Without SSL or on my Linux Desktop it works without problems.

The issue is reproduceable with your http-restful-server example. I added an index.html that posts data with the javascript fetch api to the `/api/stats` endpoint.

```
<html>
    <body>
        <p id="status"></p>
        <a href="javascript:post()">Post</a>
        <script>
            async function post() {
                document.getElementById('status').textContent = "Sending...";
                const body = 'test';
                const response = await fetch('/api/stats', {
                    method: 'POST',
                    mode: 'same-origin',
                    credentials: 'same-origin',
                    cache: 'no-store',
                    redirect: 'follow',
                    headers: {
                        'Content-Type': 'text/plain',
                        'Content-Length': body.length.toString()
                    },
                    body: body
                });
                if (response) {
                    document.getElementById('status').textContent = await response.text();
                }
                return false;
            }
        </script>
    </body>
</html>
```

Your example is compiled with: `make CFLAGS_EXTRA="-DMG_TLS=MG_TLS_OPENSSL -lssl -lcrypto"`

Some requests are answered as expected, but for others there are no replies from mongoose.

- My question is: seems to be a bug?

# Environment

- mongoose version: latest master
- Compiler/IDE and SDK: `gcc version 13.2.0 (Ubuntu 13.2.0-4ubuntu3)`
- Target hardware/board: `Linux t14 6.5.0-15-generic #15-Ubuntu SMP PREEMPT_DYNAMIC Tue Jan  9 17:03:36 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`

cpq:


src/sock.c
```diff
@@ -281,8 +283,7 @@ static void read_conn(struct mg_connection *c) {
       if (!ioalloc(c, &c->rtls)) return;
       n = recv_raw(c, (char *) &c->rtls.buf[c->rtls.len],
                    c->rtls.size - c->rtls.len);
-      // MG_DEBUG(("%lu %ld", c->id, n));
-      if (n == MG_IO_ERR && mg_tls_pending(c) == 0 && c->rtls.len == 0) {
+      if (n == MG_IO_ERR && c->rtls.len == 0) {
         // Close only if we have fully drained both raw (rtls) and TLS buffers
```
scaprile, line 287: Should we also check for mg_tls_pending() as before ? Otherwise I think the comment is not accurate

scaprile (approved):

src/sock.c
cpq: Good question - but I am unsure about that.
It looks like we might not need `mg_tls_pending()` at all.
TLS exchanges data by so-called "records", with "record length" specified at the beginning of the "record". so I expect TLS code to retrieve data from `c->rtls` by records. Meaning, if `c->rtls` is empty, there could not be anything pending. But that's a theory, that's why I did not remove `mg_tls_pending()` yet.
I suggest to give it some runway and see.

PR #2621, opened 2024-02-22
Adopt CCRH compiler

cpq:


src/util.h
```diff
@@ -39,9 +39,12 @@ uint64_t mg_now(void);     // Return milliseconds since Epoch
...
-#ifdef __GNUC__
+#if defined(__GNUC__)
 #define MG_ARM_DISABLE_IRQ() asm volatile("cpsid i" : : : "memory")
 #define MG_ARM_ENABLE_IRQ() asm volatile("cpsie i" : : : "memory")
+#elif defined(__CCRH__)
+#define MG_RH850_DISABLE_IRQ() __DI()
+#define MG_RH850_ENABLE_IRQ() __EI()
 #else
 #define MG_ARM_DISABLE_IRQ()
 #define MG_ARM_ENABLE_IRQ()
```
scaprile, line 50: Well, following this same criteria we should also define those 850 macros empty when CCRM is not present. And this also assumes no one will build 850 code using gcc...

scaprile (approved):

PR #2636, opened 2024-03-02
Increase local file buffer size in mg_http_upload()

cpq:


test/unit_test.c
```diff
@@ -3369,8 +3368,13 @@ int main(void) {
   test_http_no_content_length();
   test_http_pipeline();
   test_http_range();
+#ifndef LOCALHOST_ONLY
   test_sntp();
   test_mqtt();
+  test_http_client();
+#else
+  (void) test_sntp, (void) test_mqtt, (void) test_http_client;
+#endif
```
scaprile, line 3377: I guess this is a leftover from the Debian saga

cpq: Ah, yeah

PR #2639, opened 2024-03-02
Better acks

cpq:


cpq:
Committing with TBR (To Be Reviewed)

src/net_builtin.c
```diff
@@ -489,6 +490,14 @@ static void rx_udp(struct mg_tcpip_if *ifp, struct pkt *pkt) {
...
                      uint8_t flags, uint16_t sport, uint16_t dport,
                      uint32_t seq, uint32_t ack, const void *buf, size_t len) {
+#if 0
+  uint8_t opts[] = {2, 4, 5, 0xb4, 4, 2, 0, 0};  // MSS = 1460, SACK permitted
+  if (flags & TH_SYN) {
+    // Handshake? Set MSS
+    buf = opts;
+    len = sizeof(opts);
+  }
+#endif
```
scaprile, line 500: Perhaps this can be conditionally enabled ? (see other case)

src/net_builtin.c
```diff
@@ -501,6 +510,8 @@ static size_t tx_tcp(struct mg_tcpip_if *ifp, uint8_t *dst_mac, uint32_t dst_ip,
   tcp->flags = flags;
   tcp->win = mg_htons(MIP_TCP_WIN);
   tcp->off = (uint8_t) (sizeof(*tcp) / 4 << 4);
+  // if (flags & TH_SYN) tcp->off = 0x70;  // Handshake? header size 28 bytes
+
```
scaprile, line 514: Perhaps this can be conditionally enabled ? (follows from other case)
Being in two different places makes it easy to overlook. I suggest we either remove these or leave them with a conditional, whether we use it or not.

src/net_builtin.c
```diff
@@ -654,16 +665,17 @@ static void read_conn(struct mg_connection *c, struct pkt *pkt) {
     MG_VERBOSE(("%lu SEQ %x -> %x", c->id, mg_htonl(pkt->tcp->seq), s->ack));
     // Advance ACK counter
     s->ack = (uint32_t) (mg_htonl(pkt->tcp->seq) + pkt->pay.len);
-    s->noack += pkt->pay.len;
-    if (s->noack + 1500 >= MIP_TCP_WIN) {
+    s->unacked += pkt->pay.len;
+    // size_t diff = s->acked <= s->ack ? s->ack - s->acked : s->ack;
```
scaprile, line 669: confusing "comment"...

src/net_builtin.c
```diff
@@ -654,16 +665,17 @@ static void read_conn(struct mg_connection *c, struct pkt *pkt) {
...
+    if (s->unacked > MIP_TCP_WIN / 2 && s->acked != s->ack) {
       // Send ACK immediately
-      MG_VERBOSE(("  imm ACK", c->id, mg_htonl(pkt->tcp->seq), s->ack));
+      MG_VERBOSE(("%lu imm ACK %lu", c->id, s->acked));
```
scaprile, line 672: `->acked` is the last sent ACK, current data being acked lies within `->acked` and `->ack`. We will actually send `->ack` in our next statement

src/net_builtin.c
```diff
@@ -654,16 +665,17 @@ static void read_conn(struct mg_connection *c, struct pkt *pkt) {
     MG_VERBOSE(("%lu SEQ %x -> %x", c->id, mg_htonl(pkt->tcp->seq), s->ack));
     // Advance ACK counter
     s->ack = (uint32_t) (mg_htonl(pkt->tcp->seq) + pkt->pay.len);
-    s->noack += pkt->pay.len;
-    if (s->noack + 1500 >= MIP_TCP_WIN) {
+    s->unacked += pkt->pay.len;
+    // size_t diff = s->acked <= s->ack ? s->ack - s->acked : s->ack;
+    if (s->unacked > MIP_TCP_WIN / 2 && s->acked != s->ack) {
```
scaprile, line 670: I think this right side check is needed because we don't zero `->unacked` when we ACK on timeout.

My belief is that `unacked` will slowly grow and reach `MIP_TCP_WIN / 2` while ACKs are being sent on timeouts (usual traffic) as we've been doing. Then next frame received will trigger an unnecessary immediate ACK. Nothing to worry about, though.

I still fail to see the mechanism for those many dup acks in a row, so let's not make changes here yet.

PR #2642, opened 2024-03-10
Add driver init to mg_mgr_init()

cpq:


src/config.h
```diff
@@ -157,3 +157,33 @@
...
+#define MG_MAC_ADDRESS_STM32_UID(ADDR)                       \
+  {                                                          \
+    2, ((uint8_t *) (ADDR))[0] ^ ((uint8_t *) (ADDR))[1],    \
+        ((uint8_t *) (ADDR))[2] ^ ((uint8_t *) (ADDR))[3],   \
+        ((uint8_t *) (ADDR))[4] ^ ((uint8_t *) (ADDR))[5],   \
+        ((uint8_t *) (ADDR))[6] ^ ((uint8_t *) (ADDR))[7] ^  \
+            ((uint8_t *) (ADDR))[8],                         \
+        ((uint8_t *) (ADDR))[9] ^ ((uint8_t *) (ADDR))[10] ^ \
+            ((uint8_t *) (ADDR))[11]                         \
+  }\
```
scaprile, line 189: Some STM32s do not allow byte reading, some Hs require word reading.
This actually depends on the processor. We can probably add this to the driver header files and deal with processor differences there.
I don't see this belonging to config.h

src/config.h
```diff
@@ -157,3 +157,33 @@
 #ifndef MG_ENABLE_PROFILE
 #define MG_ENABLE_PROFILE 0
 #endif
+
+#ifndef MG_ENABLE_DRIVER_INIT
```
scaprile, line 161: driver... what driver ? flash driver ? tcpip driver ? future driver ?
This is tempting but I think we have to prefix with TCPIP, as the drivers themselves are, if not for clarity just for consistency

src/drivers/stm32f.h
```diff
@@ -16,3 +16,42 @@ struct mg_tcpip_driver_stm32f_data {
...
+  } while (0)
+
+#define MG_TCPIP_DRIVER_INIT(mgr)                            \
+  do {                                                       \
+    struct mg_tcpip_if mif = {                               \
+        .mac = MG_MAC_ADDRESS,                               \
+    };                                                       \
+    mg_tcpip_init((mgr), &mif);                              \
+    MG_INFO(("Driver: stm32fxx, MAC: %M", mg_print_mac, 0)); \
+  } while (0)
```
scaprile, line 57: To me this is too large and a big chunk will almost be the same for all processors.
I would just add this to net_builtin.c and have small macros with the particulars, as minimum code as possible, maybe static inlines, as this is a header file

src/net.c
```diff
@@ -276,4 +276,8 @@ void mg_mgr_init(struct mg_mgr *mgr) {
   mgr->dns4.url = "udp://8.8.8.8:53";
   mgr->dns6.url = "udp://[2001:4860:4860::8888]:53";
   mg_tls_ctx_init(mgr);
+
+#if (MG_ENABLE_DRIVER_INIT == 1) && defined(MG_TCPIP_DRIVER_INIT)
+  MG_TCPIP_DRIVER_INIT(mgr);
+#endif
```
scaprile, line 282: I think this does not belong here but in net_builtin.c
Here we can call that code when using our own stack, that is, here, we just call `mg_tcpip_init()` `#if MG_TCPIP_ENABLE`
`mg_tcpip_init` would take care to initialize the driver or not.

cpq: Kind of agree with that, but how `net_builtin.c` would know about which driver to init ?
My thinking is that we already have `MG_ENABLE_DRIVER_xxx`, and so each of those respective drivers would #define an init macro in its header file - that's exactly what I did in this PR

scaprile: How does it know now ?
It gets a struct that has a pointer to the driver "methods" struct. That is included by having the proper driver enabled. Enabling more than one driver won't work at all, though.

src/config.h
```diff
@@ -157,3 +157,33 @@
...
+#endif
+
+#ifndef MG_TCPIP_GW
+#define MG_TCPIP_GW 0
+#endif
+
+#define MG_MAC_ADDRESS_RANDOM \
+  { 0, 0, 0, 0, 0, 0 }
+
+#define MG_MAC_ADDRESS_STM32_UID(ADDR)                       \
```
robertc2000, line 180: I would move MG_MAC_ADDRESS_${DEVICE}_UID into drivers/${DEVICE}.h, since this is dependent on the platform and have it like before (without any arguments).

src/config.h
```diff
@@ -157,3 +157,33 @@
 #ifndef MG_ENABLE_PROFILE
 #define MG_ENABLE_PROFILE 0
 #endif
+
+#ifndef MG_ENABLE_DRIVER_INIT
+#define MG_ENABLE_DRIVER_INIT 1
+#endif
+
+#ifndef MG_TCPIP_IP
+#define MG_TCPIP_IP 0
```
robertc2000, line 166: I think we should replace this with `MG_TCPIP_IP MG_U32(0, 0, 0, 0)`. If the user wants a static IP, this way we give them a hint on how to define it

mongoose.h
```diff
@@ -833,6 +833,24 @@ struct timeval {
 #define MG_ENABLE_PROFILE 0
 #endif
 
+#ifndef MG_ENABLE_TCPIP_DRIVER_INIT    // mg_mgr_init() will also initialize
+#define MG_ENABLE_TCPIP_DRIVER_INIT 0  // enabled built-in driver for
```
cpq, line 837: I really would prefer that to be 1, and avoid setting it in the mongoose_custom.h

mongoose.h
```diff
@@ -833,6 +833,24 @@ struct timeval {
...
+#endif                                 // Mongoose built-in network stack
+
+#ifndef MG_TCPIP_IP  // e.g. mg_htonl(MG_U32(192, 168, 0, 223))
+#define MG_TCPIP_IP mg_htonl(MG_U32(0, 0, 0, 0))  // or leave as 0 for DHCP
```
cpq, line 841: mg_htonl() is a function, and Keil complains about non-constant initialiser.
Let's create a `MG_IP(a,b,c,d)` macro that creates a number directly in the network byte order?

scaprile:
> The ` MG_INFO(("MAC: %M. Waiting for IP...", mg_print_mac, mif->mac));` - let's just remove it.

We still need to wait on the ifc... shall we write an mg_function for that ? Then embedded and OS diverge again... unless that function checks the OS stack is up

PR #2647, opened 2024-03-14
Split PHY code into phy.{c,h}

cpq:


src/drivers/phy.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include "net_builtin.h"
+
+struct mg_phy {
+  uint16_t (*read_reg)(uint8_t addr, uint8_t reg);
+  void (*write_reg)(uint8_t addr, uint8_t reg, uint16_t value);
+};
+
+enum { MG_PHY_CONF_CLOCKING = 0, MG_PHY_CONF_CLOCKED = 1, MG_PHY_CONF_LEDS_AH = 2};
```
cpq, line 10: I don't understand those values...
Could you choose a bit more descriptive names, perhaps longer ones?
Or, add comments (more descriptive names are preferred, cause comments are not visible where those values are used)

PR #2648, opened 2024-03-14
Refactor autoinit code

cpq:


src/drivers/stm32h.h
```diff
@@ -12,26 +12,29 @@ struct mg_tcpip_driver_stm32h_data {
   //    100-150 MHz   HCLK/62        1
   //    20-35 MHz     HCLK/16        2
   //    35-60 MHz     HCLK/26        3
-  //    150-250 MHz   HCLK/102       4  <-- value for Nucleo-H* on max speed driven by HSI
-  //    250-300 MHz   HCLK/124       5  <-- value for Nucleo-H* on max speed driven by CSI
+  //    150-250 MHz   HCLK/102       4  <-- value for max speed HSI
+  //    250-300 MHz   HCLK/124       5  <-- value for Nucleo-H* on CSI
```
scaprile, line 16: First one was redundant, but second one now is imprecise, the processor con run from CSI but not necessarily at max speed...

src/drivers/stm32f.h
```diff
@@ -32,13 +32,19 @@ struct mg_tcpip_driver_stm32f_data {
...
-
-#define MG_TCPIP_DRIVER_CODE &mg_tcpip_driver_stm32f
-#define MG_TCPIP_DRIVER_NAME "stm32f"
+#define MG_TCPIP_DRIVER_INIT(mgr)                                 \
+  do {                                                            \
+    static struct mg_tcpip_driver_stm32f_data driver_data_;       \
+    static struct mg_tcpip_if mif_ = {.mac = MG_MAC_ADDRESS};     \
+    driver_data_.mdc_cr = MG_DRIVER_MDC_CR;                       \
+    driver_data_.phy_addr = MG_TCPIP_PHY_ADDR;                    \
+    mif_.ip = MG_TCPIP_IP;                                        \
```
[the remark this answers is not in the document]
cpq: Are you sure? It is code, not an initaliser.

src/drivers/stm32f.h
```diff
@@ -32,13 +32,19 @@ struct mg_tcpip_driver_stm32f_data {
...
-      .phy_addr = MG_TCPIP_PHY_ADDR,                        \
-  };
-
-#define MG_TCPIP_DRIVER_CODE &mg_tcpip_driver_stm32f
-#define MG_TCPIP_DRIVER_NAME "stm32f"
+#define MG_TCPIP_DRIVER_INIT(mgr)                                 \
+  do {                                                            \
+    static struct mg_tcpip_driver_stm32f_data driver_data_;       \
+    static struct mg_tcpip_if mif_ = {.mac = MG_MAC_ADDRESS};     \
+    driver_data_.mdc_cr = MG_DRIVER_MDC_CR;                       \
```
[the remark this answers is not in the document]
cpq: Yeah, agreed. The other way is to declare on-stack

```c
uint8_t mac_[6] = MG_MAC_ADDRESS;
```

And then memcpy() to mif_.mac.
But I still don't see how we can reliably pass UID_BASE to there anyway

src/drivers/stm32h.h
cpq: What would be the correct comment ?

src/drivers/stm32f.h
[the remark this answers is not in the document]
scaprile: Oops, I missed that.
You are safe with this one, but we still have the problem with .mac above

There was a reason why I did not followed this code approach... I can't remember now. I did it and then went back to the auto + copy to static approach.

src/drivers/stm32f.h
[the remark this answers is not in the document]
scaprile: Yeah, that is why I hardcoded the address in mongoose_custom.h
I don't like including CMSIS headers inside that, and again that is no guarantee, name might changed among different processors... they actually do with NXPs...

src/drivers/stm32h.h
scaprile: Maybe "max speed on HSI" "max speed on CSI" is concise and clear enough ? WDYT ?

cpq: You've got the right comment in the other PR!

examples/stm32/nucleo-f746zg-make-baremetal-builtin/main.c
```diff
@@ -41,23 +37,7 @@ int main(void) {
...
-                            // .gw = mg_htonl(MG_U32(192, 168, 0, 1)),
-                            .driver = &mg_tcpip_driver_stm32f,
-                            .driver_data = &driver_data};
-  mg_tcpip_init(&mgr, &mif);
-  mg_timer_add(&mgr, BLINK_PERIOD_MS, MG_TIMER_REPEAT, timer_fn, &mif);
-
-  MG_INFO(("MAC: %M. Waiting for IP...", mg_print_mac, mif.mac));
-  while (mif.state != MG_TCPIP_STATE_READY) {
-    mg_mgr_poll(&mgr, 0);
-  }
```
scaprile, line 60: We still need to wait, chatted on this maybe two days ago

src/net.c
```diff
@@ -275,8 +270,8 @@ void mg_mgr_init(struct mg_mgr *mgr) {
   // Ignore SIGPIPE signal, so if client cancels the request, it
   // won't kill the whole process.
   signal(SIGPIPE, SIG_IGN);
-#elif MG_ENABLE_TCPIP && MG_ENABLE_TCPIP_DRIVER_INIT
-  mg_tcpip_auto_init(mgr);
+#elif MG_ENABLE_TCPIP_DRIVER_INIT && defined(MG_TCPIP_DRIVER_INIT)
+  MG_TCPIP_DRIVER_INIT(mgr);
 #endif
```
scaprile, line 275: I don't think this should be here, it is strictly for MIP and so IMHO I think it belongs in net_builtin,c

cpq: Why? What's the point of adding a function what calls that macro?
What's the benefit of having extra code?

scaprile: Clarity ? Keeping stacks separate ? I see net.c as a layer for a generic interface with the underlying specific protocol stack. That clearly does not have drivers. Drivers belong to protocol stacks and should be initialized by protocol stacks.
We add another protocol stack, I don't know, some Matter-like, that will have its drivers and those will need initialization, that will happen in net_someothernetstack.c, for example.

examples/stm32/nucleo-f746zg-make-baremetal-builtin/main.c
scaprile: I beg to differ, clients will fail, and I pasted this very example where the device-dashboard SNTP request fails:
```
0      2 main.c:40:main                 Starting, CPU freq 216 MHz
6      2 net_builtin.c:1146:mg_tcpip_au Driver: stm32f, MAC: 02:33:47:5b:3e:32
d      2 main.c:54:main                 Initialising application...
13     3 net.c:202:mg_listen            1 0 http://0.0.0.0/
18     3 net.c:202:mg_listen            2 0 https://0.0.0.0/
1e     2 main.c:57:main                 Starting event loop
23     1 net_builtin.c:880:mg_tcpip_pol Network is down
28     3 net.c:177:mg_connect           3 -1 udp://time.google.com:123
2f     3 net.c:177:mg_connect           4 -1 udp://8.8.8.8:53
34     3 net_builtin.c:1018:mg_connect_ 4 0.0.0.0:61329 -> 8.8.8.8:53
3a     1 event.c:31:mg_error            4 -1 net down
3f     1 event.c:31:mg_error            4 -1 DNS send
44     1 event.c:31:mg_error            3 -1 DNS error
49     3 net.c:151:mg_close_conn        4 -1 closed
4e     3 net.c:151:mg_close_conn        3 -1 closed
3ee    1 net_builtin.c:880:mg_tcpip_pol Network is down
40b    2 main.c:31:timer_fn             Ethernet: down, IP: 0.0.0.0, rx:0, tx:0, dr:0, er:0
7d6    3 stm32f.c:207:mg_tcpip_driver_s Link is 100M full-duplex
7db    1 net_builtin.c:206:onstatechang Link up
7e0    3 net_builtin.c:306:tx_dhcp_disc DHCP discover sent. Our MAC: 02:33:47:5b:3e:32
7f3    2 main.c:31:timer_fn             Ethernet: up, IP: 0.0.0.0, rx:1, tx:1, dr:0, er:0
8f5    3 net_builtin.c:285:tx_dhcp_requ DHCP req sent
a15    2 net_builtin.c:413:rx_dhcp_clie Lease: 86400 sec (86402)
a1b    2 net_builtin.c:201:onstatechang READY, IP: 192.168.0.28
a21    2 net_builtin.c:202:onstatechang        GW: 192.168.0.1
a26    2 net_builtin.c:203:onstatechang       MAC: 02:33:47:5b:3e:32
bdb    2 main.c:31:timer_fn             Ethernet: ready, IP: 192.168.0.28, rx:5, tx:3, dr:0, er:0
fc3    2 main.c:31:timer_fn             Ethernet: ready, IP: 192.168.0.28, rx:8, tx:3, dr:0, er:0
```
No SNTP retries after this failure.
HTTP or MQTT clients will have an mg_connect before the main loop and will also fail.
Anything that does not implement error handling and retries on timeout, will fail
This of course does not happen with OSs
We also explicitly wait on the ESP32 for network

Should we wait somewhere else or queue requests ? Maybe

src/net.c
cpq: IMO adding more indirection here, would be less clear.
With those three lines, you see EXACTLY what is going on, and why.

What's the reason to add same three lines but calling some mg_tcpip_whatever function?

scaprile: They wouldn't be the same lines...
```
#elif MG_ENABLE_TCPIP           // if we are using our own stack, init it
  mg_tcpip_init();
#elif MG_ENABLE_WHATEVERSTACK   // following up on the example above
  mg_whateverstack_init();
#endif
```
That is all net.c needs to know. It does not know nor cares about drivers and their initialization
Now, net_builtin.c does know about MG_ENABLE_TCPIP_DRIVER_INIT, it is TCPIP-land, and it knows about drivers and stuff, so those famous three lines would be in net_builtin.c, inside  mg_tcpip_init()

Anyway, this is philosophical... for the sake of efficiency we should write the Ethernet driver here, and for the sake of separation we should have a gazillion modules and take forever to call each one... You surely know better where to draw the line.

examples/stm32/nucleo-f746zg-make-baremetal-builtin/main.c
cpq: Yeah right but this is a different problem that needs to be solved separately.
That extra loop, waiting for IP, was just hiding the problem.
We can't really block  forever waiting for IP. firmware code needs to run regardless.

scaprile: Yes, you are right.

cpq: And the reason you did not see SNTP request is that time timer function runs hourly. So when the first invocation fails, due to the DNS failure, it won't fire another one for 1h. Which should be fixed in the app logic, not in MIP.

scaprile: Yes, this is UDP so the app shouldn't be so naive. I can only say that it will fail in a systematic fashion. Perhaps we need an mg_netisup(), this same condition can arise in non-embedded. No buffer growth to check as in TCP
An HTTP client and MQTT client will probably have MIP retransmitting the sync, do we ? I don't remember.

cpq: IMO the SNTP logic should retry often (like every second) if sync never happened, and rarely (e.g. hourly or daily) when sync did happen.
HTTP and MQTT should just buffer data.

scaprile (approved):

PR #2650, opened 2024-03-15
Fix #2612 - struct mg_str::ptr -> buf

### LINKED ISSUE #2612: Change struct mg_str :: ptr to struct mg_str :: buf
That one was  bugging me for a long time.
Seemingly unimportant and unnecessary change - though I believe it is a net positive long term.
In addition, let's remove `const`: `const char *ptr` -> `char *buf`.
We could add `const` to `struct mg_str` when required.

cpq:


scaprile (approved):

scaprile (changes requested):
Some tests are actually failing for a real reason

PR #2657, opened 2024-03-22
Added ethernet driver for Infineon XMC4 boards

robertc2000:


src/drivers/xmc.c
```diff
@@ -0,0 +1,208 @@
...
+  //Enable interrupts
+  ETH0->INTERRUPT_ENABLE = ETH_INTERRUPT_ENABLE_NIE_Msk |
+      ETH_INTERRUPT_ENABLE_RIE_Msk | ETH_INTERRUPT_ENABLE_TIE_Msk;
+
+  // Enable MAC transmission and reception
+  ETH0->MAC_CONFIGURATION |= ETH_MAC_CONFIGURATION_TE_Msk | ETH_MAC_CONFIGURATION_RE_Msk;
+  // Enable DMA transmission and reception
+  ETH0->OPERATION_MODE |= ETH_OPERATION_MODE_ST_Msk | ETH_OPERATION_MODE_SR_Msk;
+
+  NVIC_EnableIRQ(ETH0_0_IRQn);
```
scaprile, line 142: This does not belong in here. We do that in our HAL or the user does.

scaprile: That is CMSIS', we are (mostly) CMSIS-free inside Mongoose library.

src/drivers/xmc.c
```diff
@@ -0,0 +1,208 @@
...
+  ETH0->TRANSMIT_POLL_DEMAND = 0;
+  return len;
+}
+
+static bool mg_tcpip_driver_xmc_up(struct mg_tcpip_if *ifp) {
+  uint16_t bsr = eth_read_phy(MG_PHY_ADDR, MG_PHYREG_BSR);
+  bool up = bsr & MG_BIT(2) ? 1 : 0;
+  (void) ifp;
+  return up;
+}
```
scaprile, line 176: Same comment (phy.c)

scaprile (changes requested):
Regarding CMSIS stuff, wait for Sergey's opinion. We only have SAMD using that so far.
I've found that in many cases there are subtle naming differences, we wouldn't be able to support as many STM32F and H as we are (without mods...) if we were using CMSIS here.
So far we've been providing our own ETH struct and using MG_BIT with proper comments. That guarantees any other device with the same controller at the same address will work, regardless of how the header writers decided to name the bits and registers

Regarding PHY code, check it out, it is cool. Instant support for all boards with any one of those PHYs. STM and IMXRT drivers have already been ported. I left SAMD as is because there was something that smelled different in there... and TM4C has a built-in PHY.

mongoose.h
```diff
@@ -254,30 +254,30 @@ static inline int mg_mkdir(const char *path, mode_t mode) {
...
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+#include <time.h>
+
+#ifndef MG_IO_SIZE
+#define MG_IO_SIZE 1460
+#endif
```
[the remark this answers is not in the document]
robertc2000: That confused me as well. Github shows there is a diff, although there is none.. I'll check it out

src/drivers/xmc.c
```diff
@@ -0,0 +1,208 @@
...
+
+  // init phy
+  eth_write_phy(MG_PHY_ADDR, MG_PHYREG_BCR, 0x8000); // reset
+  uint8_t bcr = eth_read_phy(MG_PHY_ADDR, MG_PHYREG_BCR);
+  while ((bcr & 0x8000) != 0) {
+    bcr = eth_read_phy(MG_PHY_ADDR, MG_PHYREG_BCR);
+  }
+
+  // auto-negotiation, 100M, full-duplex
+  eth_write_phy(MG_PHY_ADDR, MG_PHYREG_BCR, 0x3100);
```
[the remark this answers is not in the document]
robertc2000: I tried this just before pushing code, but for some reason, something didn't work and the interface appeared to be down. It's a KSZ8081 phy, which by looking at the code, we seem to support it, but in my case it didn't work. The issue is either in the initialization part of the PHY or reading the status of the interface, I didn't investigate that far (but I plan to do it in the future), so I decided to go with our old way instead for now.

src/drivers/xmc.c
```diff
@@ -0,0 +1,208 @@
...
+
+#if MG_ENABLE_TCPIP && defined(MG_ENABLE_DRIVER_XMC) && MG_ENABLE_DRIVER_XMC
+
+#ifdef XMC4400_F100x512
+#include <XMC4400.h>
+#else
+#ifdef XMC4700_F144x2048
+#include <XMC4700.h>
+#endif
+#endif
```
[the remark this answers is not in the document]
robertc2000: I decided to include cmsis, because defining the eth structure will look very ugly in this case.. Compared to other drivers we have, it has a lot bigger and when I saw how many lines it wasted (almost double than stm or imxrt), I decided to rather include cmsis header.. Plus, it helps a lot with code readability if we use the bit masks defined in the cmsis header, rather than MG_BIT(bit)..

cpq (changes requested):
Please get rid of CMSIS things, like ETH_STATUS_RI_Msk, NVIC_*, etc.

src/drivers/xmc.c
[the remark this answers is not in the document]
scaprile: Yes, this is weird, you reset, wait for it to come back, and then set auto-negotiation and force specifics at the same time. If the PHY were hardware reset with a GPIO as in all other cases, the same init sequence should work, or something really weird is going on with your switch/hub.
Are you actually hw resetting the PHY ? We always assume that would be done, and we do that in our HALs
Are you connecting to a switch that performs auto-negotiation or do you have a (xover) cable to your notebook ?

src/drivers/xmc.c
```diff
@@ -0,0 +1,228 @@
...
+  ETH0->MAC_CONFIGURATION = MG_BIT(13) /* DO */ | MG_BIT(11) /* DM */ | 
+                            MG_BIT(14) /* FES */ | MG_BIT(24) /* TC */;
+
+  // set the MAC address
+  ETH0->MAC_ADDRESS0_HIGH = MG_U32(0, 0, ifp->mac[5], ifp->mac[4]);
+  ETH0->MAC_ADDRESS0_LOW = 
+        MG_U32(ifp->mac[3], ifp->mac[2], ifp->mac[1], ifp->mac[0]);
+
+  // Configure the receive filter
+  ETH0->MAC_FRAME_FILTER = MG_BIT(10) /* HPF */ | MG_BIT(2) /* HMC */;
```
scaprile, line 105: I. myself, IMHO, consider that bit functions can be easily get from the RM, so a simple `// HPF, HMC`-type comment (with references in the same order, left to right, as bits have been set, usually MSb to LSb) produces smaller comments, more readable code, and avoids split line.
That's just the way I do it, my opinion, my 2 cents.
We don't seem to have a strict coding style rule on this

src/drivers/xmc.h
```diff
@@ -0,0 +1,46 @@
...
+  //    -------------------------------------
+  //                                -1  <-- tell driver to guess the value
+  //    60-100 MHz    HCLK/42        0
+  //    100-150 MHz   HCLK/62        1
+  //    20-35 MHz     HCLK/16        2
+  //    35-60 MHz     HCLK/26        3
+  //    150-250 MHz   HCLK/102       4  <-- value for Nucleo-F* on max speed
+  //    250-300 MHz   HCLK/124       5
+  //    110, 111 Reserved
+  int mdc_cr;  // Valid values: -1, 0, 1, 2, 3, 4, 5
```
scaprile, line 18: "This ain't no Nucleo", please update that
Are valid values OK ?

src/drivers/xmc.h
```diff
@@ -0,0 +1,46 @@
...
+  uint8_t phy_addr;
+};
+
+#ifndef MG_TCPIP_PHY_ADDR
+#define MG_TCPIP_PHY_ADDR 0
+#endif
```
scaprile, line 24: Did you check this in the hw ? Some systems use '0' as a broadcast address and that would also explain the differences you see with the PHY initialization

scaprile (changes requested):
nit pick on xmc.h

src/drivers/xmc.h
robertc2000: Yeah, the values are valid (I took the table from Nucleo and only updated the values, forgot the rest..). It is fixed now.

scaprile (approved):

PR #2663, opened 2024-03-25
TLS client implementation

cpq:


test/unit_test.c
```diff
@@ -743,10 +709,11 @@ static int fetch(struct mg_mgr *mgr, char *buf, const char *url,
   if (c != NULL && mg_url_is_ssl(url)) {
     struct mg_tls_opts opts;
     memset(&opts, 0, sizeof(opts));  // read CA from packed_fs
+    opts.name = mg_url_host(url);
     opts.ca = mg_unpacked("/test/data/ca.pem");
-    if (strstr(url, "127.0.0.1") != NULL) {
+    if (strstr(url, "localhost") != NULL) {
       // Local connection, use self-signed certificates
-      opts.ca = mg_str(s_tls_ca);
+      opts.ca = mg_unpacked("/test/certs/ca.crt");
```
scaprile, line 716: Well, one of these was being read from code as mg_str to keep a test on that inside unit_test.c. Looks like there's none of those tests; I think we should have one.

test/unit_test.c
```diff
@@ -1207,14 +1174,14 @@ static void test_tls(void) {
...
+  const char *url = "https://localhost:12347";
   char buf[FETCH_BUF_SIZE];
   struct mg_tls_opts opts;
   struct mg_str data = mg_unpacked("/Makefile");
   memset(&opts, 0, sizeof(opts));
   // opts.ca = mg_str(s_tls_ca);
-  opts.cert = mg_str(s_tls_cert);
-  opts.key = mg_str(s_tls_key);
+  opts.cert = mg_unpacked("/test/certs/server.crt");
+  opts.key = mg_unpacked("/test/certs/server.key");
```
scaprile, line 1184: Same as above.

PR #2675, opened 2024-03-31
Implement firmware update in ESP32 device-dashboard example 

dvosully:
Implements OTA firmware update APIs for the ESP32 and integrates into the ESP32 device-dashboard example.
Uses the Espressif OTA APIs.

- Removed the SPIFFS partition and initialisation as it was not doing anything, and the flash space was required to fit the application into a 2MB (the default) ESP32 module.
- The CRC32 display is unimplemented as the ESP32 API provides nothing similar. ESP32 uses a SHA-256 for checksumming the firmware.
- The firmware flash date is parsed from the firmware binary. This datestamp is only regenerated when the firmware is built from clean. The date displayed in the web UI may differ from expected, as the datestamp in the firmware records local build time and the build timezone is not available.
- The number of wifi connect retries was increased from 3 to 10, as 3 would sometimes fail to connect.


scaprile:
Thank you, @dvosully , you have already contributed before and signed the CLA, we'll review your PR.

PR #2676, opened 2024-04-01
examples: fix some comments

stayweek:


scaprile:
Those are vendor comments. We won't modify vendor files

Issue #2689, opened 2024-04-05
QoS1 and QoS2 cannot be implemented with the API


RicoBiedermann:
**- My goal is:** 
Implement QoS 1 and QoS 2 for MQTT 3.1.1 and MQTT 5 based on the example: https://mongoose.ws/documentation/tutorials/mqtt-client/

**- My actions were:** 
I have readed the mqtt specs and some sites from mqtt broker companys. I have also readed the mongoose code.
https://www.emqx.com/en/blog/introduction-to-mqtt-qos (this site shows it realy good)
http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718099
https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901026
https://www.hivemq.com/blog/mqtt-essentials-part-4-mqtt-publish-subscribe-unsubscribe/
http://www.steves-internet-guide.com/understanding-mqtt-qos-levels-part-1/

**- My expectation was:** 
That I'm able to implement QoS 1 and QoS 2 with the tutorial.

**- The result I saw:**
I think it is not possible to archiv this with the current mongoose mqtt api.
To archiv this, we need a way to do a "retransmition" where I can set the packet-id and the duplication flag. 
If I sent a message with QoS1 and I dont get a MQTT_PUBACK for that message, I need to send it again.  But i need to send it with the same packet id and the duplication flag needs to be true. A Broker or a other client has no other way of recognising that it is a message that has already been sent, or that several messages are the same. 
If the message has a new packet id, it is a new message. 
If I do a subscription, I also need this information when I receive a message to determine if it is a new message or not.
On QoS2 i need the option to send a MQTT_CMD_PUBREL.

**- My question is:** 
Is there a way that I don't know yet?
Could it be a way to expand the mg_mqtt_message struct with a field for the retainflag, duplicationflag and to add the user porperties like the mg_mqtt_opts struct? 
The mg_mqtt_message struct could be used for mg_mqtt_pub() and mg_mqtt_sub() and with some changes in this functions it could work. 

# Environment

- mongoose version: 7.13
- Target hardware/board: imx8mp
- Target RTOS/OS (if applicable): Debian based linux

RicoBiedermann:
Why do you have closed that issue? 

I have checked the documentation, examples and the tutorials. And also I have checked now the unit_test. 
[...]

PR #2691, opened 2024-04-05
add simple API support for app MQTT pub retries

scaprile:
#2689 
- Should work as long as the id counter does not rollback (65536 messages sent while waiting to retransmit)
- Does not break current API

### DIFF
--- a/test/unit_test.c
+++ b/test/unit_test.c
@@ -504,10 +504,10 @@ static void test_mqtt_basic(void) {
 
   // Publish with QoS1 to subscribed topic and check reception
   // keep former opts.topic
-  opts.message = mg_str("hi1"), opts.qos = 1, opts.retain = false;
+  opts.message = mg_str("hi1"), opts.qos = 1, opts.retain = false, opts.retransmit_id = 0;
   retries = 0;  // don't do retries for test speed
   do {          // retry on failure after an expected timeout
-    mg_mqtt_pub(c, &opts);
+    opts.retransmit_id = mg_mqtt_pub(c, &opts);  // save id for possible resend
     for (i = 0; i < 500 && test_data.flags == 0; i++) mg_mgr_poll(&mgr, 10);
   } while (test_data.flags == 0 && retries--);
   ASSERT(test_data.flags == flags_published);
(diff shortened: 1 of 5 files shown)


test/unit_test.c
```diff
@@ -589,12 +593,14 @@ static void test_mqtt_ver(uint8_t mqtt_version) {
   retries = 0;  // don't do retries for test speed
   do {          // retry on failure after an expected timeout
     mg_mqtt_pub(c, &opts);
+    opts.id = c->mgr->mqtt_id;  // save id for possible resend
```
cpq, line 596: why do we need to keep it in opts?
can't we just do this?
```c
uint16_t id = c->mgr->mqtt_id;
```

scaprile: mg_mqtt_pub() needs a way to know whether to:
- send a new id
- send a specific id and set the DUP flag
I modified it to make the decision on the value of opts.id. If it is 0, it will create a new id. If it is !=0, it will use that id and set the DUP flag. The caller is responsible for keeping and clearing the id (and the message text for retransmission)

cpq: so that opts.id is used as a signal to set the flags, I see..

scaprile: The retransmission operation requires using the very same id of the former message, and, setting the DUP flag.
We need a way to do a retransmission instead of plain publish.

Using a new function was my first choice, that was discarded.

We can break the mg_mqtt_pub() signature by adding an extra argument, or two arguments.
- 1 argument: address of id holder, pub stores the id there and there's no need to hold it in options nor to pick it up from the manager. Hidden logic: we set the DUP flag for the user. Seems obvious to me, as one implies the other. Cons: the user has to provide a clean id holder or a NULL pointer if don't care (to simplify usage in usual conditions).
- 2 arguments: same as above, but the user sets the DUP flag. Cons: the user has to actually do know about this even if they just want to send QoS0 or don't care about retransmissions. Seems more complicated to me, for the same reason. It is quite prone to user errors

For the reasons above, I though hiding that "lovely" id in options was simpler and didn't break the current API
If we are going to break the API, then 
```
QoS0+ with no interest in retransmissions:
mg_mqtt_pub(..., NULL);
QoS1+ with retransmission capabilities:
uint16_t id=0;
mg_mqtt_pub(..., &id);  // regular send, fills 'id'
mg_mqtt_pub(..., &id);  // retransmission, sets DUP flag and uses 'id'
```
`id = 0` is invalid in the protocol, so it is a valid sign that we are sending a brand new message.

scaprile: BTW, our customer opening the issue likes the hidden logic https://github.com/cesanta/mongoose/issues/2689#issuecomment-2049611160

PR #2752, opened 2024-05-14
add mg_str_to_num

scaprile:


src/http.c
```diff
@@ -24,21 +24,11 @@ static int mg_ncasecmp(const char *s1, const char *s2, size_t len) {
...
-    result *= 10;
-    if (result > max - digit) return false;  // Overflow
-    result += digit;
-    i++, ndigits++;
-  }
-  while (i < str.len && (str.buf[i] == ' ' || str.buf[i] == '\t')) i++;
-  if (ndigits == 0) return false;  // #2322: Content-Length = 1 * DIGIT
-  if (i != str.len) return false;  // Ditto
-  *val = (size_t) result;
+  if (i != str.len) return false;  // #2322: Content-Length = 1 * DIGIT
```
scaprile, line 31: here

src/str.h
```diff
@@ -15,11 +15,13 @@ struct mg_str mg_str(const char *s);
...
 int mg_strcmp(const struct mg_str str1, const struct mg_str str2);
-int mg_strcasecmp(const struct mg_str str1, const struct mg_str str2);  // this one is new
+int mg_strcasecmp(const struct mg_str str1, const struct mg_str str2);
 bool mg_match(struct mg_str str, struct mg_str pattern, struct mg_str *caps);
 bool mg_span(struct mg_str s, struct mg_str *a, struct mg_str *b, char delim);
 
 void mg_unhex(const char *buf, size_t len, unsigned char *to);
 unsigned long mg_unhexn(const char *s, size_t len);
 
 uint8_t mg_toi(char c, int base); // base: 16, 10
```
scaprile, line 25: These will go

mongoose.c
```diff
@@ -8120,6 +8110,80 @@ bool mg_span(struct mg_str s, struct mg_str *a, struct mg_str *b, char sep) {
...
+  uint64_t max = val_len == 1   ? 0xFF
+                 : val_len == 2 ? 0xFFFF
+                 : val_len == 4 ? 0xFFFFFFFF
+                                : (uint64_t) ~0;
+  uint64_t max2 = max / 10, result = 0;
+  while (i < str.len && (str.buf[i] == ' ' || str.buf[i] == '\t')) i++;
+  if (base == 0 && str.len >= 2) {
+    if (str.buf[i] == '0') {
+      i++;
+      base = str.buf[i] == 'b' ? 2 : str.buf[i] == 'x' ? 16 : 10;
```
cpq, line 8125: strtol logic assumes numbers that start with `0` but followed by digits to be octal.
Shall we employ the same logic?
i.e.

```c
 base = str.buf[i] == 'b' ? 2 : str.buf[i] == 'x' ? 16 : 8;
```

src/str.c
```diff
@@ -96,6 +96,80 @@ bool mg_span(struct mg_str s, struct mg_str *a, struct mg_str *b, char sep) {
...
+  }
+  if (ndigits == 0) return 0;  // #2322: Content-Length = 1 * DIGIT
+  if (val_len == 1) {
+    *((uint8_t *) val) = (uint8_t) result;
+  } else if (val_len == 2) {
+    *((uint16_t *) val) = (uint16_t) result;
+  } else if (val_len == 4) {
+    *((uint32_t *) val) = (uint32_t) result;
+  } else {
+    *((uint64_t *) val) = (uint64_t) result;
```
cpq, line 164: I suggest to explicitly check for a valid val_len:

```c
else if (val_len == 8) {
} else {
 return false
}
```

scaprile: Addressed at top of function to avoid lots of work in vain

src/str.c
```diff
@@ -96,6 +96,77 @@ bool mg_span(struct mg_str s, struct mg_str *a, struct mg_str *b, char sep) {
...
+      break;
+    case 16:
+      while (i < str.len) {
+        char c = str.buf[i];
+        uint64_t digit = (c >= '0' && c <= '9')   ? (uint64_t) (c - '0')
+                         : (c >= 'A' && c <= 'F') ? (uint64_t) (c - '7')
+                         : (c >= 'a' && c <= 'f') ? (uint64_t) (c - 'W')
+                                                  : (uint64_t) ~0;
+        if (digit == (uint64_t) ~0) break;
+        if (result > max2) return false;  // Overflow
```
cpq, line 145: should max2 here be max / 16 ?

robertc2000: I think so too. Also, for the base 2 case, it should be max / 2.  It's because, when we add a digit to a number, we multiply it by the base.

scaprile: Yes, thought of making it local then forgot about it...
Fixed, some more tests added.
"base 0" missing, will add it later

cpq (approved):

cpq (approved):

PR #2758, opened 2024-05-20
add support for getting the DNS server from DHCP

timaydin:
Set the option that requests the DNS server from DHCP during the DHCP Request phase. Currently, the DNS server is saved the same way that the ifp->ip, ipf->gw etc addresses are saved. No change was made to have this DNS server be used during DNS resolve. As a result, it is necessary for the application to manually assing the DNS information into mgr->dns4 or mgr->dns6.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -5824,12 +5824,16 @@ static void mg_tcpip_poll(struct mg_tcpip_if *ifp, uint64_t now) {
       mg_tcpip_rx(ifp, ifp->recv_queue.buf, len);
     }
   } else {  // Interrupt-based driver. Fills recv queue itself
-    char *buf;
-    size_t len = mg_queue_next(&ifp->recv_queue, &buf);
-    if (len > 0) {
-      mg_tcpip_rx(ifp, buf, len);
-      mg_queue_del(&ifp->recv_queue, len);
-    }
+    size_t len;
+    do
+    {
+        char *buf;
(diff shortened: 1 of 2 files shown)


scaprile:
Please read this, specially, the second bullet.
https://github.com/cesanta/mongoose#contributions
The first bullet can wait until we decide to accept what you propose.

This is not the proper way to fix a probable issue, 
If there is an actual issue, please open an issue, _honor the issue template_ so we can know your scenario and reproduce the issue if necessary, and let us know _exactly_ what you are doing.
If you have a fix, as you seem to have, and you want to propose it, you are welcome. We may or not take it as it is but we will surely fix the issue if it is in our code.

cpq:
@timaydin thanks! This is a good change. Though it cannot be integrated as - is, because it changes an amalgamated mongoose.c/h, not the source. If you fix your PR and update the sources as well, and sign the CLA https://cesanta.com/cla.html - we'll integrate. 

Note that adding `ifp->dns` has no effect. THere is no point in adding it. What should be done instead, is when we receive DHCP response with DNS server in it, we should construct `mgr->dns4.url` to be  `udp://x.x.x.x:53`. 

The freertos-tcp change should be separate. @scaprile let's update the freertos-tcp version we pull in examples/stm32/nucleo-f746zg-make-freertos-tcp/Makefile to the latest, V4.1.0, and integrate the suggested change if it is required.

timaydin:
close

PR #2794, opened 2024-06-17
handle `Connection: close`

### LINKED ISSUE #2765: Mongoose ignores `Connection: close` headers
To see this for yourself,

0. Compile and run the `http-restful-server` tutorial.

1. Send it a pipeline consisting of two `GET` requests in which the first request has a `Connection: close` header:
```
printf 'GET / HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\nGET / HTTP/1.1\r\nHost: a\r\n\r\n' | nc localhost 8000
```

2. Observe that the server responds twice, even though the second request should be ignored because it was sent after the connection was closed.

# Environment

- mongoose version: master (6bb40e6ec96f95bfd36a816b430ea2726fac9d05)
- Compiler: GCC 13.2.0
- Platform: AMD64 Linux

scaprile:
fix #2765 

This makes Mongoose honor `Connection: close` headers. Had to "flush" the receive buffer because `c->is_draining` is not enough, the parser calls `http_cb()` again as it sees more data, no matter the value of `c->is_draining`

To add unit_tests, there is another issue to be fixed first. `mg_http_serve_file()` closes the connection when it finishes. So, directory listings showed the behavior described by the OP, while file requests did not.
With this PR, `Connection: close` is **always** handled, BUT, as long as one of the pipelined requests is for a file, our server will drop the connection after serving that file, no matter how many requests are left, and whether or not `Connection: close` was set at or after the file request.


mongoose.c
```diff
@@ -3297,8 +3297,16 @@ static void http_cb(struct mg_connection *c, int ev, void *ev_data) {
...
+      if (c->is_accepted) {
+        struct mg_str *cc;
+        c->is_resp = 1;  // Start generating response
+        if ((cc = mg_http_get_header(&hm, "Connection")) != NULL &&
+            mg_strcasecmp(*cc, mg_str("close")) == 0) {
+          c->is_draining = 1;  // honor "Connection: close"
+          c->recv.len = 0;     // discard remaining requests, if any
+        }
+      }
+      mg_call(c, MG_EV_HTTP_MSG, &hm);  // User handler can clear is_resp
```
cpq, line 3309: This code changes the logic. Before, we were mg_call-ing always, for both client and server connections. Now, we ignore client connections. I guess we should keep the logic intact, by moving the mg_call outside of the `if` block.

I also would not clear `c->recv.len`, especially before calling the handler, as handler might use it.

A pedantic nitpick on the code - I suggest a slight refactoring:

```c
      if (c->is_accepted) {
        struct mg_str *cc = mg_http_get_header(&hm, "Connection");
        c->is_resp = 1;  // Start generating response
        if (cc != NULL && mg_strcasecmp(*cc, mg_str("close")) == 0) {
          c->is_draining = 1;  // honor "Connection: close"
        }
      }
```

scaprile: Nope, `mg_call()` was outside of the `if`. clang-format changed the comment spacing, so that line showed as modified.

Just setting c->is_draining is not enough, the user handler reply clears c->is_resp so the loop above continues and parses the next message:
https://github.com/cesanta/mongoose/blob/793171e9e0a36a5343313b53b98cf481ae2bf17f/src/http.c#L1006
Nevertheless, I see your point with clearing the buffer, got carried away by thinking pipelined requests are GETs and the message had already been parsed.
Refactored as suggested and placed it AFTER `mg_call`. Replaced buffer purge with a `break`, actually exiting the loop.

cpq: Looks good! Approved.

Unit test please.

cpq (approved):

mongoose.c
scaprile: > Unit test please

@cpq pls read comment on top, there's an issue on that. I could write a more complex test not using files but would need to add that one later.

PR #2807, opened 2024-06-27
Fix raw TLS overflow for large uploads

cpq:


src/sock.c
```diff
@@ -276,17 +276,21 @@ static void read_conn(struct mg_connection *c) {
...
-      if (n == MG_IO_ERR && c->rtls.len == 0) {
-        // Close only if we have fully drained both raw (rtls) and TLS buffers
-        c->is_closing = 1;
-      } else {
-        if (n > 0) c->rtls.len += (size_t) n;
-        if (c->is_tls_hs) mg_tls_handshake(c);
-        n = c->is_tls_hs ? (long) MG_IO_WAIT : mg_tls_recv(c, buf, len);
+      // Do not read to the raw TLS buffer if it already has enough.
+      // This is to prevent overflowing c->rtls if our reads are slow
+      if (c->rtls.len < 4096) {
```
scaprile, line 281: Where did this magic number come from ?
We have `MG_IO_SIZE` ruling that buffer resizing.
In fact, I addressed a couple of Unix issues/questions by contributing to our _best practices_ section:
> in scenarios where data accumulates faster than it can be consumed, as for example, in TLS large file transfers, you need to make MG_IO_SIZE larger to avoid reaching the MG_MAX_RECV_SIZE limit while buffering.

With this patch, we don't fill rtls, then data will accumulate in recv... but the user can throttle there catching EV_READ, right ?

So, should this be written as a function of `MG_IO_SIZE` or is it a fixed value related to TLS block size or something ?

scaprile (changes requested):
```
$ make
$ ./example &
$ curl --insecure -H Expect: https://localhost:8443/upload/a.txt --data-binary @largeenoughfile
185caf22 3 mongoose.c:7688:accept_conn  3 6 accepted 127.0.0.1:39868 -> 0.0.0.0:8443
185caf44 3 mongoose.c:7543:read_conn    3 6 0:0:0 -2 err 0
185caf45 3 mongoose.c:7543:read_conn    3 6 0:0:0 -2 err 0
185caf45 3 mongoose.c:7543:read_conn    3 6 0:0:1860 166 err 0
185caf45 3 mongoose.c:7543:read_conn    3 6 0:0:2048 -2 err 0
185caf45 3 mongoose.c:7543:read_conn    3 6 0:0:4096 -2 err 0
185caf45 3 mongoose.c:7543:read_conn    3 6 0:0:6144 -2 err 0
185caf45 3 mongoose.c:7543:read_conn    3 6 0:0:8192 -2 err 0
185caf45 3 mongoose.c:7543:read_conn    3 6 0:0:10240 -2 err 0
185caf46 3 mongoose.c:7543:read_conn    3 6 0:0:12288 -2 err 0
185caf46 3 mongoose.c:7543:read_conn    3 6 0:0:14336 -2 err 0
185caf46 3 mongoose.c:7543:read_conn    3 6 0:0:16384 -2 err 0
185caf46 3 mongoose.c:7543:read_conn    3 6 0:0:16384 -2 err 0
185caf46 3 mongoose.c:7543:read_conn    3 6 0:0:16384 -2 err 0
185caf46 3 mongoose.c:7543:read_conn    3 6 0:0:16384 -2 err 0
185caf46 3 mongoose.c:7543:read_conn    3 6 0:0:16384 -2 err 0
185caf46 3 mongoose.c:7543:read_conn    3 6 0:0:16384 -2 err 0
forever
```

src/sock.c
```diff
@@ -276,17 +276,21 @@ static void read_conn(struct mg_connection *c) {
...
+      if (c->rtls.len < 16 * 1024) {  // 16k is the max TLS message size
+        if (!ioalloc(c, &c->rtls)) return;
+        n = recv_raw(c, (char *) &c->rtls.buf[c->rtls.len],
+                     c->rtls.size - c->rtls.len);
+        if (n == MG_IO_ERR && c->rtls.len == 0) {
+          // Close only if we have fully drained both raw (rtls) and TLS buffers
+          c->is_closing = 1;
+        } else {
+          if (n > 0) c->rtls.len += (size_t) n;
+          if (c->is_tls_hs) mg_tls_handshake(c);
```
scaprile, line 290: no TLS HS if buffer is full (shouldn't ever happen...)

scaprile:
> the record protocol does introduce a few important implications that we need to be aware of:
>   - Maximum TLS record size is 16 KB
>   - Each record contains a 5-byte header, a MAC (up to 20 bytes for SSLv3, TLS 1.0, TLS 1.1, and up to 32 bytes for TLS 1.2), and padding if a block cipher is used.
>  - To decrypt and verify the record, the entire record must be available. 

So the magic number _today_ has to be 16384 + 5 + 32 + possible padding

scaprile (approved):

PR #2816, opened 2024-07-03
fix read after free

scaprile:
When receiving more than 1 TLS record, that is content > 16K, data is buffered in rtls until a full record is received. This buffer grows at MG_IO_SIZE units when needed.
At this point, our built-in TLS marks a record has been found and holds a pointer to rtls stored in tls->recv. It decrypts and transfers to recv MG_IO_SIZE units.
As more data arrives, rtls is resized again, been freed and re-alloc'ed. But our built-in TLS still has a pointer to the freed content.
This works as long as the content stays there, but eventually triggers some complex failure mechanism.

This PR adds a remap function so our built-in TLS is able to remap its buffer pointers when the actual buffer changes, keeping opaque sections that way.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -9925,7 +9925,10 @@ static int mg_tls_recv_record(struct mg_connection *c) {
 
 void mg_tls_ioremap(struct mg_connection *c, size_t offset) {
   struct tls_data *tls = (struct tls_data *) c->tls;
-  if (tls->recv.buf != NULL) tls->recv.buf += offset;
+  if (tls->recv.buf != NULL) {
+    size_t newptr = ((size_t) tls->recv.buf) + offset;
+    tls->recv.buf = (unsigned char *) newptr;
+  }
 }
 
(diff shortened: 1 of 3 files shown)


scaprile:
@cpq 
The pointer difference is captured on a size_t. Adding a size_t to a pointer preserves the difference due to the rules of two's complement (same discussion as for timers, long ago). A negative difference (new buffer at a lower address) is arithmetically handled correctly, with an addition overflow.
However, either this situation happens only on MacOS or there is something particular there, as the sanitizer does not like the overflow:
```
src/tls_builtin.c:476:44: runtime error: addition of unsigned offset to 0x00010940008f overflowed to 0x000106a44d0f
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/tls_builtin.c:476:44 in 
```
As you see, in spite of the overflow, the address is correct. However, the overflow seems not to be a good practice here.
How do you suggest to work this out, convert everything to size_t before the addition and then storing the pointer ?

PR #2843, opened 2024-07-23
Refactor tls receive pointer scheme (fix read after free)

scaprile:
When receiving more than 1 TLS record, that is content > 16K, data is buffered in rtls until a full record is received. This buffer grows at MG_IO_SIZE units when needed.
At this point, our built-in TLS marks a record has been found and holds a pointer to rtls stored in tls->recv. It decrypts and transfers to recv MG_IO_SIZE units.
As more data arrives, rtls is resized again, been freed and re-alloc'ed. But our built-in TLS still has a pointer to the freed content.
This works as long as the content stays there, but eventually triggers some complex failure mechanism.

This PR rewrites the TLS code to store and work with an offset instead. There is no performance penalty, some RAM saved, and there are no dependencies on buffer management.

NOTE: This has been opaqued by #2807 that limits the amount of buffer usage for rtls. Even though testing will pass now, reverting 2807 in the future for whatever reason will expose this again.
NOTE: #2817 fixed another bug that manifested after 2807 shadowed this one.

closes #2816

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -9504,9 +9504,9 @@ struct tls_data {
   enum mg_tls_hs_state state;  // keep track of connection handshake progress
 
   struct mg_iobuf send;  // For the receive path, we're reusing c->rtls
-  struct mg_iobuf recv;  // While c->rtls contains full records, recv reuses
-                         // the same underlying buffer but points at individual
-                         // decrypted messages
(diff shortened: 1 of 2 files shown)


scaprile:
Rebased after cleanup changes

PR #2844, opened 2024-07-24
FreeRTOS requires a valid proto arg to socket()

quartzjer:
Simple fix.

### DIFF
--- a/src/sock.c
+++ b/src/sock.c
@@ -359,8 +359,9 @@ static void setsockopts(struct mg_connection *c) {
 
 void mg_connect_resolved(struct mg_connection *c) {
   int type = c->is_udp ? SOCK_DGRAM : SOCK_STREAM;
+  int proto = type == SOCK_DGRAM ? IPPROTO_UDP : IPPROTO_TCP;
   int rc, af = c->rem.is_ip6 ? AF_INET6 : AF_INET;  // c->rem has resolved IP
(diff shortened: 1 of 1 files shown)


scaprile:
Can you please explain why and where is this needed ? We build every day for FreeRTOS and don't need anything extra.

quartzjer:
It builds fine without this, the problem we encountered is when you pass `0` into FreeRTOS's `socket()` wrapper, at runtime it returns an error and fails.

It hits this assert with a zero value passed in: https://github.com/particle-iot/freertos/blob/1f3aae61b26354cbf38e87b36994e2a4b78d7593/FreeRTOS-Plus/Source/FreeRTOS-Plus-TCP/FreeRTOS_Sockets.c#L255

cpq:
Ah. It's FreeRTOS+TCP , being BSD socket compatible but not su much. This change looks good to me
thank you @quartzjer

scaprile:
@cpq I plan to test on Monday and also change other two calls to socket(), so the OP wouldn't need to sign the CLA. WDYT ?

quartzjer:
Sorry I didn't get the CLA signed in time, but very glad to hear you got the fix incorporated regardless, thanks!

scaprile:
We didn't actually did exactly what you submitted, so no CLA was necessary as we didn't include your code.

Issue #2719, opened 2024-04-23
Fix for #2619 breaks fix for #2603


jcorporation:
Sorry, that I must open a new issue regarding above issues.

- My goal is: Read post requests from a https socket issued by javascript
- My actions were: `MG_EV_HTTP_MSG with struct mg_http_message *hm = (struct mg_http_message *) ev_data`; as used in your examples
- My expectation was: mongoose reads the complete post data
- The result I saw: Reading large post data works with latest master, but reading small post data is failing. After reverting commit 3ac9bd0468969ae2d40bc84379c8038de6578355 reading small post data works, but reading large post data is failing. I tested this with your http-restful-server example. The issue appears only with enabled TLS (OpenSSL) and only with iPhone as a client, with my laptop browser there are no problems.
- My question is: No question, but it seems the fix for #2619 breaks fix for #2603.

# Environment

It is the same test environment as in #2619 and 2603.

- mongoose version: latest master
- Compiler/IDE and SDK: `gcc version 13.2.0 (Ubuntu 13.2.0-4ubuntu3)`
- Target hardware/board: `Linux t14 6.5.0-15-generic #15-Ubuntu SMP PREEMPT_DYNAMIC Tue Jan  9 17:03:36 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`
- Client: iPhone (tested with safari and firefox mobile)

Attached a log with two post request. First is small request that is failing. Second is a bigger request that works.

[log.txt](https://github.com/cesanta/mongoose/files/15074305/log.txt)

scaprile:
@jcorporation Jürgen, can you please check #2720 ? (In both scenarios) Thank you !

@cpq : I think we've been going back and forth with two corner cases. Using "blame" I can see we've been using `mg_tls_pending()`, then resorted to `c->rtls.len` and then back to `mg_tls_pending()`.
I think there are two corner cases here:

- Msg is very short and gets sent along with the headers. `mg_tls_pending()` returns 0 as it has no data, BUT `c->rtls.len > 0` as there is outstanding data after parsing headers. That is #2603 , we fixed it but broke the next scenario
- Msg is of such a length that a whole block, or a group of them, fit the transfer. `c->rtls.len` is zero, BUT `mg_tls_pending()` will indicate it has a block to process. That seems to be #2619, its fix goes back to how things were before, causing a regression on the former scenario.

[...]

PR #2846, opened 2024-07-26
Move http-restful-server to builtin TLS, plus minor nits

cpq:


src/sock.c
```diff
@@ -282,7 +282,7 @@ static void read_conn(struct mg_connection *c) {
         if (!ioalloc(c, &c->rtls)) return;
         n = recv_raw(c, (char *) &c->rtls.buf[c->rtls.len],
                      c->rtls.size - c->rtls.len);
-        if (n == MG_IO_ERR && c->rtls.len == 0) {
```
scaprile, line 285: We've been going back and forth with this, breaking it several times.
Is there an actual reason to do this ? We've already been there and it didn't work, I think because there can be outstanding data in the buffer and TLS processed a whole record and hence not pending, so we never deliver the last chunk.

In reverse chronological order:
Screenshot from 2024-07-26 16-40-59
Screenshot from 2024-07-26 16-41-45
Screenshot from 2024-07-26 16-42-35
Screenshot from 2024-07-26 16-43-20
Screenshot from 2024-07-26 16-44-39

cpq: Yes there was a reason for that.
I started a HTTPS server on port 8443
And ran http client against it.
rtls buffered plain text HTTP request, thinking it is a TLS record, and thinking is was incomplete, so a connection was waiting for more.
I stopped HTTP client , but on the server side, a connection was still lingering in the infinite "read loop" .

scaprile: Yeah, I imagined something like that could happen... Then this should be the code, but for some reason we changed it, as we already changed the very one you just merged.
image
I'll revisit those issues on Monday and we can chat having data to support our brainstorming. WDYT ?

scaprile: Looks like that is a new corner case.
Your code will break some stuff.
We've been here
https://github.com/cesanta/mongoose/issues/2603#issuecomment-1938244497
https://github.com/cesanta/mongoose/issues/2719#issuecomment-2072530370
https://github.com/cesanta/mongoose/pull/2720
We should fix this without breaking those cases again.
I would revert that line back and probably re-think the whole scenario. As it is now, we have a regression and we'll probably have jcorporation posting an issue soon.

scaprile: @cpq

scaprile: IIUC, the problem is that non-TLS data in the buffer prevents proper closure... Is it right ?
Then, once we get IO_ERR there's nothing we can do after we deliver that data to TLS; so, what if we see IO_ERR, see data in rtls, remember that condition, keep going, deliver to TLS. Next time, we see we've been there and then close. WDYT ?
Something like that won't break those cases mentioned above.
After #2720, mg_tls_pending() or data in rtls causes the connection to be pending, that guarantees we'll arrive here, so we just need to remember we've already seen IO_ERR so if there is still outstanding data, whatever was there couldn't be processed by TLS. (Could this scenario also happen if a connection loss happens in the middle of a TLS block ? In such a case, asking for tls_pending won't help, but a fix like the one I'm proposing will work.)

cpq: Sergio, your reasoning seems correct to - please go ahead and push a fix

PR #2862, opened 2024-08-14
Disable POSIX FS by default

cpq:


scaprile (approved):
Well, tests pass.
Maybe Keil complains, I'll test later.
Maybe we could also enable in Zephyr.

Surely we'll need a large cleanup to get rid of all unnecessary mkdir definitions everywhere. I'll take care of that, slowly.

PR #2880, opened 2024-08-31
Make mg_random() fallible. Abort TLS on weak RNG

cpq:


tutorials/http/http-server/Makefile
```diff
@@ -13,7 +13,7 @@ CFLAGS_MONGOOSE += -DMG_TLS=MG_TLS_BUILTIN
 ifeq ($(OS),Windows_NT)   # Windows settings. Assume MinGW compiler. To use VC: make CC=cl CFLAGS=/MD OUT=/Feprog.exe
   PROG ?= example.exe           # Use .exe suffix for the binary
   CC = gcc                      # Use MinGW gcc compiler
-  CFLAGS += -lws2_32            # Link against Winsock library
+  CFLAGS += -lws2_32 -lbcrypt   # Link against Winsock library
```
scaprile, line 16: and bcrypt is ??

tutorials/http/http-server/Makefile
```diff
@@ -3,7 +3,7 @@ DELETE = rm -rf                   # Command to remove files
 OUT ?= -o $(PROG)                 # Compiler argument for output file
 SOURCES = main.c mongoose.c       # Source code files
 CFLAGS = -W -Wall -Wextra -g -I.  # Build options
-CFLAGS += -fsanitize=address,undefined,alignment
+#CFLAGS += -fsanitize=address,undefined,alignment
```
scaprile, line 6: yesterday added, today removed. Why removing it ? Windoze ?

src/tls_builtin.c
```diff
@@ -623,7 +623,7 @@ static void mg_tls_server_send_hello(struct mg_connection *c) {
   // calculate keyshare
   uint8_t x25519_pub[X25519_BYTES];
   uint8_t x25519_prv[X25519_BYTES];
-  mg_random(x25519_prv, sizeof(x25519_prv));
+  if (!mg_random(x25519_prv, sizeof(x25519_prv))) mg_error(c, "RNG");
```
scaprile, line 626: Can't we just fail in mg_tls_init() instead of failing in three different places ? We can ignore result here and try a bogus random number on init; if it fails, we error (disconnect).

tutorials/http/http-server/Makefile
cpq: Yeah, windows build.
Added for debugging

src/tls_builtin.c
cpq: Why should we ignore RNG failure after TLS init ?

scaprile: To avoid having the same check in 3 different places, and waiting for the Hello to fail; when we can just elegantly fail on connection establishment, which happens before any exchange, when the user calls mg_tls_init().
So, instead of forcing the TLS handshake to start, and then (in 3 different places) fail because we don't have a good RNG (which to me is read as a lack of prevision, no look ahead), we check that before attempting anything, we know we won't be able to fulfill the request, hence we fail right from the start.

scaprile: If we failed on TLS init, no TLS init has taken place, hence those calls can't happen, connection has been dropped off.

scaprile: To avoid having the same check in 3 different places, and waiting for the Hello to fail; when we can just elegantly fail on connection establishment, which happens before any exchange, when the user calls mg_tls_init().
So, instead of forcing the TLS handshake to start, and then (in 3 different places) fail because we don't have a good RNG (which to me is read as a lack of prevision, no look ahead), we check that before attempting anything, we know we won't be able to fulfill the request, hence we fail right from the start.
If init fails, the other calls don't have a chance to be made. Or do they ?

cpq: My understanding is that mg_random() can succeed at one invocation, and fail in the other. That could be true for UNIX for sure - mg_random opens and reads /dev/urandom, which can succeed or fail depending on the state of the process. Thus I think we MUST check every single mg_random() invocation.

scaprile: I was convinced that the "weak" condition could be determined in one call.
So does this mean that some connections will work and other wills just fail ? That is a poor user experience. Is there any way to detect / remember the PRNG is a weak one and fail TLS attempts ?

src/drivers/phy.c
```diff
@@ -71,9 +65,10 @@ void mg_phy_init(struct mg_phy *phy, uint8_t phy_addr, uint8_t config) {
       phy->write_reg(phy_addr, MG_PHY_DP83x_REG_RCSR, MG_BIT(7) | MG_BIT(0));
     } else if (id1 == MG_PHY_KSZ8x) {
       // Disable isolation (override hw, it doesn't make sense at this point)
-      phy->write_reg(  // #2848, some NXP boards set ISO, even though
-          phy_addr, MG_PHY_REG_BCR,  // docs say they don't
-          phy->read_reg(phy_addr, MG_PHY_REG_BCR) & (uint16_t) ~MG_BIT(10));
+      // #2848, some NXP boards set ISO, even though docs say they don't
+      uint16_t bcr = phy->read_reg(phy_addr, MG_PHY_REG_BCR);
+      bcr &= ~(((uint16_t) 1U) << 10U);
+      phy->write_reg(phy_addr, MG_PHY_REG_BCR, bcr);
```
scaprile, line 71: This change is breaking tests.
What is the reason for it ? Looks like it is just style. Inverting bits of a uint16 makes it an unsigned int, that is why it was casted that way, after silencing all compilers at that time, when doing that change.
Why 3 sentences instead of a clear 1 is better ? I think the read-modify-right operation is clear with one sentence; anyway, at least keep the way the bit is cleared, I do think it is easier to read, and it definitely does not trigger test failures

It is not related to the RNG change, anyway

scaprile (changes requested):
Please address unrelated PHY driver changes causing tests to fail. Comment explains

src/tls_builtin.c
cpq: Yeah, TLS connection might unexpectedly fails, if RNG degrades. That's what the security researcher suggests. I guess we should follow the advice.

src/drivers/phy.c
cpq: The reason is MSVC 98 (our C89 compiler) giving warning about the current code about conversion size mismatch. Yeah, not related to RNG

src/tls_builtin.c
scaprile: Oh, I didn't know about degradation. Cool.

src/drivers/phy.c
scaprile: Jeeze... please see chat

scaprile (approved):

PR #2881, opened 2024-09-02
Fix #2871 - accept `*` in mg_http_serve_opts::mime_types

cpq:


robertc2000 (approved):

Keith-Cancel:
I saw this in reference to my pull request. I will say though looking at the code when the wild card item gets used depends on the order of the list. If it's the first item it will always get used, if it's the last item it will get used if there are no other matches. Somewhere in the middle depends will give potential confusing behavior on when it gets used. Since for some files it may get used or may not ect...

Compared to what I written for https://github.com/cesanta/mongoose/pull/2871
https://github.com/cesanta/mongoose/blob/4ee43133aea4d31bb77877a5df2393ce719b450c/src/http.c#L534-L552

I first check that see if any extensions in the list match before using the wild card. I think that is a lot more consistent.

cpq:
The logic in this PR is exactly as it meant.

Keith-Cancel:
> The logic in this PR is exactly as it meant.

Why would you want the it depend on order though?

PR #2890, opened 2024-09-09
Fix multiple TLS records in buffer

### LINKED ISSUE #2668: When using Mbed TLS, not all data will be processed.
- My goal is: Use MbedTLS
- My actions were: Connect using TLS, process received data
- My expectation was: I would receive data in a timely manner
- The result I saw: If the server sends two messages in rapid succession, only the first message is processed in a timely manner, the other is only processed after a significant delay
- My question is: How can mongoose be modified to read all data?

# Environment

- mongoose version: 7.13
- Compiler/IDE and SDK: GCC 13.2 (arm-none-eabi-gcc), Raspberry Pi Pico SDK 1.5.1, Mbed TLS 2.28.1 (this is the version bundled with the Pi Pico SDK)
- Target hardware/board: RP2040
- Connectivity chip/module: Wiznet w5500, Mongoose built-in TCP
- Target RTOS/OS (if applicable): None

If more than one TLS record is in the same TCP packet Mongoose will process only the first record, leaving the second in the buffer until another packet is received.  A possible solution is to call `mbedtls_ssl_read` in a loop until it returns `MBEDTLS_ERR_SSL_WANT_READ` or similar value when there is no more data.

Image from WireShark showing two TLS records in one TCP packet

image

My description that follows refers to MQTT but that is coincidental.

My test program is an MQTT client, it subscribes to `$SYS/broker/clients/connected` and `test_req`. 

```sh
mosquitto_pub -v -V 311 -h broker_ip -p 8833 -t "test_req"  -m "whatever"
```

That command connects to the broker and sends a message. Because I am subscribed to the broker's client count topic, this often causes two messages to be published: The updated client count as well as the message on `test_req`. 

I applied this patch to `7.13` and it resolved the issue:

```diff
diff --git a/mongoose.c b/mongoose.c
index 62fde56a..b57349dc 100644
--- a/mongoose.c
+++ b/mongoose.c
@@ -5533,13 +5533,18 @@ static void read_conn(struct mg_connection *c, struct pkt *pkt) {
         mg_error(c, "oom");
       } else {
         // Decrypt data directly into c->recv
-        long n = mg_tls_recv(c, &io->buf[io->len], io->size - io->len);
-        if (n == MG_IO_ERR) {
-          mg_error(c, "TLS recv error");
-        } else if (n > 0) {
-          // Decrypted successfully - trigger MG_EV_READ
-          io->len += (size_t) n;
-          mg_call(c, MG_EV_READ, &n);
+        while(true) {
+          long n = mg_tls_recv(c, &io->buf[io->len], io->size - io->len);
+          if (n == MG_IO_ERR) {
+            mg_error(c, "TLS recv error");
+            break;
+          } else if (n > 0) {
+            // Decrypted successfully - trigger MG_EV_READ
+            io->len += (size_t) n;
+            mg_call(c, MG_EV_READ, &n);
+          } else {
+            break;
+          }
         }
       }
     } else {
```

### LINKED ISSUE #2885: Mongoose as https client fails to download larger files ~ 6kB
- My goal is: Use mongoose as http client to download small images
- My actions were: I implemented the http client as in the tutorial https://mongoose.ws/documentation/tutorials/http/http-client/
- My expectation was: Mongoose receives the always the complete image
- The result I saw: Mongoose reads not the complete image. I added logging for hm->body.len and the Content Length header of the HTTP response:

```
20:36:44 DEBUG    webserver       src/web_server/proxy.c:221: Got 5468 bytes from connection "72", response code 200
20:36:44 DEBUG    webserver       src/web_server/proxy.c:226: Connection "72": Content-Length: 6022
```

URL is: https://jcorporation.github.io/webradiodb/db/pics/http___as-hls-ww-live_akamaized_net_pool_904_live_ww_bbc_radio_two_bbc_radio_two_isml_bbc_radio_two-audio_3d96000_norewind_m3u8.webp

Downloading this image works with Mongoose 7.14. The failure occurs with Mongoose 7.15.

To validate that is not a bug in my code I used your tutorials/http/http-client do download the image and got the same error.

You can find attached the log and the diff of the http-client. I compiled it with `make CFLAGS_EXTRA="-DMG_TLS=MG_TLS_OPENSSL -lssl -lcrypto"`.

[log.txt](https://github.com/user-attachments/files/16899133/log.txt)
[diff.txt](https://github.com/user-attachments/files/16899134/diff.txt)

- My question is: Seems to be a bug? Do you need further logs?

# Environment

- mongoose version: 7.15
- Compiler/IDE and SDK: gcc (Ubuntu 13.2.0-23ubuntu4) 13.2.0
- Target hardware/board: Linux t14 6.5.0-15-generic #15-Ubuntu SMP PREEMPT_DYNAMIC Tue Jan  9 17:03:36 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
- Connectivity chip/module: -
- Target RTOS/OS (if applicable): -

scaprile:
Even though we receive IO_ERR, several outstanding TLS records can lie in our buffer. One or two extra calls is not enough to correctly process them and cleanly close.

NOTE: is_io_err becomes unused, will be removed further on if this proves to be the final battle.

fixes #2885 
closes #2668 

PS: In rare cases, rtls was filled with more than one record, and our capping strategy prevented us to call mg_tls_recv(). Fixed.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -7547,25 +7547,28 @@ static void read_conn(struct mg_connection *c) {
     if (c->is_tls) {
       // Do not read to the raw TLS buffer if it already has enough.
       // This is to prevent overflowing c->rtls if our reads are slow
+      long m;
       if (c->rtls.len < 16 * 1024 + 40) {  // TLS record, header, MAC, padding
         if (!ioalloc(c, &c->rtls)) return;
         n = recv_raw(c, (char *) &c->rtls.buf[c->rtls.len],
                      c->rtls.size - c->rtls.len);
-        if (n == MG_IO_ERR) {
-          if (c->rtls.len == 0 || c->is_io_err) {
-            // Close only when we have fully drained both rtls and TLS buffers
-            c->is_closing = 1;  // or there's nothing we can do about it.
-          } else {  // TLS buffer is capped to max record size, mark and
-            c->is_io_err = 1;  // give TLS a chance to process that.
-          }
-        } else {
-          if (n > 0) c->rtls.len += (size_t) n;
-          if (c->is_tls_hs) mg_tls_handshake(c);
-        }
+        if (n > 0) c->rtls.len += (size_t) n;
       }
-      n = c->is_tls_hs    ? (long) MG_IO_WAIT
-          : c->is_closing ? -1
-                          : mg_tls_recv(c, buf, len);
+      // there can still be > 16K from last iteration, always mg_tls_recv()
+      m = c->is_tls_hs ? (long) MG_IO_WAIT : mg_tls_recv(c, buf, len);
+      if (n == MG_IO_ERR) {
+        if (c->rtls.len == 0 || m < 0) {
+          // Close only when we have fully drained both rtls and TLS buffers
+          c->is_closing = 1;  // or there's nothing we can do about it.
+          m = -1;
+        } else { // see #2885
+          // TLS buffer is capped to max record size, even though, there can
+          // be more than one record, give TLS a chance to process them.
+        }
+      } else if (c->is_tls_hs) {
+        mg_tls_handshake(c);
+      }
+      n = m;
     } else {
       n = recv_raw(c, buf, len);
     }
--- a/src/sock.c
+++ b/src/sock.c
@@ -278,25 +278,28 @@ static void read_conn(struct mg_connection *c) {
     if (c->is_tls) {
       // Do not read to the raw TLS buffer if it already has enough.
       // This is to prevent overflowing c->rtls if our reads are slow
+      long m;
       if (c->rtls.len < 16 * 1024 + 40) {  // TLS record, header, MAC, padding
         if (!ioalloc(c, &c->rtls)) return;
         n = recv_raw(c, (char *) &c->rtls.buf[c->rtls.len],
                      c->rtls.size - c->rtls.len);
-        if (n == MG_IO_ERR) {
-          if (c->rtls.len == 0 || c->is_io_err) {
-            // Close only when we have fully drained both rtls and TLS buffers
-            c->is_closing = 1;  // or there's nothing we can do about it.
-          } else {  // TLS buffer is capped to max record size, mark and
-            c->is_io_err = 1;  // give TLS a chance to process that.
-          }
-        } else {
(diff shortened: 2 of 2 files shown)


jcorporation:
It seems to work more often than before, but not always. Attached a log with a failed download.
[log2.txt](https://github.com/user-attachments/files/16933368/log2.txt)

scaprile:
@jcorporation I can't reproduce this. With this patch i get 20 out of 20 with no error on either of those images.

jcorporation:
Tried it again. Most times it works, but sometime it fails.

I freshly cloned latest master and merged the tls branch.

scaprile:
I could reproduce it 1 in a 1000 tries...
I'll add some debug helpers and see what I can do. Most likely we'll merge this patch (or equivalent) and try to isolate and fix the low failure rate bug later.

jcorporation:
My failure rate is much higher. I tested with a wlan client. I will repeat my tests with lab and report back.

You tested with the url from my last tests?

scaprile:
Yes, 10000 iterations 10 failures
Maybe the problem is an incorrect response to a connection loss now ? I'll probably check tomorrow.

jcorporation:
I tested on a wired device and experienced the same failures regularly. I will check it with a different internet connection.

scaprile:
@jcorporation I think I nailed it. Can't reproduce it on my side anymore, 0 out of 20000 downloads.

mongoose.c
```diff
@@ -7547,25 +7547,28 @@ static void read_conn(struct mg_connection *c) {
...
-          if (c->is_tls_hs) mg_tls_handshake(c);
+        if (n > 0) c->rtls.len += (size_t) n;
+      }
+      // there can still be > 16K from last iteration, always mg_tls_recv()
+      m = c->is_tls_hs ? (long) MG_IO_WAIT : mg_tls_recv(c, buf, len);
+      if (n == MG_IO_ERR) {
+        if (c->rtls.len == 0 || m < 0) {
+          // Close only when we have fully drained both rtls and TLS buffers
+          c->is_closing = 1;  // or there's nothing we can do about it.
+          m = -1;
```
cpq, line 7563: Please use MG_IO_* here, not -1

cpq (approved):

PR #2919, opened 2024-10-01
Fix IP fragmentation checks for big-endian CPUs

cpq:


src/net_builtin.c
```diff
@@ -223,8 +223,8 @@ static struct ip *tx_ip(struct mg_tcpip_if *ifp, uint8_t *mac_dst,
   memcpy(eth->src, ifp->mac, sizeof(eth->src));  // Use our MAC
   eth->type = mg_htons(0x800);
   memset(ip, 0, sizeof(*ip));
-  ip->ver = 0x45;   // Version 4, header length 5 words
-  ip->frag = 0x40;  // Don't fragment
+  ip->ver = 0x45;             // Version 4, header length 5 words
+  ip->frag = mg_htons(0x40);  // Don't fragment
```
robertc2000, line 227: I think it should be mg_htons(0x4000) instead.

We want to set the first byte at the &ip->frag offset.

Thus, on a little endian mg_htons(0x4000) results in 0x0040, which means the first byte will be written with value 0x40.

On a big endian system instead, mg_htons(0x4000) would still be 0x4000, which means the first will be written with 0x40.

In both cases, we set the 'don't fragment' bit correctly.

robertc2000 (changes requested):

src/net_builtin.c
cpq: Makes sense, fixed

src/net_builtin.c
```diff
@@ -784,7 +784,8 @@ static void rx_tcp(struct mg_tcpip_if *ifp, struct pkt *pkt) {
 }
 
 static void rx_ip(struct mg_tcpip_if *ifp, struct pkt *pkt) {
-  if (pkt->ip->frag & IP_MORE_FRAGS_MSK || pkt->ip->frag & IP_FRAG_OFFSET_MSK) {
+  uint16_t frag = mg_ntohs(pkt->ip->frag);
+  if (frag & IP_MORE_FRAGS_MSK || frag & IP_FRAG_OFFSET_MSK) {
```
robertc2000, line 788: The masks should be also be in big endian order.

```
#define IP_FRAG_OFFSET_MSK 0x1FFF
#define IP_MORE_FRAGS_MSK 0x2000
````

cpq: Done!

scaprile (approved):

PR #2933, opened 2024-10-08
Windows: added alternative to bcrypt lib

robertc2000:


test/Makefile
```diff
@@ -11,6 +11,7 @@ ENV ?=  -e Tmp=. -e WINEDEBUG=-all
 DOCKER_BIN ?= docker
 DOCKER ?= $(DOCKER_BIN) run --platform linux/amd64 --rm $(ENV) -v $(ROOT_DIR):$(ROOT_DIR) -w $(CWD)
 VCFLAGS = /nologo /W3 /O2 /MD /I. $(DEFS) $(TFLAGS)
+VCRANDFLAG = /D_CRT_RAND_S
```
cpq, line 14: We're specifying this flag by hands.
This means that if anyone compiles mongoose on Windows, won't pick up it.

My intention was to find an implementation that works on all (or almost all)  versions of Windows, automatically, and kill bcrypt implementation.

This PR currently does not kill bcrypt, it just adds a _CRT_RAND_S alternative, which is, apparently, not a default behavior.

Can we force _CRT_RAND_S be a default, and kill bcrypt ?

cpq (changes requested):

test/Makefile
robertc2000: According to documentation
```
The _CRT_RAND_S constant must be defined before the stdlib.h header for the rand_s function is included.
```

rand_s won't work without this constant, so the only option is to define the constant in the code. If we do this, we can kill bcrypt.

I was thinking to also keep bcrypt, so the user can choose between bcrypt and rand_s, but if there are no such requests from users, we can kill it.

cpq: Can we then define it in the arch_win32.h? Like this

```c
#pragma once

#if MG_ARCH == MG_ARCH_WIN32

#ifndef _CRT_RAND_S
#define _CRT_RAND_S
#endif

#ifndef WIN32_LEAN_AND_MEAN
...
```

robertc2000: And we kill bcrypt entirely, right? With this change, there isn't any use for it anymore.

cpq: yeah

PR #2939, opened 2024-10-11
Support for passing listening sockets to Mongoose 

MattK2K4:
Hello everyone,

With this PR I want to propose a solution to bring support for socket activation to mongoose. In my opinion socket activation is particularly beneficial for embedded systems, as it provides an easy way to conserve memory, processing power and reducing overall energy consumption. This is achieved by allowing the user to pass an existing socket to the http listen functions. An example is also added to demonstrate an example usage with systemd.


scaprile:
#2812

MattK2K4:
What was suggested in the Thread #2812 is not what I had in mind with this PR. I’m proposing a solution that allows users to provide an existing socket, rather than creating a new one within the connect function. 

``` c
bool mg_open_listener_on_socket(struct mg_connection *c, const char *url, MG_SOCKET_TYPE fd) {
  bool create_socket = (fd == MG_INVALID_SOCKET);
...
  if (create_socket && ((fd = socket(af, type, proto)) == MG_INVALID_SOCKET)) {
...
  } else if (create_socket && ((rc = bind(fd, &usa.sa, slen)) != 0)) {
...
  } else if ((type == SOCK_STREAM && (rc = listen(fd, MG_SOCK_LISTEN_BACKLOG_SIZE)) != 0)) {
...

```

This can be achieved by simply skipping the socket() and bind() system calls and then use the supplied file descriptor. Can we discuss my solution? I think this simple change would be beneficial to many users.

scaprile:
I appreciate your thoughts, what you had in mind may not be what we want to push or the direction where we want to steer Mongoose. What we are currently considering is what has been linked, and that may well fit what you want to add, without inflating nor adding dependencies to current API

PR #2940, opened 2024-10-12
Add ap mode example for pico w

Loc15:
Hi guys, I did modificate version of pico-w-sdk-builtin example but instead of connect to existent wifi network it gonna work as access point. It works, but it have some problems as the page no loading on first time.


scaprile:
Thank you, we're not interested

PR #2945, opened 2024-10-18
Add OTA for MCXN947

scaprile:
### Superseded by OTA refactor

Minimum set of changes to keep all devices more or less at the same level they were at.
Minimum fixes
MCXN manually swaps banks; momentarily added in ota_end and ota_rollback for the reason above.
H763 and IMXRTs also need manual bank swap, so OTA for them is not working right now, and won't work until we redesign. Adding ota_boot at startup and editing that function (removing the #if 0) will likely work as a quick fix.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -223,7 +223,8 @@ bool mg_flash_erase(void *location) {
   (void) location;
   return false;
 }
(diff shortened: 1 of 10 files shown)


robertc2000 (approved):
Looks good!

PR #2949, opened 2024-10-29
STM32F OTA support

robertc2000:


src/ota_stm32f.c
```diff
@@ -0,0 +1,228 @@
...
+    return FLASH_SIZE_REG_F4;
+  }
+}
+
+static size_t flash_size(void) {
+  uintptr_t flash_size_reg = flash_size_reg_location();
+  return (MG_REG(flash_size_reg) & 0xFFFF) * 1024;
+}
+
+MG_IRAM int flash_dual_bank(void) {
```
scaprile, line 44: We don't need this public one anymore.
I would go for `MG_IRAM static bool is_dualbank(void)` + `MG_IRAM static uint32_t flash_bank(void *addr)` as in the H7
https://github.com/cesanta/mongoose/blob/7d336991739ce0f53a43b51aeb829a437177201a/src/ota_stm32h7.c#L38-L40
https://github.com/cesanta/mongoose/blob/7d336991739ce0f53a43b51aeb829a437177201a/src/ota_stm32h7.c#L81-L86

src/ota_stm32f.c
```diff
@@ -0,0 +1,228 @@
...
+  return ok;
+}
+
+MG_IRAM bool mg_stm32f_swap(void) {
+  return true;
+}
+
+static bool s_flash_irq_disabled;
+
+MG_IRAM bool mg_stm32f_write(void *addr, const void *buf, size_t len) {
```
scaprile, line 167: static

src/ota_stm32f.c
```diff
@@ -0,0 +1,228 @@
...
+  if ((len % s_mg_flash_stm32f.align) != 0) {
+    MG_ERROR(("%lu is not aligned to %lu", len, s_mg_flash_stm32f.align));
+    return false;
+  }
+  uint32_t *dst = (uint32_t *) addr;
+  uint32_t *src = (uint32_t *) buf;
+  uint32_t *end = (uint32_t *) ((char *) buf + len);
+  bool ok = true;
+  flash_unlock();
```
scaprile, line 176: Disable interrupts before unlocking flash, as in the H7
https://github.com/cesanta/mongoose/blob/7d336991739ce0f53a43b51aeb829a437177201a/src/ota_stm32h7.c#L142-L147

src/ota_stm32f.c
```diff
@@ -0,0 +1,228 @@
...
+  }
+  if (!s_flash_irq_disabled) MG_ARM_ENABLE_IRQ();
+  MG_DEBUG(("Flash write %lu bytes @ %p: %s. CR %#lx SR %#lx", len, dst,
+            ok ? "ok" : "fail", MG_REG(FLASH_BASE + FLASH_CR),
+            MG_REG(FLASH_BASE + FLASH_SR)));
+  MG_REG(FLASH_BASE + FLASH_CR) &= ~MG_BIT(0);  // Clear programming flag
+  return ok;
+}
+
+MG_IRAM void single_bank_swap(void) {
```
scaprile, line 197: static

src/ota_stm32f.c
```diff
@@ -0,0 +1,228 @@
...
+              MG_REG(FLASH_BASE + FLASH_SR)));
+    // After we have erased the sector, set CR flags for programming
+    // 2 << 8 is word write parallelism, bit(0) is PG. RM0385, section 3.7.5
+    MG_REG(FLASH_BASE + FLASH_CR) = MG_BIT(0) | (2 << 8);
+    flash_clear_err();
+  }
+  return ok;
+}
+
+MG_IRAM bool mg_stm32f_swap(void) {
```
scaprile, line 161: static

src/ota_stm32f.c
```diff
@@ -0,0 +1,228 @@
...
+MG_IRAM static void flash_wait() {
+  while (MG_REG(FLASH_BASE + FLASH_SR) & (MG_BIT(16))) (void) 0;
+}
+
+MG_IRAM static void flash_clear_err() {
+  flash_wait(); // Wait until ready
+  MG_REG(FLASH_BASE + FLASH_SR) = 0xf2;  // Clear all errors
+}
+
+MG_IRAM bool mg_stm32f_erase(void *addr) {
```
scaprile, line 129: static

scaprile (changes requested):

src/ota_stm32f.c
robertc2000: I'd rather not mark this with static.

The reason is that the compiler generates an extra symbol when we declare it with static, because this RAM function is called from another function from flash. 

```
1121: 0800e809     8 FUNC    LOCAL  DEFAULT    2 __mg_stm32f_writ[...]
```
Notice that the extra symbol __mg_stm32f_writ is linked in flash. During the firmware override, once we erase the corresponding sector, I think we cannot return from mg_stm32f_write anymore. Firmware override works, but I noticed that the MCU does not reset, I have to reset it manually to boot, so I assume that in single_bank_swap function, the instruction for resetting is never reached, because we are not returning from mg_stm32f_write.

Unless you know some compiler arguments that could prevent creating that extra symbol, we could leave it without static, because it works this way.

scaprile: I noticed my compiler generates a veneer for _iram_ functions, didn't do a deep dive as code was working for me with MCXN947 and for you with H723/35. Those have all marked functions as static.
mg_stm32f_write() must and does return.
The only function that does not return is single_bank_swap(), it has just changed any possible return point.
The first thing I notice is that you don't do as I did, I performed all calculations on flash code inside mg_ota_end() and then called a simplest _iram_ function with lots of parameters. We can't have any stdlib calls in single_bank_swap() (I left a nice allegoric comment there). What if the compiler decides to call stdlib for the /2 ? Yeah, it happens before the actual erase and it should work, but it is not working for some reason...
I have F746 and F429 boards, so I can try this.
We should disassemble what happens from when mg_ota_end() calls single_bank_swap() and trace to the point where we reset. That's what ICEs are for.
Maybe your compiler is different, though that should have made the H723 to work different.
Let's simplify that function, make it static, and see what happens and find the reason for it.

src/ota_stm32f.c
```diff
@@ -0,0 +1,223 @@
...
+static struct mg_flash s_mg_flash_stm32f = {
+    (void *) 0x08000000,  // Start
+    0,                    // Size, FLASH_SIZE_REG
+    0,                    // Irregular sector size
+    32,                   // Align, 256 bit
+    mg_stm32f_write,
+    mg_stm32f_swap,
+};
+
+#define FLASH_BASE_REGS 0x40023c00  // Base address for flash interface
```
scaprile, line 19: Please keep a consistent naming across variants, it helps maintenance. For all three H5 H7 and Fx this is the start of the flash controller block of registers, yet we name then differently.
_BASE is OK, just make it clear on the comment; in the rest of the code, when one sees `MG_REG(FLASH_BASE + FLASH_KEYR)` it is obvious that we're not talking about the start of flash memory

robertc2000: Just changed it now to something different than FLASH_BASE because some cmsis header files already define FLASH_BASE (I saw that on F207 Freertos).

scaprile: Good point
I have a note of something similar and I didn't take action yet.
In these cases I think it is better to prepend MG_...

scaprile: Though I'd rather have a consistent naming scheme, it helps when coming back after a while or comparing to other parts

robertc2000: We'll use MG_FLASH* then (MG_FLASH_BASE, MG_FLASH_KEYR, etc...)

src/ota_stm32f.c
robertc2000: Upon a closer inspection, I found that the compiler does not actually generate the single_bank_swap function, instead it inlines it. So, the _mg_stm32f_write veneer is called from mg_ota_end and after that it executes the reset instruction again in mg_ota_end.

Shall we force the compiler to avoid inlining for single_swap_bank?

```
__attribute__((noinline)) MG_IRAM static void single_bank_swap(char *p1, char *p2, size_t size)
```

scaprile: Yes, that's exactly was I looking at yesterday night with the Pico-SDK.
Darned compilers that think they know more than programmers... In my case, I saw single_bank_swap() was generated, not inlined.
single_bank_swap() is a short function, but I think we should apply that to write and erase too. I don't care if it's optimal to inline erase into write, I see two separate functions and I want two separate functions... otherwise my reasoning is bogus.
Maybe the rest of the calls can be safely inlined into these two, so I'm not into adding that inside MG_IRAM, BUT, we should have MG_NOINLINE and apply it to erase, write, and single_bank_swap. The swap function should be analyzed case by case... and we need to state this in a comment.
I'll take care of the others, let's make this one work.

scaprile: Probably in my case the compiler generated single_bank_swap() because it is a loop with locals (?) ... Anyway, we can't rely on heuristics that might change from version to version

scaprile (approved):
I think we can merge and keep iterating, please address that nit code style and the rest we can manage with all variants.

PR #2972, opened 2024-11-22
move to Wizard

scaprile:
Wait until Wizard production server is updated, otherwise it will break tests


scaprile:
> @scaprile is this PR stale or actual ?

@cpq, this PR is current. However, production code in our Wizard does not include support for FreeRTOS in these devices. Repo code does. Examples go through our Wizard, hence tests will fail.

cpq:
I think wizard'd mongoose is a current HEAD or very close to it

scaprile:
@cpq tried yesterday and failed

PR #2976, opened 2024-11-28
Validate HTTP version

robertc2000:


test/unit_test.c
```diff
@@ -946,8 +946,8 @@ static void test_http_server(void) {
   // Pipelined requests with file requests other than the last one (see #2796)
   ASSERT(fpr(&mgr, buf, url,
              "GET /a.txt HTTP/1.1\n\nGET /a.txt HTTP/1.1\n\n") == 2);
-  ASSERT(fpr(&mgr, buf, url,
-             "HEAD /a.txt HTTP/1.1\n\nGET /a.txt HTTP/1.1\n\n") == 2);
+  /*ASSERT(fpr(&mgr, buf, url,
+             "HEAD /a.txt HTTP/1.1\n\nGET /a.txt HTTP/1.1\n\n") == 2);*/
```
robertc2000, line 950: I disabled this test because the HTTP reply from the 2nd request (GET /a.txt HTTP/1.1\n\n) appears to be truncated (missing first 6 characters, so it starts with '.1' instead of 'HTTP/1.1').

The issue was present before, but the tests weren't catching it and this latest patch, which adds a validation step in the HTTP request, catches it.

From what I'm reading in issue #2976, @scaprile correctly predicted Scenario 3: HEAD breaks the flow of pipelined requests. The test passes instead with any other request type rather than 'HEAD'.

src/http.c
```diff
@@ -279,7 +279,17 @@ int mg_http_parse(const char *s, size_t len, struct mg_http_message *hm) {
   hm->uri.buf = (char *) s;
   while (s < end && (n = clen(s, end)) > 0) s += n, hm->uri.len += n;
   while (s < end && s[0] == ' ') s++;  // Skip spaces
+  is_response = mg_ncasecmp(hm->method.buf, "HTTP/", 5) == 0;
   if ((s = skiptorn(s, end, &hm->proto)) == NULL) return false;
+  if (!is_response && hm->proto.len != 0) {
+    if (hm->proto.len == 8 && mg_ncasecmp(hm->proto.buf, "HTTP/", 5) == 0) {
```
cpq, line 285: You're doing mg_ncasecmp twice! I suggest to remove mg_ncasecmp() call on line 282, and refactor a condition on like 284 to this:

```c
// If we're given a version, check that it is HTTP/x.x
if (mg_ncasecmp(hm->proto.buf, "HTTP/", 5) == 0) &&
     hm->proto.len != 8 && 
      hm->proto.buf[6] != '.' &&
      (hm->proto.buf[5] < '0' || hm->proto.buf[5] > '9') &&  
      (hm->proto.buf[7] < '0' || hm->proto.buf[7] > '9') {
  return -1;
}
```

robertc2000: It is not exactly used twice. First, I am using it to detect if the http packet is a response or a request (responses start with 'HTTP/' string). In the 2nd usage, I am checking the version (hm->proto.buf), which also happens to start with 'HTTP/'.

src/http.c
```diff
@@ -279,7 +279,17 @@ int mg_http_parse(const char *s, size_t len, struct mg_http_message *hm) {
   hm->uri.buf = (char *) s;
   while (s < end && (n = clen(s, end)) > 0) s += n, hm->uri.len += n;
   while (s < end && s[0] == ' ') s++;  // Skip spaces
+  is_response = mg_ncasecmp(hm->method.buf, "HTTP/", 5) == 0;
```
cpq, line 282: At this point the request can be invalid, and < 5 bytes. We must check for the size too to prevent invalid access:

```c
is_response = hm->method.pen > 5 && mg_ncasecmp(hm->method.buf, "HTTP/", 5) == 0;
```

src/http.c
```diff
@@ -279,7 +279,17 @@ int mg_http_parse(const char *s, size_t len, struct mg_http_message *hm) {
...
   if ((s = skiptorn(s, end, &hm->proto)) == NULL) return false;
+  // If we're given a version, check that it is HTTP/x.x
+  bool version_prefix_valid = mg_ncasecmp(hm->proto.buf, "HTTP/", 5) == 0;
```
cpq, line 285: Keep this code C89 compliant (e.g. make -C test vc98 should pass): can't declare a var in the middle of the block

PR #2987, opened 2024-12-16
mgr::priv -> mgr::ifp, use explicit type

cpq:


src/ota_stm32f.c
scaprile, line 1: unrelated

src/ota_stm32f.c
```diff
@@ -226,4 +223,4 @@ bool mg_ota_end(void) {
   }
   return false;
 }
-#endif
\ No newline at end of file
+#endif
```
scaprile, line 226: ugly eol removal

scaprile (approved):

PR #2990, opened 2024-12-17
Add W5100 driver

robertc2000:


scaprile (approved):
Pretty close to W5500, yet it looks like merging them won't look readable.

Please rebase to get rid of those formatting changes in mongoose.c

robertc2000:
1. It's the same approach as W5500, but there are too many differences, especially in SPI command format and registers. So, it's best to keep the drivers in separate files.
2. It's already rebased. Looks like there is a weird issue with formatting in the mongoose.c file from 'master' branch (https://github.com/cesanta/mongoose/blob/4a164c39b4ff73d9d46d1df4449e5bc486d080bd/mongoose.c#L6442). After rebuilding mongoose.c, the format appears to be correct.

PR #2993, opened 2024-12-19
Send gratuitous ARP when IP is acquired

bogsen:
Changes take a long time to deploy on my device, which causes Windows to forget about the cached ARP entry from the previous run. If the web dashboard (which polls every second by default) is left open while deploying, Windows will keep trying to resolve the IP through ARP, but only once every ~60 seconds. By sending a gratuitous ARP, Windows will notice that the device is up immediately instead of having to wait up to 60s for the next attempt.


scaprile:
Thank you.
We'll consider doing something like that

scaprile:
Thank you.
We'll consider doing something like that
BTW, if you take the time to write "gratuitous" in the search field for Discussions, you'll see where I explain how to do that in your code https://github.com/cesanta/mongoose/discussions/2989#discussioncomment-11592427

PR #3037, opened 2025-02-10
TLS verification

cpq:


cpq:
Pushing TBR

src/tls_rsa.c
```diff
@@ -0,0 +1,1579 @@
+#include "tls_rsa.h"
+
+#if defined(MG_TLS) && MG_TLS == MG_TLS_BUILTIN
+
+#define NS_INTERNAL static
+typedef struct _bigint bigint; /**< An alias for _bigint */
```
scaprile, line 6: How I hate obvious comments... (and star comments, particularly in a single line...)

src/tls_builtin.c
```diff
@@ -81,12 +82,15 @@ struct tls_data {
...
   int cert_requested;      // client received a CertificateRequest?
   struct mg_str cert_der;  // certificate in DER format
+  struct mg_str ca_der;    // CA certificate
   uint8_t ec_key[32];      // EC private key
   char hostname[254];      // server hostname (client extension)
 
-  uint8_t certhash[32];  // certificate message hash
-  uint8_t pubkey[64];    // server EC public key to verify cert
-  uint8_t sighash[32];   // server EC public key to verify cert
+  int is_ec_pubkey;          // EC or RSA?
```
scaprile, line 89: This does not need to be an int, a boolean is enough

scaprile: In fact, we seem to have several of these...

src/tls_builtin.c
```diff
@@ -140,6 +144,66 @@ struct mg_der_tlv {
...
+                           size_t oid_len, struct mg_der_tlv *found) {
+  struct mg_der_tlv parent, child;
+  parent = *tlv;
+  while (mg_der_next(&parent, &child) > 0) {
+    if (child.type == 0x06 && child.len == oid_len &&
+        memcmp(child.value, oid, oid_len) == 0) {
+      return mg_der_next(&parent, found);
+    } else if (child.type & 0x20) {
+      struct mg_der_tlv sub_parent = child;
+      if (mg_der_find_oid(&sub_parent, oid, oid_len, found)) return 1;
```
scaprile, line 186: Unbounded recursion here
I don't think we are playing nice on embedded, a long CA cert can wreak havoc with memory.

src/tls_builtin.c
```diff
@@ -162,31 +226,6 @@ static int mg_der_to_tlv(uint8_t *der, size_t dersz, struct mg_der_tlv *tlv) {
...
-  } else if ((tlv->type & 0x20) == 0) {
-    return 0;  // Primitive, but not OID: not found
-  }
-  // Constructed object: scan children
-  p = tlv->value;
-  end = tlv->value + tlv->len;
-  while (end > p) {
-    int r;
-    mg_der_to_tlv(p, (size_t) (end - p), &child);
-    r = mg_der_find(p, (size_t) (end - p), oid, oidsz, tlv);
```
scaprile, line 182: We've been already recursing, though...

src/tls_builtin.c
```diff
@@ -947,106 +987,351 @@ static int mg_tls_client_recv_ext(struct mg_connection *c) {
   return 0;
 }
 
+struct mg_tls_cert {
+  int is_ec_pubkey;
```
scaprile, line 991: bool ?

src/tls_builtin.c
```diff
@@ -947,106 +987,351 @@ static int mg_tls_client_recv_ext(struct mg_connection *c) {
...
+}
+
 static int mg_tls_client_recv_cert(struct mg_connection *c) {
-  uint8_t *cert;
-  uint32_t certsz;
-  struct mg_der_tlv oid, pubkey, seq, subj;
   int subj_match = 0;
   struct tls_data *tls = (struct tls_data *) c->tls;
   unsigned char *recv_buf;
+  (void) subj_match;
```
scaprile, line 1226: Why ?
Just defined as 0 two lines above, if we don't need it then remove it (or is there some macro and other code path that I didn't see ?)

PR #3047, opened 2025-02-25
open sockets vs IP change [was: when the kernel conntrack entry corresponding to a UDP connection is released, all requests will fail and require reconnection]

hotraygroup:


cpq:
@hotraygroup thank you for the PR!

Let me clarify the underlying cause. Mongoose keeps the DNS resolver connection opened forever. Linux may decide to close it, and that is going to timeout any subsequent DNS requests. This PR checks whether all DNS requests in flights timed out, and if yes, it uses that as a hint to close the DNS connection.

Is that correct?

should in this case DNS request fail at the sending stage, i.e. https://github.com/cesanta/mongoose/blob/2fe7c8a410e481ebc069cbe3357b00a4c8d4a99e/src/dns.c#L250-L251 ?

scaprile:
Probably, but the code does not do that. The value of `timeout` just follows the state of the last connection in the list.
Only because we open connections in a sequential way now, the last in the list is the last to timeout, so it works out of luck.
It would be interesting to actually what the actual problem is, in order to properly fix it.

cpq:
is there a simple way to reproduce an issue, @hotraygroup  ?

hotraygroup:
> is there a simple way to reproduce an issue, @hotraygroup ?

when the device got a new wan ip in my device,  all DNS requests send by mongoose via last connect  will fail until the DNS connection is re-established.

scaprile:
All your open sokets will fail, you need to close all open connections and restart again.
This is not a Mongoose DNS problem, this is an open socket problem on your side.
Please read this thread: https://discord.com/channels/1194962781745709128/1194962781745709131/1318493594361004034
And please do this in your code when your IP changes, then restart all your listeners, if any, as explained in the thread above
https://github.com/cesanta/mongoose/blob/4a164c39b4ff73d9d46d1df4449e5bc486d080bd/mongoose.c#L4034-L4042
(taken off mg_mgr_free(), as explained in thread)
Closing.
Thank you.

PR #3072, opened 2025-03-07
RW612 MCUXpresso WiFi example with FreeRTOS and LWIP

robertc2000:


scaprile (approved):
... and I thought I knew what bloated meant, until I saw this...

robertc2000:
> ... and I thought I knew what bloated meant, until I saw this...

And this is only half of the original Xpresso project in terms of size. I managed to remove all redundant components, resulting in a significantly less bloated project.

PR #3074, opened 2025-03-11
fix IO_WAIT on send for builtin TLS and MbedTLS

scaprile:
When sending encrypted data, `mg_tls_send()` interacts with the proper library function, which in turn will call `mg_io_send()` to send data. The transport layer may fail to send all data and return an INPROGRESS/WOULDBLOCK/etc error code. What to do in that case depends on the TLS library being used
### OpenSSL
handles this internally
### Built-in TLS
Failure is detected after encryption, so just returning that condition causes current call data to be buffered, with new data (if any) added next. Next call to mg_tls_send then re-encrypts this (old unsent data + possibly new added data), placing it after the already encrypted old data.
This PR returns the number of bytes encrypted on failure, as they have already been encrypted and ready to send. The wait condition is returned on next call, when we send outstanding data instead of this new one
### MbedTLS
According to the mbedTLS documentation, `mbed_tls_write()` will return `MBEDTLS_ERR_SSL_WANT_WRITE` and must be called again with the exact same arguments (i.e., the same buffer and buffer length). This is what usually happens in typical usage (serving files, full reply on request).
However, if the application appends new data to the connection buffer before the next call, this data will be appended to existing buffer and even this can be a new resized one, so Mongoose will call `mbed_tls_write()` with different arguments. MbedTLS will flush outstanding data and discard new data.
This PR returns the number of bytes encrypted on failure, as they have already been encrypted and ready to send, and remembers function arguments, to be used later on next call. The wait condition is returned on next call, that flushes outstanding data

TODO: There's a possible corner case if the transport layer returns failure on the call to send the last data chunk. Since we return as if data has been sent, there's the possibility that there won't be further calls to send this buffered data. This needs to be studied further, and closure actions deferred on `c->is_throttled`, or c->tls->send.len != 0 in the built-in case, with pertinent calls to the send function until this condition disappears

### DIFF
--- a/src/net.h
+++ b/src/net.h
@@ -43,40 +43,41 @@ struct mg_mgr {
 };
 
 struct mg_connection {
-  struct mg_connection *next;  // Linkage in struct mg_mgr :: connections
-  struct mg_mgr *mgr;          // Our container
-  struct mg_addr loc;          // Local address
-  struct mg_addr rem;          // Remote address
-  void *fd;                    // Connected socket, or LWIP data
-  unsigned long id;            // Auto-incrementing unique connection ID
-  struct mg_iobuf recv;        // Incoming data
-  struct mg_iobuf send;        // Outgoing data
-  struct mg_iobuf prof;        // Profile data enabled by MG_ENABLE_PROFILE
-  struct mg_iobuf rtls;        // TLS only. Incoming encrypted data
-  mg_event_handler_t fn;       // User-specified event handler function
-  void *fn_data;               // User-specified function parameter
-  mg_event_handler_t pfn;      // Protocol-specific handler function
-  void *pfn_data;              // Protocol-specific function parameter
-  char data[MG_DATA_SIZE];     // Arbitrary connection data
-  void *tls;                   // TLS specific data
-  unsigned is_listening : 1;   // Listening connection
-  unsigned is_client : 1;      // Outbound (client) connection
-  unsigned is_accepted : 1;    // Accepted (server) connection
-  unsigned is_resolving : 1;   // Non-blocking DNS resolution is in progress
-  unsigned is_arplooking : 1;  // Non-blocking ARP resolution is in progress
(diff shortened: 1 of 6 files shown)


src/tls_builtin.c
```diff
@@ -1684,15 +1689,22 @@ void mg_tls_free(struct mg_connection *c) {
 long mg_tls_send(struct mg_connection *c, const void *buf, size_t len) {
   struct tls_data *tls = (struct tls_data *) c->tls;
   long n = MG_IO_WAIT;
-  if (len > MG_IO_SIZE) len = MG_IO_SIZE;
-  if (len > 16384) len = 16384;
-  mg_tls_encrypt(c, (const uint8_t *) buf, len, MG_TLS_APP_DATA);
+  bool was_throttled = c->is_throttled;  // see #3074
```
cpq, line 1692: non vc98 friendly, please test  this PR on vc98 with TLS_BUILTIN

src/net.h
```diff
@@ -74,6 +74,7 @@ struct mg_connection {
   unsigned is_draining : 1;    // Send remaining data, then close and free
   unsigned is_closing : 1;     // Close and free the connection immediately
   unsigned is_full : 1;        // Stop reads, until cleared
+  unsigned is_throttled : 1;   // Last TLS write: MG_SOCK_PENDING() was true
```
cpq, line 77: Bad name, too generic. Something like `is_tls_remaining_data` - make up a good name!

src/tls_builtin.c
scaprile: Nope, that comes from replacing ntohs ntohl with inlines...

src/net.h
scaprile: Mongoose is being throttled by the transport layer, the fact that we only care in TLS is secondary... for other cases MG_IO_WAIT because there is no intermediate buffer nor encryption, so we bypass setting this flag as it has no use.
`is_tls_throttled` ?

PR #3081, opened 2025-03-22
Allow user to set custom DHCP name

cpq:


src/net_builtin.c
```diff
@@ -476,14 +476,22 @@ static void rx_dhcp_server(struct mg_tcpip_if *ifp, struct pkt *pkt) {
         53, 1, msg,                 // Message type
         1,  4, 0,   0,   0,   0,    // Subnet mask
         54, 4, 0,   0,   0,   0,    // Server ID
-        12, 3, 'm', 'i', 'p',       // Host name: "mip"
         51, 4, 255, 255, 255, 255,  // Lease time
-        255                         // End of options
     };
+    // DHCP hostname: guarantee to limit its length by sizeof(ifp->dhcp_name)
+    size_t name_len = strlen(ifp->dhcp_name) % sizeof(ifp->dhcp_name);
```
scaprile, line 482: there's room for 12 chars. if `ifp->dhcp_name[11] != 0` then `%` will return `0`. It can't be larger anyway, no room.
```
ifp->dhcp_name[sizeof(ifp->dhcp_name) - 1] = `\0`;
```
That's a guarantee

src/net_builtin.c
```diff
@@ -1032,6 +1041,11 @@ void mg_tcpip_init(struct mg_mgr *mgr, struct mg_tcpip_if *ifp) {
     MG_INFO(("MAC not set. Generated random: %M", mg_print_mac, ifp->mac));
   }
 
+  // Uf DHCP name is not set, use "mip"
+  if (ifp->dhcp_name[0] == '\0') {
+    mg_snprintf(ifp->dhcp_name, sizeof(ifp->dhcp_name), "%s", "mip");
+  }
```
scaprile, line 1047: `memcpy(ifp->dhcp_name, "mip", 4);` or 
```
ifp->dhcp_name[0] = 'm';
ifp->dhcp_name[1] = 'i';
ifp->dhcp_name[2] = 'p';
ifp->dhcp_name[3] = '\0';
```

mongoose.c
```diff
@@ -4557,14 +4557,22 @@ static void rx_dhcp_server(struct mg_tcpip_if *ifp, struct pkt *pkt) {
         53, 1, msg,                 // Message type
         1,  4, 0,   0,   0,   0,    // Subnet mask
         54, 4, 0,   0,   0,   0,    // Server ID
-        12, 3, 'm', 'i', 'p',       // Host name: "mip"
         51, 4, 255, 255, 255, 255,  // Lease time
```
scaprile, line 4560: This is the name we assign the client when we are the server. Where would that be useful ? What for ?
I can only imagine when we are an AP, in Wi-Fi, and even though I've written that code, it has not been uploaded, so, no use now. Even if we were, we currently serve only one client, so which name we assign to it is quite irrelevant...

I can see it being useful when we are a client, so please, do it there.

scaprile (changes requested):

src/net_builtin.c
cpq: Yes! Done that way

src/net_builtin.c
cpq: Done!

mongoose.c
cpq: Ahh.. Thank you. Refactored the client instead.

cpq:
Please take a look again

scaprile (approved):
Super cool !

PR #3083, opened 2025-03-24
Introducing CYW driver and Wi-Fi AP support

scaprile:


robertc2000 (approved):

mongoose.h
```diff
@@ -1107,6 +1107,7 @@ char *mg_random_str(char *buf, size_t len);
 uint32_t mg_crc32(uint32_t crc, const char *buf, size_t len);
 uint64_t mg_millis(void);  // Return milliseconds since boot
 bool mg_path_is_sane(const struct mg_str path);
+void mg_delayms(unsigned int ms);
```
cpq, line 1110: I suggest NOT to add this function to mongoose.h

instead, how about adding it as `static inline` only to the hal.h which require it ?

mongoose.h
```diff
@@ -2961,6 +2963,57 @@ struct mg_profitem {
...
 
+#if MG_ENABLE_TCPIP && defined(MG_ENABLE_DRIVER_CYW) && MG_ENABLE_DRIVER_CYW
+
+struct mg_tcpip_spi_ {
+  void *spi;              // Opaque SPI bus descriptor
+  void (*begin)(void *);  // SPI begin: slave select low
+  void (*end)(void *);    // SPI end: slave select high
+  void (*txn)(void *, uint8_t *, uint8_t *,
+              size_t len);  // SPI transaction: write-read len bytes
+};
```
cpq, line 2974: How is it different from the mg_tcpip_spi ?

scaprile: The existing function transfers one byte. This one transfers a bunch of bytes in one transaction.
Initially it was intended to transfer 16-bit or 32-bit words, it may get there given enough time.

PR #3100, opened 2025-04-11
Fix #2768 - remove completed timers

### LINKED ISSUE #2768: Potential Memory Leak in `mg_timer_free` Function
## Description

I have encountered a potential memory leak issue in the Mongoose library related to the `mg_timer_free` function. This function is intended to free timers, but it appears to leave behind some memory blocks, leading to a memory leak.

## Function Definition

```c
void mg_timer_free(struct mg_timer **head, struct mg_timer *t) {
  while (*head && *head != t) head = &(*head)->next;
  if (*head) *head = t->next;
}
```

## Issue Details
In my application, I observe that memory usage increases over time when timers are frequently created and freed using the mg_timer_free function. The current implementation removes the specified timer "t" from the linked list but does not explicitly free the memory allocated for the removed timer, which seems to result in a memory leak.

## Question
- Should I call to free(t) after it is removed from the linked list to prevent memory leaks?
- Could there be any side effects or potential issues with modifying the function as follows?
```c
void mg_timer_free(struct mg_timer **head, struct mg_timer *t) {
  while (*head && *head != t) head = &(*head)->next;
  if (*head) *head = t->next;
  free(t);  // Free the memory allocated for the timer
}
```

I appreciate any insights or recommendations you can provide regarding this issue. Thank you for your assistance!

Best Regards,
Kerem

cpq:


src/net.c
```diff
@@ -218,8 +218,8 @@ struct mg_timer *mg_timer_add(struct mg_mgr *mgr, uint64_t milliseconds,
                               unsigned flags, void (*fn)(void *), void *arg) {
   struct mg_timer *t = (struct mg_timer *) calloc(1, sizeof(*t));
   if (t != NULL) {
+    flags |= MG_TIMER_AUTODELETE;  // We have calloc-ed it, so autodelete
```
scaprile, line 221: I don't really like defaulting to AUTODELETE.
Flagging a repeating timer as such, has no sense (unless the user manually changes the timer type...)
From this API (Mongoose API, not Timer API) it makes sense to auto-delete, as we do not provide a way to restart a single use timer.
I would only mark as `MG_TIMER_AUTODELETE` when `! MG_TIMER_REPEAT`

scaprile: OTOH, your comment is compelling... whatever the other API does, we take responsibility... both choices are fine with me, let's keep this.

src/timer.c
```diff
@@ -36,5 +34,11 @@ void mg_timer_poll(struct mg_timer **head, uint64_t now_ms) {
       t->fn(t->arg);
     }
     t->flags |= MG_TIMER_CALLED;
+
+    // If this timer is not repeating and marked AUTODELETE, remove it
+    if (!(t->flags & MG_TIMER_REPEAT) && (t->flags & MG_TIMER_AUTODELETE)) {
+      mg_timer_free(head, t);
+      free(t);
+    }
   }
```
scaprile, line 43: One day we should resume our brainstorm... this Timer API has no way to re-use a single-use timer. That could avoid unnecessary memory fragmentation.
OTOH, it has served well for a long time and there are very few, if any, uses of run-once timers in Mongoose code.

scaprile (approved):
Please read comments

Issue #3125, opened 2025-05-03
"mdns-server" crashes on Windows


gvanem:
I wanted to try this new example program `mdns-server.exe` on Windows.
But it crashes since the `bind()` address `224.0.0.251:5353` returns `WSAEADDRNOTAVAIL`
and the sample program ignores a `c == NULL`. That's easy to fix:
```diff
--- a/tutorials/udp/mdns-server/main.c 2025-05-03 13:34:39
+++ b/tutorials/udp/mdns-server/main.c 2025-05-03 14:15:52
@@ -12,6 +12,9 @@

   // Desired name must NOT have any dots in it, nor a domain
   c = mg_mdns_listen(&mgr, "Mongoose");  // Start mDNS server
+  if (!c)
+    return 1;
+
   // if not using our built-in TCP/IP stack, pass the IP address you want to
```

But I fail to understand why Winsock2 won't accept this address in `bind()`.

BTW. what is `-DMG_ENABLE_MDNS=1` for? AFAICS, no code tests this?

scaprile:
> on Windows. But it crashes since the `bind()` address `224.0.0.251:5353` returns `WSAEADDRNOTAVAIL` and the sample program 

Please give us more data, our tests on Win worked fine. There shouldn't be an error returned, hence no test, this is not production code and should be as simple as possible.

> But I fail to understand why Winsock2 won't accept this address in `bind()`.

Exactly, it worked for us
 
> BTW. what is `-DMG_ENABLE_MDNS=1` for? AFAICS, no code tests this?

That's a leftover, initially we didn't default to enabling multicast due to many issues with small embedded systems, later we devised a cooler way but the OS code was not updated. I'll do it in the upcoming weeks, thanks for asking, I completely forgot about it.

gvanem:
> Please give us more data, our tests on Win worked fine. 

Seems both Chrome and Svchost is using this port; `netstat -4nau | grep 5353`:
```
udp   0.0.0.0:5353           -                         NONE         chrome.exe            16708
udp   0.0.0.0:5353           -                         NONE         svchost.exe            2608
udp   0.0.0.0:5353           -                         NONE         chrome.exe            15712
```

Hence this `WSAEADDRNOTAVAIL` in `mdns-server.exe` (but not in Chrome!?). 
`svchost.exe` / PID=2606 is the *DnsCache service* which is part of the *DNS Client* service. I'd rather not disable that.

PR #3128, opened 2025-05-07
Fixed mdns bind issue on windows

robertc2000:
I added a fix for Windows MDNS issue mentioned in #3125.

1. Using the multicast address 224.0.0.251 for Windows triggers WSAEADDRNOTAVAIL (Cannot assign requested address) when attempting to bind. It appears Windows does not recognize this as valid for the local computer. Therefore, I changed the address to 0.0.0.0 for Windows.

2. Typically there are multiple processes listening on 0.0.0.0:5353, which means that using SO_EXCLUSIVEADDRUSE (as the code logic does prior to this PR in sock.c mg_open_listener) will result in another failure when binding (WSAEADDRINUSE: Address already in use). For this case, I think we should use SO_REUSEADDR for UDP sockets on Windows, even if SO_EXCLUSIVEADDRUSE is defined by the SDK. WDYT?

### DIFF
--- a/tutorials/udp/mdns-server/main.c
+++ b/tutorials/udp/mdns-server/main.c
@@ -12,6 +12,7 @@ int main(void) {
 
   // Desired name must NOT have any dots in it, nor a domain
   c = mg_mdns_listen(&mgr, "Mongoose");  // Start mDNS server
+  if (!c) return 1;
   // if not using our built-in TCP/IP stack, pass the IP address you want to
   // use as a response, this depends on your underlying TCP/IP stack and number
   // of interfaces available
--- a/mongoose.c
+++ b/mongoose.c
@@ -461,8 +461,13 @@ static void mdns_cb(struct mg_connection *c, int ev, void *ev_data) {
 
 void mg_multicast_add(struct mg_connection *c, char *ip);
 struct mg_connection *mg_mdns_listen(struct mg_mgr *mgr, char *name) {
+#if MG_ARCH == MG_ARCH_WIN32
+  const char *mcast_url = "udp://0.0.0.0:5353";
+#else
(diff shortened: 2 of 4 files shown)


scaprile:
... I don't know...
What is the actual issue ?
Does it fail because there already is another listener ? Then fail, that's the user's choice.
Does it work when there is no other listener ? Or there's no way to be the only listener ? If it won't work in Windoze, so be it. If it can work when there is no Avahi or equivalent, then great.
I don't like all that scrambling around just for an unlikely usage scenario

tutorials/udp/mdns-server/main.c
```diff
@@ -12,6 +12,7 @@ int main(void) {
 
   // Desired name must NOT have any dots in it, nor a domain
   c = mg_mdns_listen(&mgr, "Mongoose");  // Start mDNS server
+  if (!c) return 1;
```
scaprile, line 15: `c` is not a boolean or bit, so we should check for `== NULL` (it is less non-elegant when checking for !=0 because there is no '!' shouting)

robertc2000:
Yes, it fails because there are other listeners, including critical Windows processes (svchost). So we can't use SO_EXCLUSIVEADDRUSE in this case.

I agree, it is a very unlikely usage scenario.

scaprile:
Agreed, I did try on Linux, in fact I developed the whole mDNS code running on Linux, and IIRC I had Avahi running

robertc2000:
For Windows, I initially tested in a windows docker inside a Linux machine and I had no issues.

But now I decided to try on a Windows host and thus I got to confirm the issue. Inside the docker, I missed it.

scaprile:
Didn't know you can do that. I use VMs, closest to the real thing I know. Looks like Docker managed to fix Windoze...

PR #3132, opened 2025-05-09
stronger bounds checking

scaprile:
Reject Hello Client records with length < 50
Rewrite bounds checking

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -10999,7 +10999,9 @@ static int mg_tls_server_recv_hello(struct mg_connection *c) {
     mg_error(c, "not a client hello packet");
     return -1;
   }
+  if (rio->len < 50) goto fail;
   msgsz = MG_LOAD_BE16(rio->buf + 3);
+  if (((uint32_t) msgsz + 4) > rio->len) goto fail;
   mg_sha256_update(&tls->sha256, rio->buf + 5, msgsz);
   // store client random
   memmove(tls->random, rio->buf + 11, sizeof(tls->random));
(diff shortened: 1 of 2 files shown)


mongoose.c
```diff
@@ -11011,10 +11013,11 @@ static int mg_tls_server_recv_hello(struct mg_connection *c) {
     MG_INFO(("bad session id len"));
   }
   cipher_suites_len = MG_LOAD_BE16(rio->buf + 44 + session_id_len);
-  if (cipher_suites_len > (rio->len - 46 - session_id_len)) goto fail;
```
cpq, line 11014: Is that 46 , and 46 + 4  = 50 be somehow #defined ?

mongoose.c
```diff
@@ -10999,7 +10999,9 @@ static int mg_tls_server_recv_hello(struct mg_connection *c) {
     mg_error(c, "not a client hello packet");
     return -1;
   }
+  if (rio->len < 50) goto fail;
```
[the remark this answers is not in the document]
scaprile: It comes from former 11017 below

mongoose.c
scaprile: All those magic numbers were already there, I just moved them from left to right.
They stop being magic once you have the record structure in your mind, it didn't take me long to get it. That 4 is 2 x BE16, respective length fields being checked

scaprile: 46 is 11 + 32 client random + byte length + BE16 length...
11 is fixed header

scaprile: Not quite different from what you did with DHCP... yes, when other dev does it it sends shivers down one's spine, I know.

scaprile: Being these records very dynamic, with many variable lengths, I don't know if being more academic (less hardcoded) would actually help. Maybe we could add some comments and leave those numbers as they are...

cpq:
Is there a chance to have a failing unit test for that specific thing ? Maybe calling an offending function directly with a crafted buffer? Btw, we can make fuzzer do it.

scaprile:
> Is there a chance to have a failing unit test for that specific thing ? Maybe calling an offending function directly with a crafted buffer? Btw, we can make fuzzer do it.

Pushed a fuzzer, will add it to GH tests later, if you approve it

PR #3134, opened 2025-05-12
OTA for FRDM-RW612

robertc2000:


scaprile (approved):
Pretty similar to iMXRT, I'm tempted to join them...

robertc2000:
Yes, there are strong similarities. The flexspi config structures are the same, I was thinking about creating a separate .h file to contain these and the .c files could share it.

However, there are some differences. The ROM API definition is different to IMXRT and RW612 needs a different `struct mg_flexspi_nor_config` in order to work. Also, RW612 requires to call flash_init at least once, IMXRT does not need to do it.

This is why I thought it will be best if we keep them separate. Merging them will make the code more hard to read and this can also make it more prone to introducing compatibility bugs and harder to track them.

robertc2000:
I'm merging this for now. We can see more about merging the 2 implementations in the near future.

scaprile:
> Merging them will make the code more hard to read and this can also make it more prone to introducing compatibility bugs and harder to track them

That was exactly my trade-off, at first sight couldn't guess whether that would make it more maintainable (single file) or the opposite (too many conditional compilation paths)

PR #3135, opened 2025-05-13
Handle (specific) TCP options

scaprile:
- Add TCP options and announce MSS when sending SYN and SYN ACK (starting and accepting connections)
- Parse TCP options and accept peer announced MSS, honor it and cap TCP data sent, BUT, only when connecting.
  - To be able to get MSS when accepting connections, we need some kind of backlog strategy

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -4178,11 +4178,12 @@ void mg_mgr_init(struct mg_mgr *mgr) {
 
 struct connstate {
   uint32_t seq, ack;           // TCP seq/ack counters
-  uint64_t timer;              // TCP keep-alive / ACK timer
+  uint64_t timer;              // TCP timer (see 'ttype' below)
   uint32_t acked;              // Last ACK-ed number
   size_t unacked;              // Not acked bytes
+  uint16_t dmss;               // destination MSS (from TCP opts)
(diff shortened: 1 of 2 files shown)


mongoose.c
```diff
@@ -4909,13 +4912,33 @@ static void read_conn(struct mg_connection *c, struct pkt *pkt) {
...
+  s->dmss = 536;     // assume default, RFC-9293 3.7.1
+  while (len > 0) {  // RFC-9293 3.1 3.2
+    uint8_t kind = opts[0], optlen = 1;
+    if (kind != 1) {         // No-Operation
+      if (kind == 0) break;  // End of Option List
+      optlen = opts[1];
+      if (kind == 2 && optlen == 4)  // set received MSS
+        s->dmss = (uint16_t) (((uint16_t) opts[2] << 8) + opts[3]);
+    }
+    MG_INFO(("kind: %u, optlen: %u, len: %d\n", kind, optlen, len));
```
cpq, line 4928: MG_DEBUG !

scaprile: Ouch, leftover, that one intended to be a _VERBOSE... sorry about that.

PR #3167, opened 2025-06-10
Add MG_ARCH_TREADX

cpq:


scaprile (changes requested):
We already have an AZURERTOS arch.
Azure RTOS -> Eclipse ThreadX https://threadx.io/
`Microsoft has contributed the Azure RTOS technology to the Eclipse Foundation. With the Eclipse Foundation as its new home, Azure RTOS now becomes Eclipse ThreadX – an advanced embedded development suite including a small but powerful operating system that provides reliable, ultra-fast performance for resource-constrained devices.`

src/http.c
```diff
@@ -977,7 +975,7 @@ long mg_http_upload(struct mg_connection *c, struct mg_http_message *hm,
       mg_http_reply(c, 400, "", "%s: offset mismatch", path);
       res = -5;
     } else if ((fd = mg_fs_open(fs, path, MG_FS_WRITE)) == NULL) {
-      mg_http_reply(c, 400, "", "open(%s): %d", path, errno);
+      mg_http_reply(c, 400, "", "open(%s)", path);
```
scaprile, line 978: same comment for errno, if it survives our own stack, why not do the same with ThreadX ?

src/sock.c
```diff
@@ -160,8 +160,10 @@ static void mg_set_non_blocking_mode(MG_SOCKET_TYPE fd) {
   if (setsockopt(fd, 0, FREERTOS_SO_SNDTIMEO, &off, sizeof(off)) != 0) (void) 0;
 #elif MG_ENABLE_LWIP
   lwip_fcntl(fd, F_SETFL, O_NONBLOCK);
-#elif MG_ARCH == MG_ARCH_AZURERTOS
-  fcntl(fd, F_SETFL, O_NONBLOCK);
+#elif MG_ARCH == MG_ARCH_THREADX
+  // NextDuo fails to send large blocks of data to the non-blocking sockets
```
scaprile, line 164: NetX ?

src/sock.c
```diff
@@ -602,11 +604,11 @@ static void mg_iotest(struct mg_mgr *mgr, int ms) {
     if (c->is_closing) tvp = &tv_zero;
   }
 
-  if ((rc = select((int) maxfd + 1, &rset, &wset, &eset, tvp)) < 0) {
+  if ((rc = select((int) maxfd + 1, &rset, &wset, &eset, tvp)) <= 0) {
```
scaprile, line 607: This affects all other systems, for example, rc = 0 means that Windows will now check maxfd (it did not before), and all will execure FD_ZERO (they did not before).
I would introduce a special case for 0 with MG_ARCH == MG_ARCH_THREADX. wdyt ?

src/sock.c
```diff
@@ -615,7 +617,12 @@ static void mg_iotest(struct mg_mgr *mgr, int ms) {
 
   for (c = mgr->conns; c != NULL; c = c->next) {
     if (FD(c) != MG_INVALID_SOCKET && FD_ISSET(FD(c), &eset)) {
+#if MG_ARCH == MG_ARCH_THREADX
+      // NextDuo stack returns exceptions for listening connection after accept
```
scaprile, line 621: NetX ?

src/net.c
```diff
@@ -184,7 +184,7 @@ struct mg_connection *mg_listen(struct mg_mgr *mgr, const char *url,
   if ((c = mg_alloc_conn(mgr)) == NULL) {
     MG_ERROR(("OOM %s", url));
   } else if (!mg_open_listener(c, url)) {
-    MG_ERROR(("Failed: %s, errno %d", url, errno));
+    MG_ERROR(("Failed: %s", url));
```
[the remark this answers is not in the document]
cpq: referencing errno from here pulls the errno.h which could be an issue on lwip / netxduo, cause they redefine errors
sock.c already has a correct wrapper for errno, and reports that error, so there is no point to report it twice - in fact this whole MG_ERROR could be removed without loosing any useful info

src/arch.h
```diff
@@ -30,7 +30,7 @@
 #endif
 
 #if !defined(MG_ARCH)
-#error "MG_ARCH is not specified and we couldn't guess it. Define MG_ARCH=... in your compiler"
+#error "MG_ARCH is not specified and we couldn't guess it. Define MG_ARCH=... in mongoose_config.h"
```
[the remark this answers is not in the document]
cpq: For OS stuff it is almost sure we have autodetection working
The only time we see this triggered is embedded context, hence the suggestion - it is more practical, tells exactly what to do.

src/http.c
cpq: Same reason as above!

src/sock.c
cpq: Ah, yeah.

scaprile (approved):

PR #3171, opened 2025-06-13
Make logs colorful

LovelyFlowerCat:
Users can redefine their display color in `color_tag` array


scaprile:
Not interested, thanks. Colors are not standard.

PR #3188, opened 2025-06-27
Fix tutorials/http/http-server build with windows gcc

cpq:


scaprile (commented):
You told me to take this over, last week.

Those conditionals have side issues, I just fixed one of those yesterday: FreeRTOS (for example, maybe others), can run over Linux and Windows. We must not refer to an arch as WIN32 but as MG_ARCH_WIN32; then yes, we can address compiler specifics with their signatures, but keeping in mind that the underlying arch may not be Windoze or Linux, or even Mac.
No one would probably use FreeRTOS over Linux in real life, but they do use it for development purposes. I bet ThreadX has something similar.

According to your changes, then it is not the http server that does not run, but anything that uses RSA.
I'm working on the RSA code today, I'll import this PR.

PR #3235, opened 2025-08-12
Refactor HTTP client

cpq:


tutorials/http/http-client/main.c
```diff
@@ -49,24 +50,45 @@ static void fn(struct mg_connection *c, int ev, void *ev_data) {
     // Response is received. Print it
     struct mg_http_message *hm = (struct mg_http_message *) ev_data;
     printf("%.*s", (int) hm->message.len, hm->message.buf);
-    c->is_draining = 1;        // Tell mongoose to close this connection
+    c->is_closing = 1;            // Tell mongoose to close this connection
```
scaprile, line 53: Nope, c->is_draining is correct.
is_closing produces an immediate close, particularly on our built-in stack, where the closure process is not clean and any pending data could be lost.
The proper way described in our tutorials is to use is_draining
is_closing should only use internally or when deliberately closing on error.
Even though it doesn't make a difference here, we should follow our docs

scaprile (changes requested):
I can do the requested change, and cleanup files no longer in use. Let me know

cpq:
> I can do the requested change, and cleanup files no longer in use. Let me know

Yes please, take over this !

scaprile (approved):

PR #3239, opened 2025-08-12
Fix: Use mg_ppp_poll instead of mg_ppp_up in ppp driver

idy:
This PR fixes a build break introduced by https://github.com/cesanta/mongoose/commit/065027843e52b09a8b5bc8429c1f8ed1da7c10ac


scaprile:
Thx, that means we are not testing that for build.

PR #3265, opened 2025-09-08
align mg_addr to 32- and 64-bits

scaprile:
This fixes an act of faith (we assumed 32-bit alignment at one place, and it was not guaranteed except by the struct itself providing that due to element order or compiler good will), and removes several memcpy() calls, simplifying future IPv6 address comparison to two 64-bit comparisons instead of a memcmp() call; same for addr copy.
This may increase mg_conn size due to padding to achieve 64-bit alignment, thing that can be remedied by reordering the struct elements.

### DIFF
--- a/mongoose.h
+++ b/mongoose.h
@@ -1605,7 +1605,11 @@ struct mg_dns {
 };
 
 struct mg_addr {
-  uint8_t ip[16];    // Holds IPv4 or IPv6 address, in network byte order
+  union {    // Holds IPv4 or IPv6 address, in network byte order
+    uint8_t ip[16];
+    uint32_t ip4;
+    uint64_t ip6[2];
+  };
   uint16_t port;     // TCP or UDP port in network byte order
   uint8_t scope_id;  // IPv6 scope ID
   bool is_ip6;       // True when address is IPv6 address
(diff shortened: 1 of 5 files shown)


mongoose.h
```diff
@@ -1605,7 +1605,11 @@ struct mg_dns {
 };
 
 struct mg_addr {
-  uint8_t ip[16];    // Holds IPv4 or IPv6 address, in network byte order
+  union {    // Holds IPv4 or IPv6 address, in network byte order
+    uint8_t ip[16];
+    uint32_t ip4;
+    uint64_t ip6[2];
```
cpq, line 1611: maybe uint32_t[4] ? easier to deal with

cpq (approved):

mongoose.h
scaprile: Yeah, but that would mean 4 32-bit comparisons, it will only remove the 64-bit alignment constraint. Are you thinking on old compilers not being able to handle uint64_t ?

MGPhyo17:
sequenceDiagram
IDKit ->> Bridge: POST /request
Bridge ->> IDKit: <id>
IDKit ->> Bridge: Poll for updates GET /response/:id
WorldApp ->> Bridge: GET /request/:id
Bridge ->> WorldApp: <request>
WorldApp ->> Bridge: PUT /response/:id
IDKit ->> Bridge: Poll for updates GET /response/:id
Bridge ->> IDKit: <response>

PR #3289, opened 2025-09-26
fix use of va_args in mg_queue_vprintf, fixes #3285

### LINKED ISSUE #3285: Crash in mg_queue_printf
Hi, there appears to be a bug in `mg_queue_printf` where certain (valid) format strings can cause a segfault.

Tested on the most recent commit `2c5a6d4d`
OS: Linux/x86
Compiler: clang 17.06

The following testcase demonstrates the issue. Both `%u` and `%s` work fine individually, but for some reason together this crashes inside `mg_queue_printf`. I think it has something to do with reusing the `va_args` list (as it is evaluated twice).
```cpp
#include <cstdio>
#include <cstring>
extern "C" {
#include "mongoose.h"
}

int main() {
	mg_queue q{};
	char buf[100];
	mg_queue_init(&q, buf, sizeof(buf));

	unsigned u = 0;
	const char *s = "x";

	mg_queue_printf(&q, "u=%u s=%s", u, s);

	return 0;
}
```

ASAN report:
```
==12==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000f08 (pc 0x7f0d48be60bc bp 0x7ffec9c1adb0 sp 0x7ffec9c1a568 T0)
==12==The signal is caused by a READ memory access.
==12==Hint: address points to the zero page.
    #0 0x7f0d48be60bc  string/../sysdeps/x86_64/multiarch/strlen-evex.S:77
    #1 0x5607e5ed01fa in strlen (/fuzz/workspace/test+0x4a1fa) (BuildId: 375464416b5ee6eb57987e2235b5079878d514fc)
    #2 0x5607e5ff4db3 in mg_vxprintf.363 /fuzz/src/mongoose.c:801:45
    #3 0x5607e5fa3c5d in mg_vsnprintf /fuzz/src/mongoose.c:7855:14
    #4 0x5607e5fa3c5d in mg_snprintf /fuzz/src/mongoose.c:7864:7
    #5 0x5607e5fcc883 in mg_queue_vprintf /fuzz/src/mongoose.c:7814:16
    #6 0x5607e5fcd4d3 in mg_queue_printf /fuzz/src/mongoose.c:7829:9
    #7 0x5607e5f9562e in main /fuzz/workspace/test.cpp:15:2
    #8 0x7f0d48a5dd8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #9 0x7f0d48a5de3f in __libc_start_main csu/../csu/libc-start.c:392:3
    #10 0x5607e5eba404 in _start (/fuzz/workspace/test+0x34404) (BuildId: 375464416b5ee6eb57987e2235b5079878d514fc)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV string/../sysdeps/x86_64/multiarch/strlen-evex.S:77
```

hgarrereyn:
Implements spec compliant handling of va_args in `mg_queue_vprintf`.

See comments in #3285 for more details

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -7812,7 +7812,9 @@ bool mg_ota_end(void) {
 
 
 size_t mg_queue_vprintf(struct mg_queue *q, const char *fmt, va_list *ap) {
-  size_t len = mg_snprintf(NULL, 0, fmt, ap);
+  va_list ap_copy;
+  va_copy(ap_copy, *ap);
+  size_t len = mg_vsnprintf(NULL, 0, fmt, &ap_copy);
   char *buf;
   if (len == 0 || mg_queue_book(q, &buf, len + 1) < len + 1) {
     len = 0;  // Nah. Not enough space
--- a/src/printf.c
+++ b/src/printf.c
@@ -3,7 +3,9 @@
 #include "util.h"
 
 size_t mg_queue_vprintf(struct mg_queue *q, const char *fmt, va_list *ap) {
-  size_t len = mg_snprintf(NULL, 0, fmt, ap);
+  va_list ap_copy;
+  va_copy(ap_copy, *ap);
+  size_t len = mg_vsnprintf(NULL, 0, fmt, &ap_copy);
   char *buf;
   if (len == 0 || mg_queue_book(q, &buf, len + 1) < len + 1) {
     len = 0;  // Nah. Not enough space


scaprile:
@cpq We pass va lists a lot, please take a look (https://github.com/cesanta/mongoose/issues/3285#issuecomment-3339963961)
@hgarrereyn Thank you ! (We'll ask you to sign the CLA if we go your way)

scaprile:
@hgarrereyn Can you please point to authoritative C documentation on this ? Remember Mongoose is single-threaded. Thank you.

hgarrereyn:
Hi @scaprile, the C spec is a bit confusing on this point.

In ISO/IEC 9899:2024 section 7.16.4 says:
> The object ap may be passed as an argument to another function; if that function invokes the va_arg macro with parameter ap, the representation of ap in the calling function is indeterminate and shall be passed to the va_end macro prior to any further reference to ap.

yet has a footnote saying:

> A pointer to a va_list can be created and passed to another function, in which case the original function can make further use of the original list after the other function returns.

This seems to indicate that the `va_list` does not get corrupted as long as you pass by reference and not by value -- but experimenting with this it seems that this is not always followed (e.g. clang-17 seems to corrupt in both cases).

---

However, the use of the following kind of construction is erroneous (i.e. used in `mg_queue_vprintf`):

```c
void foo(int x, ...) {
  va_list ap;
  va_start(x, ap);
  // ...
  va_end(x);
}

void bar(int x, ...) {
  va_list ap;
  va_start(x, ap);
  foo(x, &ap); 
  va_end(x);
}
```

The resulting `ap` in `foo` is _not_ equivalent to the `ap` in `bar`, it is a _new_ va_list containing the pointer to bar's `ap`.

So as a result, inside the first iteration of `mg_snprintf`, its actually iterating over a completely new `va_list` with different arguments. This is why you don't see corruption in the second iteration -- because it was never actually invoking `va_arg` on the original va_list in the first place.

I added a bunch of debug statements inside mongoose to illustrate that it is actually iterating over different values. (as a result, it is actually computing the target buffer size incorrectly, in most cases over-estimating, but it is possible to underestimate as well).

Consider the following testcase running on `acd65f2` (with added print statements).

**testcase.c**
```c
#include "mongoose.h"

int main() {
	struct mg_queue q;
	char buf[54];
	mg_queue_init(&q, buf, sizeof(buf));

	void *p = 0xffffffffffffffff;

	mg_queue_printf(&q, "A%p%p%pB", p, p, p);
	mg_hexdump(buf,52);

	return 0;
}
```

**output**
```
[+] mg_queue_printf
[+] mg_snprintf (computing length)
[+] mg_vsnprintf - io {buf=(nil), size=0, len=0, align=0}
tmp=fffffcc95160��
tmp=aaaad8e8d643��
tmp=ffff9ec037a3��
computed n=44
[+] mg_queue_book - len=45
[+] mg_queue_book - enough space
[+] mg_vsnprintf (filling buffer)
[+] mg_vsnprintf - io {buf=0xfffffcc95344, size=45, len=0, align=0}
tmp=ffffffffffffffff
tmp=ffffffffffffffff
tmp=ffffffffffffffff
computed n=56
0000   38 00 00 00 41 30 78 66 66 66 66 66 66 66 66 66   8...A0xfffffffff
0010   66 66 66 66 66 66 66 30 78 66 66 66 66 66 66 66   fffffff0xfffffff
0020   66 66 66 66 66 66 66 66 66 30 78 66 66 66 66 66   fffffffff0xfffff
0030   00 00 00 00                                       ....
```

In the first invocation to `mg_snprintf` (to compute the length), it passes the `va_list` incorrectly, thus the three pointer values it dereferences to compute the size are all random stack values -- likely pointers so they are fewer than 64-bits. Thus the resulting size is estimated to be 44.

There is enough space in the buffer for this, so it continues and invokes `mg_vsnprintf` which constructs an internal `mg_iobuf` of only 45 chars (the estimated size).

Since `mg_vsnprintf` _does_ take a `va_list *` directly, it is actually operating on the right arguments and we see the original pointers passed in. However, these are larger than expected and as such the resulting output is truncated. Luckily, there is a check on the buffer size so it doesn't overflow -- it just gets truncated.

---

This patch fixes the issue by both making a copy of the `va_list` and invoking `mg_vsnprintf` in the first iteration (instead of `mg_snprintf`) so that both iterations process the correct arguments.

scaprile:
Thank you!
I will check this on Monday, our code is C89 compatible (yes, C89), though we venture into C99 where we know it fits, like in Ethernet drivers. We do not recommend building on C++ as other than C.
IIRC, we do a lot of variadic args passing in all the printf functions, so I'd better check this.

scaprile:
With these mods (final text is larger than your buffer and printing fails):
```
#include "mongoose.h"

int main() {
	struct mg_queue q;
	char buf[128];
	mg_queue_init(&q, buf, sizeof(buf));

	void *p = (void *) 0xffffffffffffffff;

	size_t len = mg_queue_printf(&q, "A%p%p%pB", p, p, p);
	mg_hexdump(buf, len);

	return 0;
}
```
and adding "NEW" to your patch:
```
size_t mg_queue_vprintf(struct mg_queue *q, const char *fmt, va_list *ap) {
  va_list ap_copy;
  va_copy(ap_copy, *ap);
  MG_INFO(("NEW"));
  size_t len = mg_snprintf(NULL, 0, fmt, &ap_copy);
  char *buf;
```
I still get a faulty output with or without it:
```
$ vi mongoose.c
$ gcc -o q q.c mongoose.c 
$ ./q
5d6c9533 2 mongoose.c:7817:mg_queue_vpr NEW
0000   38 00 00 00 41 30 78 66 66 66 66 66 66 66 66 66   8...A0xfffffffff
0010   66 66 66 66 66 66 66 30 78 66 66 66 66 66 66 66   fffffff0xfffffff
0020   66 66 66 66 66 66 00 00 00 00 c0 00 00 00 00 00   ffffff..........
0030   ff ff ff ff ff ff ff ff                           ........        
$ cp mongoose.c pp
$ git restore mongoose.c 
$ gcc -o q q.c mongoose.c 
$ ./q
0000   38 00 00 00 41 30 78 66 66 66 66 66 66 66 66 66   8...A0xfffffffff
0010   66 66 66 66 66 66 66 30 78 66 66 66 66 66 66 66   fffffff0xfffffff
0020   66 66 66 66 66 66 66 66 66 30 78 66 66 66 66 66   fffffffff0xfffff
0030   66 66 66 66 00 ff ff ff                           ffff....        
```

Looks like we're on the right track, but, as said before, we do this liberally so maybe there are more issues down the rabbit hole.
I'll take a deeper look at this as soon as I can.

hgarrereyn:
It looks like you are still calling `mg_snprintf` instead of `mg_vsnprintf` in that patch.

scaprile:
Oh, I see ! There's two issues in here, the call to `mg_snprintf()` is not correct !
That wrong call obscured the ap_list issue but was giving a wrong length; fixing that I can see pointers have been changed and your patch fixing that.

Please sign the CLA if you haven't, and let me know https://cesanta.com/cla.html

hgarrereyn:
Just signed!

PR #3303, opened 2025-10-09
fix: Add Poly1305 MAC verification to ChaCha20-Poly1305 decrypt

### LINKED ISSUE #2918: Lack of poly1305 certification in chacha20-poly1305 BUILT_IN implementation
Hi! I found that in the BUILT_IN implementation of chacha20-poly1305, there is no operation for poly1305 authentication; instead, it directly uses the chacha20 algorithm for decryption. This could lead to a failure to detect if the ciphertext has been tampered with by an attacker.

https://github.com/cesanta/mongoose/blob/c00962e881093cf676588df596649704e4ff8e36/src/tls_chacha20.c#L1327

Deenaadayalan:
Fixes #2918

   ## Problem
   The `mg_chacha20_poly1305_decrypt` function was missing Poly1305 MAC verification, making it vulnerable to tampering attacks. ChaCha20-Poly1305 is an authenticated encryption algorithm (AEAD), but the decrypt function only performed decryption without verifying the authentication tag.

   ## Solution
   Added proper MAC verification before decryption:

   1. **Extract MAC tag** from the end of ciphertext
   2. **Compute expected MAC** using `poly1305_calculate_mac` 
   3. **Constant-time comparison** to prevent timing attacks
   4. **Fail-safe behavior**: Return error WITHOUT decrypting if MAC is invalid
   5. **Decrypt only after verification** succeeds

   ## Security Impact
   - ✅ Prevents tampering attacks on encrypted data
   - ✅ Implements full AEAD as per RFC 8439
   - ✅ Uses constant-time comparison to prevent timing side-channels
   - ✅ Matches the encrypt function's MAC generation logic

   ## Testing
   The fix has been verified to:
   - Use the same `poly1305_calculate_mac` function as encryption
   - Verify MAC before decryption (correct order of operations)
   - Use constant-time comparison loop
   - Return error before any decryption on MAC mismatch

   ## Changes
   - Modified `mg_chacha20_poly1305_decrypt` in `src/tls_chacha20.c`
   - Added 21 lines implementing MAC verification
   - No breaking changes to API

### DIFF
--- a/src/tls_chacha20.c
+++ b/src/tls_chacha20.c
@@ -1335,7 +1335,26 @@ PORTABLE_8439_DECL size_t mg_chacha20_poly1305_decrypt(
   if (MG_OVERLAPPING(plain_text, actual_size, cipher_text, cipher_text_size)) {
     return (size_t) -1;
   }
-
+  
+  // SECURITY FIX: Verify MAC before decrypting
+  uint8_t computed_mac[RFC_8439_TAG_SIZE];
+  const uint8_t *provided_mac = cipher_text + actual_size;
+  
+  // Compute expected MAC
+  poly1305_calculate_mac(computed_mac, cipher_text, actual_size, key, nonce, NULL, 0);
+  
+  // Constant-time comparison to prevent timing attacks
+  unsigned char result = 0;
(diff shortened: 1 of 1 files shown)


scaprile:
Thank you, this is not in our style but we'll add the idea and link to your submission.

PR #3304, opened 2025-10-09
uniformize memory attributes

scaprile:


scaprile:
@robertc2000 in the XMC7 driver, the static indexes, thing that we ourselves manually handle in the IRQ and TX functions, are declared as requiring alignment. This doesn't make sense to me, can you please check this ? (I don't have such a board)

robertc2000:
> @robertc2000 in the XMC7 driver, the static indexes, thing that we ourselves manually handle in the IRQ and TX functions, are declared as requiring alignment. This doesn't make sense to me, can you please check this ? (I don't have such a board)

Yes, it is a leftover. I tested now just to make sure. You can remove the alignment.

robertc2000:
Shouldn't we do the same for the rest of the drivers? I notice the changes have not been applied to some of them (ra, same54, xmc, tm4c).

scaprile:
> Shouldn't we do the same for the rest of the drivers? I notice the changes have not been applied to some of them (ra, same54, xmc, tm4c).

Yeah, I only changed those that actually had an alignment directive, but the memory section should be applied where there is a placement requirement, and can be applied everywhere as long as we leave that section as yet one more where it is not special.
I'd rather leave it alone when it is not needed, so we know there is no placement requirement. Can you hint me on SAME and XMC ? I'm friends with TM4C and RA is an acquaintance.

robertc2000:
Ok, then we can leave it unmarked when there is no alignment needed.

The only change should then be for XMC. Currently it defines `ETH_RAM_SECTION` macro instead of `MG_ETH_RAM`.. I will then take care to update the XMC examples to use `MG_ETH_RAM` instead.

scaprile:
> The only change should then be for XMC. Currently it defines `ETH_RAM_SECTION` macro instead of `MG_ETH_RAM`.. I will then take care to update the XMC examples to use `MG_ETH_RAM` instead.

Done, already took care of the examples, let's see when this gets deployed

PR #3322, opened 2025-10-27
fix error logs not uploading on test failure

scaprile:
@robertc2000 This fix I added to a branch that suddenly bumped into a wall due to strange behavior in a vendor library. I switched to other tasks and it was never patched. Added now. Let me know if ALL tests upload now, no matter they pass or fail.
Logs should only fail to upload when that test is cancelled, which is correct, as the test did not actually run to completion


robertc2000:
Yes, all the tests upload now. This is very good, there are no more yellow tests for today.

Now that all tests are successfully uploading, I don't think there is a strong need anymore for checking when was the last status for a given test. However, if this issue will reoccur in the future, the table cell should display the yellow sign and we should check more thoroughly the tests that fail to upload.

scaprile:
@robertc2000 Yes, indeed, I do agree. A fail to upload is either GH cancelling the test for some reason (sometimes they run out of resources and do that, sometimes downloading a Docker image fails just out of thin air...) or a bug of ours in the test script itself.
(or a networking problem when scp-ing)

PR #3325, opened 2025-10-27
Handle OOM situations

scaprile:


scaprile:
> There is no change to mg_vprintf() in net.c - is that intentional ?

That's already been merged after testing, since you suggested it I assumed you'd approve it.

https://github.com/cesanta/mongoose/blob/b3b73f4751a2dc19921959e0545004c0dc3cba4b/src/net.c#L10-L20

cpq:
@scaprile should we kill this PR ?

scaprile:
> @scaprile should we kill this PR ?

@cpq No, this PR handles everything else not printing, particularly lots of actions in TLS. Without this PR, we may send malformed TLS records or MQTT messages or even HTTP requests/responses under OOM conditions.
It is waiting for your review, it has been done in two stages.

PR #3382, opened 2025-12-22
Make mcast/bcast consistent across stacks

scaprile:
When working with UDP multicast/broadcast, we can send to a multicast/broadcast address and get unicast or multicast/broadcast responses.
When getting a response, `c->rem` gets filled with the respondent IP data.
To be able to send multicast/broadcast again, we need to copy back to `c->rem` the original IP information.
- `mg_connect`: we instruct to save `c->rem` contents somewhere (e.g.: `c->data`), and then copy back (SSDP tutorial)
- `mg_listen`: we can copy back from `c->loc`, that's what our mDNS code does

With net_builtin, the same happens with the l2addr (MAC); if we don't restore it, we end up sending to the multicast/broadcast IP but to the requester MAC (mDNS, this just works but is wrong), or last respondent MAC (SSDP, blatantly wrong).
The fix is simple, but, in this case, it is not necessary to save the MAC, as it can be mapped (again) from the multicast/broadcast IP.

So, to keep consistency across stacks, we add `mg_multicast_restore(c, from)`. This is the function to be called in both cases. Internally, this function copies `from` --> `c->rem` and for net_builtin also maps the MAC again.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -625,7 +625,7 @@ static void handle_mdns_record(struct mg_connection *c) {
       p = build_name(respname, p);
       p = build_a_record(c, p);
     }
-    if (!req.is_unicast) memcpy(&c->rem, &c->loc, sizeof(c->rem));
+    if (!req.is_unicast) mg_multicast_restore(c, (uint8_t *) &c->loc);
     mg_send(c, buf, (size_t) (p - buf));  // And send it!
     MG_DEBUG(("mDNS %c response sent", req.is_unicast ? 'U' : 'M'));
   }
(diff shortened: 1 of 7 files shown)


cpq:
So we have a new API function mg_multicast_restore and the only thing it does, it copies into c->rem?
I have an issue with that

scaprile:
> and for net_builtin also maps the MAC again

If we don't do something like that, then, users have to either save or re-generate the MAC themselves, but only for net_builtin. Then, multicast/broadcast user code will be different when using net_builtin.
`s->mac` gets copied from `pkt->l2->src` when receiving something.
This replicates for the link the same strategy we use for IP.

The other way I can think of is to check on every send, it is way less efficient as we're penalizing regular unicast UDP traffic, AND, the "hack" is still needed at the IP level.

cpq:
What is the anticipated usage of mg_multicast_restore() ? Could you drop a little snippet that demonstrates it please?

scaprile:
> What is the anticipated usage of mg_multicast_restore() ? Could you drop a little snippet that demonstrates it please?

https://github.com/cesanta/mongoose/blob/9280110702cf5adc34c47385c7f3ec20c4a61b76/tutorials/udp/ssdp-search/main.c#L31-L35
https://github.com/cesanta/mongoose/blob/9280110702cf5adc34c47385c7f3ec20c4a61b76/src/dns.c#L507

cpq:
Ok LGTM
Can I be picky again?
I suggest to rename `c->is_unicast` to `c->is_multicast` and reverse the condition where it is used
Please merge.

scaprile:
> I suggest to rename `c->is_unicast` to `c->is_multicast` and reverse the condition where it is used

Well, it is not c-> but the mDNS request struct, which is by definition multicast (hence the 'm'). The non-usual case is when the request is explicit to perform a unicast response instead of the usual standard multicast response; so the name and usage.

Issue #3278, opened 2025-09-19
Sending HTTPS request has memory leak


general252:
- My goal is: Sending HTTPS request has memory leak
- My actions were: Send HTTPS request
- My expectation was: memory leak
- The result I saw: _provide logs or description_
- My question is: `mg_tls_free` need `mbedtls_psa_crypto_free()`

# Environment

- mongoose version:  7.19 
- Compiler/IDE and SDK: gcc version 11.4.0
- Target hardware/board: PC
- Connectivity chip/module: Intel
- Target RTOS/OS (if applicable): Ubuntu 22.04.3 LTS
- mbedtls version: 3.6.4

---
`mg_tls_free` need `mbedtls_psa_crypto_free()`

https://github.com/cesanta/mongoose/blob/fe6d35785845297de203ed93951914e0936bd2cd/mongoose.c#L14288-L14302

```C
void mg_tls_free(struct mg_connection *c) {
  struct mg_tls *tls = (struct mg_tls *) c->tls;
  if (tls != NULL) {
#if defined(MBEDTLS_VERSION_NUMBER) && MBEDTLS_VERSION_NUMBER >= 0x03000000 && \
    defined(MBEDTLS_PSA_CRYPTO_C)
      mbedtls_psa_crypto_free();
#endif

    mbedtls_ssl_free(&tls->ssl);
    mbedtls_pk_free(&tls->pk);
    mbedtls_x509_crt_free(&tls->ca);
    mbedtls_x509_crt_free(&tls->cert);
    mbedtls_ssl_config_free(&tls->conf);
#ifdef MBEDTLS_SSL_SESSION_TICKETS
    mbedtls_ssl_ticket_free(&tls->ticket);
#endif
    mg_free(tls);
    c->tls = NULL;
  }
}
```

test code
```C
#include "mongoose.h"
#include "mbedtls/build_info.h"

static const char* s_url = "https://bing.com";

static void fn(struct mg_connection* c, int ev, void* ev_data) {
     if (ev == MG_EV_CONNECT) {
        // Connected to server. Extract host name from URL
        struct mg_str host = mg_url_host(s_url);

        if (c->is_tls) {
            struct mg_tls_opts opts = { 0 };
            opts.skip_verification = 1;
            mg_tls_init(c, &opts);
        }

        // Send request
        int content_length = 0;
        mg_printf(c,
            "GET %s HTTP/1.0\r\n"
            "Host: %.*s\r\n"
[...]

PR #3386, opened 2025-12-25
Mbedtls nits

cpq:


test/unit_test.c
```diff
@@ -4141,6 +4141,14 @@ int main(void) {
   if (s_abort != 0) return EXIT_FAILURE;
 #endif
 
+
+#if 0
```
scaprile: This is incorrect, we have to call mbedtls_psa_crypto_free(). Tests are not failing because library versions involved do not require that function call. I explicitly added it in #3281 after investigating a report in #3278 with a specific build of the library
Unfortunately, I didn't realize that was using global data...

PR #3404, opened 2026-01-12
Allow comments in PEM

cpq:
Our TLS code checks for '-' as the first character to check if the cert / key is PEM or DER.
However PEM allows for arbitrary comments before the ------ BEGIN sections,
which makes our code to falsely assume that the cert is DER:

```c
               : s.buf[0] == '-'
                   ? PEM_read_bio_X509(bio, NULL, NULL, NULL)  // PEM
                   : d2i_X509_bio(bio, NULL);                  // DER
```

Instread, reverse the checks and tests for DER first, which has its first byte always '0'.

### DIFF
--- a/src/tls.h
+++ b/src/tls.h
@@ -34,7 +34,7 @@ void mg_tls_handshake(struct mg_connection *);
 // Private
 void mg_tls_ctx_init(struct mg_mgr *);
 void mg_tls_ctx_free(struct mg_mgr *);
-#define MG_IS_DER(buf) ((uint8_t) (buf)[0] == 0x30)  // DER begins with 0x30
+#define MG_IS_DER(buf) (((uint8_t *) (buf))[0] == 0x30)  // DER begins with 0x30
 
 // Low-level IO primives used by TLS layer
 enum { MG_IO_ERR = -1, MG_IO_WAIT = -2, MG_IO_RESET = -3 };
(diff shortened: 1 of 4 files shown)


src/tls.h
```diff
@@ -34,6 +34,7 @@ void mg_tls_handshake(struct mg_connection *);
 // Private
 void mg_tls_ctx_init(struct mg_mgr *);
 void mg_tls_ctx_free(struct mg_mgr *);
+#define MG_IS_DER(buf) ((uint8_t) (buf)[0] == 0x30)  // DER begins with 0x30
```
scaprile:
```
In file included from mongoose.c:20:
src/tls_mbed.c: In function ‘mg_load_cert’:
mongoose.h:1885:40: error: subscripted value is neither array nor pointer nor vector
 1885 | #define MG_IS_DER(buf) ((uint8_t) (buf)[0] == 0x30)  // DER begins with 0x30
      |                                        ^
src/tls_mbed.c:36:8: note: in expansion of macro ‘MG_IS_DER’
```

cpq:
Your updates look great Sergio!
Please merge if you're ok with the PR.

scaprile:
> Please merge if you're ok with the PR.

There are some test failures I'll check other day, more likely are not related; builds OK

PR #3424, opened 2026-01-28
Read out all pending TLS data

cpq:


scaprile:
Former approach was:

```
mg_tls_recv
    mg_tls_recv_record
        loop
            do we have a full record ?
                NO: return WAIT
            is it a known record ?
                NO: return ERROR
            is it an APP record ?
                NO: read it and drop it, loop
                YES: exit loop
        handle APP record
    do we have an APP record to process ?
        NO: return WAIT
    is it TLS_APP_DATA ?
        NO: drop it, return WAIT
    do we have a buffer to store data ?
        NO: it will be created after this call, now we know there is data; return 0
    move data, return data length
```

Problem: When the server sends more than 1 record (several non-data records and a data record) in the same segment and immediately closes the connection, we have an IO_ERROR condition from the socket and a WAIT condition from TLS, even though there is data in the TLS buffer (our TLS processes the non-data APP record and returns IOWAIT, leaving the data APP record for next poll. This is a closure indication, so we close (and there is no next poll, except in tls_builtin that waits for FIN timeout)

New approach:

```
mg_tls_recv
    loop
        mg_tls_recv_record
        do we have an APP record to process ?
            NO: return WAIT
        is it TLS_APP_DATA ?            
            NO: drop it, loop
            YES: exit loop
    do we have a buffer to store this data ?
        NO: it will be created after this call, now we know there is data; return 0
    move data, return data length
```

PR #3431, opened 2026-02-05
fix RA6M4

scaprile:
The RA6 driver had an incorrect internal setting, so it was causing the phy code to catch the NXP settings.
NXP boards have the same PHY chip, but their MAC clocks the PHY at 50MHz
The EK-RA6M4 board has a 25MHz crystal and the PHY clocks the MAC.
The EK-RA8M1 board has a different PHY, and who clocks who is ambiguous, both PHY and MAC have an external clock. As the PHY does not require special settings, this change should be safe.

Please test:
EK-RA6M4
EK-RA8M1 (just in case)

Please remember to update Wizard code once this PR is merged.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -24735,7 +24735,7 @@ bool __attribute__((weak)) mg_wifi_ap_stop(void) {
 
 
 enum {                  // ID1  ID2
-  MG_PHY_KSZ8x = 0x22,  // 0022 1561 - KSZ8081RNB
+  MG_PHY_KSZ8x = 0x22,  // 0022 156x - KSZ8081RNB, KSZ8091RNB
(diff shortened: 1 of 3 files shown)


scaprile:
> (though it appears the MAC clocks the PHY in the schematic)

To me, who clocks who is ambiguous, both PHY and MAC have an external clock. I don't see a REFCLK from the PHY going into the MAC, nor viceversa... If I had to take a side, as this is a PHY setting, probably I would agree with you. Hopefully we're not setting clocks on this PHY, and that is consistent with the generic PHY clocks MAC setting. Should've been a different MAC, maybe we'd need to introduce the config parameter into the driver data structure.
So far, it is simpler this way.

PR #3435, opened 2026-02-09
Improve mqtt props bound checks

robertc2000:


cpq:
LGTM

scaprile (approved):
Oh, yes.. damned copy-paste... I forgot to trim the check for when there was no more data after that.
Yes, zeroing the prop struct matches its documentation, you're right.

PR #3441, opened 2026-02-16
Update http-server to load certs from files

cpq:


cpq:
@scaprile the reason is that ss_server.pem contains both the cert and private key.
The private key is in the "BEGIN PRIVATE KEY" section
However our builtin TLS code expects "BEGIN EC PRIVATE KEY" section

The short answer - currently our builtin TLS code cannot parse ss_server.pem certs.
I suggest to handle that in a separate PR which fixes our parsing and converts this repo to ss_server.pem

scaprile:
Hmmm.... can we keep this example as a simple one and manage that new stuff on a separate example ? I mean, this is the main entry point for all examples, I think it should be easy and standard. We can have a cert_chain tutorial... WDYT @cpq ?
If you don't agree, just go ahead.

cpq:
Several examples for the http server is an overkill IMO
Why? Certs can be chained or not, it does not require a separate example

cpq:
OK, I'll merge this one and then make a separate PR to adopt for ss_* pem files.

PR #3449, opened 2026-02-24
Obtain IP adress for MDNS automatically

cpq:


cpq:
All your comments are valid.

But our current way just replaced the work on the customer's shoulders.
Noone wants to find their IP address in manually and inject that into the code in a weird way, through c->data.
That just does not feel right. Mongoose should do the job, not our user.

Of course, if there multiple interfaces, or no Google, or whatever - we could still have a fallback to manual setup.
But by default, Mongoose should just do it, and customer should not care.

To enable MDNS, they only need to setup the local name, and that's it.

So let's make that happen - I leave it in your capable hands!
Feel free to modify this PR the way you regard the best.

scaprile (approved):

scaprile:
> So let's make that happen - I leave it in your capable hands! Feel free to modify this PR the way you regard the best.

Addressing all my concerns, and yours:
- Address discovery is performed using the multicast address. This has no side effects and returns the proper address.
- The discovery function lies in sock.c, taking care of different TCP/IP stack differences and reusing what can be reused.
- c->loc is no longer used
- Should a different IP need to be used, an advanced example is provided, catching the request in an event handler and returning an ad hoc address. The mg_addr for that was already there, changed it to a pointer to detect mods to it.

PR #3454, opened 2026-02-26
C++ designated initializer fix

robertc2000:


robertc2000:
I thought about that one, but it would be dangerous in case we happen to make changes to the `mg_addr` struct. We would likely forget to update these two..

scaprile:
If we make changes, something will be out of place and initialization would fail. If we go with binary stuff, only length matters. 
I mean:
```c
static const struct mg_addr ip6_allrouters = {
    {0xFF, 0x02, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0x02},
    0,  0,  true
};
```
I think my problem when writing this the first time was that the first element is a union. Don't remember. Too much stuff going on at the same time and my only neuron collapses...
The above is clear

scaprile:
Now I recall... something... that Diab compiler also has a problem when initializing unions, it is not even C89 at that.
Don't worry, now I know what the issue is I'll see what convinces me the most.

PR #3458, opened 2026-03-05
Introduce new mDNS API, add mDNS client example

scaprile:


gvanem:
On Windows, the **mdns-client** suffers from the same `bind()` issue we've seen before:
```c
  bind (s, 224.0.0.251:5353) --> WSAEADDRNOTAVAIL (10049)
```

similar to this https://github.com/CCExtractor/ccextractor/issues/668

A fix would be more complicated: `bind (s, INADDR_ANY)`. And join *Multicast Group* with
`setsockopt (IP_ADD_MEMBERSHIP, 224.0.0.251)`.

I'm just saying.

PR #3466, opened 2026-03-13
Add L2 PPP and PPPoE

scaprile:
This PR:
- Splits the former PPP driver into two pieces of code
  - The AT command handling and serial port byte stuffing section is refactored to do stuffing/unstuffing on the fly, as bytes are sent and received; this is moved to a new driver: *ATcmd*, at `atcmd.{c,h}`- This driver also sends the HDLC idle (start/stop) flags. This can be easily configured out in the future, byte stuffing/unstuffing has a configuration switch to disable them now.
  - The actual PPP and HDLC section is refactored to work at the frame level (instead of byte level), and moved to an L2 implementation: *L2_PPP*, at `l2_ppp.{c,h}`. There, reusable portions are moved to internal functions and support for PPPoE is added.
- Removes the former `ppp.{c,h}` driver
- Adds two new `l2type`s: `MG_TCPIP_L2_PPP` and `MG_TCPIP_L2_PPPoE`
- Modified the SIM800 example to use the ATcmd driver and configure the PPP l2type
- Adds an STM32F746 example connecting through PPPoE
- Slightly modifies the L2 internal API, and adds an extra poll() function call. This is because some L2 need to do things on their own, independently of higher level or received traffic. PPPoE is perhaps the simplest, having to initiate a session to a server before any PPP traffic gets transferred and before any IP datagrams can be interchanged.
  - PPP still works as a responsive peer, reacting to server requests; PPPoE clients must initiate a session, sort of like DHCP works, so that basic idea was copied here for that.
- Adds an extra state to the mif interface, to separate `driver_ready` from `l2_ready`. Any L2 function can then know if its physical link or underlying layer is ready or not.
  - This is used to start sending PPPoE PADI frames after the PHY negotiates with the switch and brings the link up; and to tear down an established session if the Ethernet link gets disconnected.
  - This, along with the poll() function, can be used for other future L2 implementations that require negotiating link protocols or discovering networks (Thread, for example)

NOTE: statics should move into a struct mg_l2_ppp_data, referenced through void * ifp->l2data


scaprile:
Let's not merge this until 7.21 is out; --> draft

PR #3500, opened 2026-04-08
Add OTA url check

robertc2000:


src/ota.c
```diff
@@ -0,0 +1,203 @@
+#include "ota.h"
+#include "http.h"
+#include "log.h"
+
+#if MG_OTA != MG_OTA_NONE
+
+enum { STATUS_WAITING, STATUS_SUCCESS, STATUS_FAIL };
```
cpq, line 7: Very generic enums, they may clash with something else in the amalgamated code - I suggest to prefix with MG_OTA_

src/ota.c
```diff
@@ -0,0 +1,203 @@
...
+static int s_version_status;
+static int s_ota_status;
+static uint64_t s_start_time;
+
+static struct mg_ota_metadata {
+  char *version;
+  char *url;
+  size_t size;
+  uint32_t crc;
```
cpq, line 16: I don't think anyone is crc-ing the firmware. they either sha1 or sha-256, so maybe we could use one of those? Even md5 may be better than crc

cpq (approved):
LGTM!

src/ota.c
robertc2000: Yeah we could use sha-256, though at the moment we don't use the crc field to validate the firmware, we just read it and do nothing else. I was thinking we could add support for validating it later, since it wasn't the main priority. What do you think, @cpq? Shall we leave sha-256 for later?

cpq: Yeah sure we could leave the validation for later

PR #3512, opened 2026-04-23
fix: trigger TLS handshake on synchronous connect success

thinker132203:
Summary
This PR fixes a critical bug where the TLS handshake is never initiated if the TCP connection succeeds synchronously (e.g., connect() returns 0 immediately).
Background
This issue was identified while porting Mongoose to the Taixin 82x series (TS826/TS828) embedded platform. On this specific hardware/SDK, the network stack often completes the TCP handshake immediately for local or certain network conditions, causing connect() to return 0 instead of EINPROGRESS.
Problem
In mg_connect_resolved(), when a connection succeeds immediately, the MG_EV_CONNECT event is triggered. If the user calls mg_tls_init() inside this event, the is_tls_hs flag is correctly set to 1.
However, unlike the asynchronous path in connect_conn(), the current implementation of mg_connect_resolved() fails to check this flag or invoke mg_tls_handshake(). Consequently:
1. The client never sends the initial ClientHello packet.
2. The event loop remains idle, waiting for a read event from the server that will never occur.
3. The connection eventually hangs and leads to a timeout.
Solution
I have updated mg_connect_resolved() to include a check for c->is_tls_hs within the rc == 0 block, mirroring the established logic in connect_conn(). This ensures the first TLS handshake packet is dispatched immediately after a successful synchronous connection.
Impact
This fix ensures reliable TLS establishment across all platforms, especially on embedded systems where synchronous connection completion is a common occurrence.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -12027,6 +12027,7 @@ void mg_connect_resolved(struct mg_connection *c) {
     if (rc == 0) {                       // Success
       setlocaddr(FD(c), &c->loc);
       mg_call(c, MG_EV_CONNECT, NULL);   // Send MG_EV_CONNECT to the user
+      if (c->is_tls_hs) mg_tls_handshake(c);  // Trigger TLS handshake immediately if user called mg_tls_init()
       if (!c->is_tls_hs) c->is_tls = 0;  // user did not call mg_tls_init()
     } else if (MG_SOCK_PENDING(rc)) {    // Need to wait for TCP handshake
       MG_DEBUG(("%lu %ld -> %M pend", c->id, c->fd, mg_print_ip_port, &c->rem));


scaprile:
Thx for the PR and the description, though I can't merge something just based on that.
According to what you say
> causing connect() to return 0 instead of EINPROGRESS

that is what has to be handled, if it is not, so we'll work on that when we finish our current duties.
I'll appreciate if you can add some logs where that can be observed, and, since
> Taixin 82x series (TS826/TS828) embedded platform

is not a networking stack, please tell us what is your networking stack, and how you configured Mongoose for it and it didn't work for you
Thank you

PS: BTW, we run in lot of embedded platforms and perform local connections and haven't observed this with the usual networking stacks, including our own

thinker132203:
Hi @scaprile,

Thank you for your rigorous feedback. I fully understand the caution regarding merging changes into the core logic without sufficient evidence.

To clarify the environment: I am using lwIP (a version customized by Taixin Semiconductor for the TX82x series). You are correct that connect() typically returns EINPROGRESS in standard non-blocking scenarios. However, in this specific vendor's SDK, connect() returns 0 synchronously under certain conditions.

I would like to point out that mg_connect_resolved() already contains a logic branch dedicated to handling rc == 0:

C
if (rc == 0) {
    setlocaddr(FD(c), &c->loc);
    mg_call(c, MG_EV_CONNECT, NULL);
    if (!c->is_tls_hs) c->is_tls = 0;
}
My PR simply highlights that this existing branch is not aligned with the asynchronous branch in connect_conn() regarding TLS initialization. It currently lacks the mg_tls_handshake() call. It is highly likely that previous users who triggered this rc == 0 path were not using TLS, which is why this discrepancy remained unnoticed.

Regarding the logs: because this bug is a silent failure caused by missing logic, Mongoose's built-in logging does not provide any explicit error messages. The logs simply show "everything is normal," followed by inactivity and an eventual timeout. Therefore, standard logs are of limited reference value here. If you are still interested, I will add custom instrumentation to my local build tomorrow to provide serial logs proving that connect() is indeed returning 0.

For your reference, the official SDK (TXW82x_FPV-v2.7.0.7-41244) can be downloaded from the vendor's website: https://www.taixin-semi.com/Product?prouctSubClass=73. The relevant socket implementation is located in sdk/lib/net/lwip/src/api/sockets.c.

PS: I agree that the underlying stack's behavior might not follow the "best practices" seen in major OS environments. However, for the sake of robustness, I believe the library should handle such synchronous successes gracefully rather than assuming the underlying implementation always follows ideal asynchronous patterns.

Best regards,

scaprile:
Handling in mg_connect_resolved() responds to expected behavior, what you request is something that only happens with your vendor modified version of a third-party stack and not in that third-party stack itself, or at least we haven't found that so far.
I will convert this to an issue and we will discuss it internally. If you're a paying customer please contact us at Support to serve you better. Thank you.

PR #3515, opened 2026-04-24
Fix out-of-bounds array access in function skip_chunk()

heguorui:
Line 2829 :  if (len < i + 1 || buf[i] != '\r' || buf[i + 1] != '\n') return -1;  // Error

If len < i + 1 is true, then len >= i + 1.When len = i + 1, buf[i + 1] will cause an out-of-bounds array access.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -2826,7 +2826,7 @@ static int skip_chunk(const char *buf, int len, int *pl, int *dl) {
   while (i < len && is_hex_digit(buf[i])) i++;
   if (i == 0) return -1;                     // Error, no length specified
   if (i > (int) sizeof(int) * 2) return -1;  // Chunk length is too big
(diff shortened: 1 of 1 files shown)


heguorui:
@scaprile Hi，could I kindly consult you on this?
Is there any additional bounds checking elsewhere that prevents this from being an out-of-bounds access?
The condition implies len >= i + 1 cannot be true if len < i + 1 holds.But when len = i + 1, buf[i + 1] causes array out-of-bounds access, which is captured by AddressSanitizer at runtime.

scaprile:
We're already working on that, and we don't publish security-related stuff here. Most claimed OOB (read) accesses are harmless, though some people claim some can be somewhat exploited under specific circumstances, so we don't keep that data here. Thank you.

heguorui:
Much appreciated for your feedback.

PR #3522, opened 2026-05-06
test: add local MQTT and SNTP parser coverage

parasol-aser:
## Summary

- Adds `test_mqtt_parse_local()`: deterministic MQTT v4/v5 parser tests for CONNACK, QoS0/1 PUBLISH, MQTT5 PUBLISH with payload-format/content-type/user-property/message-expiry properties, and malformed inputs (bad varint, truncated topic, truncated property section, truncated property value).
- Adds `test_sntp_parse_local()`: deterministic SNTP parser tests for valid v4 server and v3 broadcast replies, short packets, client-mode rejection, kiss-of-death (stratum 0), and unsupported version.
- Both groups registered before the `LOCALHOST_ONLY` network gate, so they run in all CI configurations.

## Notes

During development, a `MQTT_PROP_SUBSCRIPTION_IDENTIFIER` assertion triggered a UBSAN misaligned-store error in production code:

```
src/mqtt.c: runtime error: store to misaligned address for type 'size_t', which requires 8 byte alignment
```

Root cause: `mg_mqtt_next_prop()` passes `(size_t *) &prop->iv` to `decode_varint()`, but `prop->iv` is `uint32_t` at struct offset 4 — only 4-byte aligned. The tests in this PR avoid `MQTT_PROP_TYPE_VARIABLE_INT` properties; the underlying production bug warrants a separate fix.

## Test plan

- [x] `V=0 make -C test test TFLAGS=-DLOCALHOST_ONLY` — passes, `SUCCESS. Total tests: 1627`
- [x] `V=0 make -C test coverage TFLAGS=-DLOCALHOST_ONLY` — passes, `Lines executed:79.37% of 7577`
- [x] `git diff --check` — clean

### DIFF
--- a/test/unit_test.c
+++ b/test/unit_test.c
@@ -369,6 +369,59 @@ static bool test_sntp_server(const char *url) {
   return ms > 0;
 }
 
+static void sntp_put_timestamp(unsigned char *buf, size_t ofs, int64_t ms) {
+  uint64_t msec = (uint64_t) ms;
+  uint32_t sec = (uint32_t) (msec / 1000U + 2208988800ULL);
+  uint32_t frac = (uint32_t) (((msec % 1000U) * 4294967295ULL) / 1000U);
+
+  buf[ofs] = (unsigned char) (sec >> 24);
+  buf[ofs + 1] = (unsigned char) (sec >> 16);
+  buf[ofs + 2] = (unsigned char) (sec >> 8);
+  buf[ofs + 3] = (unsigned char) sec;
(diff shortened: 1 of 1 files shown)


scaprile:
If there is an actual problem, please open an issue, not a PR. What we're not explicitly testing: we've tested somewhere else, and we won't fix an issue we don't know.
Thank you
#3523

parasol-aser:
@scaprile thank you. i see https://github.com/cesanta/mongoose/issues/3523 describes the UBSAN misaligned-store error. i left a comment there. Note: it is a correctness bug, not a security issue.

Issue #3523, opened 2026-05-06
problematic casting


scaprile:
@robertc2000 
https://github.com/cesanta/mongoose/blob/eee8c7077c031f22469ea3adfcc69fc6d86c479a/src/mqtt.c#L260
That struct element is a uint32_t and follows a uint8_t. That means 3 bytes wasted in padding (no charges) and 32-bit alignment. You could've probably got away with reversing the order, but as it is, on 64-bit systems that casting passes an unaligned pointer.
I think the best option is to probably make a local copy on a size_t local var, though the main point of this issue is to check if there are any more casts like that one in that code. Maybe decode_varint() can be refactored to accept a value instead of a pointer ? You're the man

parasol-aser:
To reproduce this bug, add the following at the end of test_mqtt_parse_local() in test/unit_test.c:         
  ```
  /* MQTT5 PUBLISH with Subscription Identifier = 5 */                                                                                
  /* 0x30        = PUBLISH QoS0                      */                                                                               
  /* 0x06        = remaining length 6                 */                                                                              
  /* 0x00, 0x01  = topic length 1                     */                                                                              
  /* 'x'         = topic                              */                                                                              
  /* 0x02        = property section length 2          */                                                                              
  /* 0x0B        = SUBSCRIPTION_IDENTIFIER property   */
  /* 0x05        = varint value 5                     */                                                                              
  const uint8_t sub_id_pkt[] = {0x30, 0x06, 0x00, 0x01, 'x', 0x02, 0x0B, 0x05};
  ASSERT(mg_mqtt_parse(sub_id_pkt, sizeof(sub_id_pkt), 5, &mm) == MQTT_OK);                                                           
  mg_mqtt_next_prop(&mm, &prop, 0);  /* triggers UBSAN misaligned store */                                                            
  ```                                                                                                               
  Then run:                                                                                                                           
```                                                                                                                     
  V=0 make -C test test TFLAGS=-DLOCALHOST_ONLY                                                                                       
```
You will see
```
src/mqtt.c:100:10: runtime error: store to misaligned address 0x7ffe3318b944 for type 'size_t', which requires 8 byte alignment
0x7ffe3318b944: note: pointer points here
  0b 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00
              ^ 
make: *** [Makefile:149: test] Error 1
```

parasol-aser:
[...]

PR #3527, opened 2026-05-07
Fix casting issue in MQTT parsing

robertc2000:


src/mqtt.c
```diff
@@ -466,8 +464,9 @@ int mg_mqtt_parse(const uint8_t *buf, size_t len, uint8_t version,
       if (p > end) return MQTT_MALFORMED;
       if (version == 5 && p + 2 < end) {
         len_len =
-            (uint32_t) decode_varint(p, (size_t) (end - p), &m->props_size);
+            (uint32_t) decode_varint(p, (size_t) (end - p), &tmp);
         if (!len_len) return MQTT_MALFORMED;
+        m->props_size = (size_t) tmp;
```
scaprile: cast is not necessary, props_size is size_t. Even the local copy is not necessary

I wonder what are varint length bounds... Maybe size_t breaks on 16-bit systems ? We should use the type that covers the specs.

robertc2000: Yes, I was thinking there could be an unaligned access on 64-bit systems, but now I realize size_t field should already be aligned on 64-bit, so cast is not needed.

Seems like varint is max 4 bytes long, so we should use `uint32_t` instead of `size_t`. I will update the code.

src/mqtt.c
```diff
@@ -465,8 +460,8 @@ int mg_mqtt_parse(const uint8_t *buf, size_t len, uint8_t version,
       }
       if (p > end) return MQTT_MALFORMED;
       if (version == 5 && p + 2 < end) {
-        len_len =
-            (uint32_t) decode_varint(p, (size_t) (end - p), &m->props_size);
+        len_len = (uint32_t) decode_varint(p, (size_t) (end - p),
+                                           (uint32_t *) &m->props_size);
```
scaprile, line 464: casting a size_t to uint32_t may not be naturally aligned on 16-bit systems. I wonder if we have more of these... anyway, I don't know if we run on any 16-bit system...
Let's remember this and move on...

scaprile (approved):

scaprile:
closes #3523

PR #3534, opened 2026-05-14
fix mqtt cast error

robertc2000:


robertc2000:
https://github.com/cesanta/mongoose/pull/3527#discussion_r3201440289

Looks like using the local variable was necessary after all, casting from size_t to uint32_t was triggering Valgrind.

scaprile:
Yeah, I was worried about that on 16-bit systems but apparently there's something else I've missed. Don't feel like looking at Intel assembly (eew) to understand it.

scaprile (approved):

PR #3537, opened 2026-05-14
Dns: randomize txnid for every request to prevent DNS spoofing

Cyberpsych0s1s:
Fixes predictable DNS transaction ids in `sendnsreq()`. Previously only
the first in-flight request got a random txnid; subsequent concurrent
requests used `prev + 1`. Combined with Mongoose's fixed DNS source
port, this enables off-path response spoofing (RFC 5452 / Kaminsky).

Fix: rejection-sample a fresh random id against the in-flight list.
See commit message for details.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -980,9 +980,17 @@ static void sendnsreq(struct mg_connection *c, struct mg_str *name, int ms,
   } else {
     struct dns_data *reqs = (struct dns_data *) c->mgr->active_dns_requests;
     uint16_t id;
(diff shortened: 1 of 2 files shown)


scaprile:
Thx, as the comment there suggests, we know what to do. We just decided that's not worth it.

Cyberpsych0s1s:
My bad. Thank you.

scaprile:
No problem, with such a small issue there's more admin work to format and accept your coding than actually coding it ourselves anyway.

PR #3547, opened 2026-05-19
MG_ARCH_CUBE detection for WiFi drivers

robertc2000:


src/arch_cube.h
```diff
@@ -36,7 +36,8 @@
 #endif
 
 #if MG_ENABLE_TCPIP && !defined(MG_ENABLE_DRIVER_STM32F) && \
-    !defined(MG_ENABLE_DRIVER_STM32H) && !defined(MG_ENABLE_DRIVER_STM32N)
+    !defined(MG_ENABLE_DRIVER_STM32H) && !defined(MG_ENABLE_DRIVER_STM32N) && \
+    !defined(MG_ENABLE_DRIVER_CYW_SDIO)
```
scaprile: CYW_SDIO is not for the Portenta, it is a generic SDIO layer that is also used with the CY8
What exactly is the issue ?

robertc2000: I was trying to create a Cube project for the Portenta and use MG_ARCH_CUBE for it and I noticed it enabled the STM32H driver by default

robertc2000: But now I think it's better to avoid using MG_ARCH_CUBE for this case, right?

scaprile: Well, I bet cpq will want it anyway...
Maybe it is any flavor of the CYW that you need to detect ? Though you will need to pass the proper driver anyway, unless you come up with a way to detect the Portenta... or the CYW...
I would add `&& !defined(MG_ENABLE_DRIVER_CYW)`, that will allow using also the RM2 board (that for some reason we haven't included in the Wizard...) and avoid problems with using the Portenta in SPI mode (possible, did it).
And then we'll have the same issue with the ST67W, so we'd better think ahead.

scaprile: It is not an issue of the driver not being compatible with ARCH_CUBE, but ARCH_CUBE not being smart enough to detect what lies underneath...

robertc2000: Yes, we would need to include the ST67W too.. `&& !defined(MG_ENABLE_DRIVER_ST67W6)`?

robertc2000: Yeah, we need to pass the driver... I am currently doing this, define both ARCH_CUBE and the driver in mongoose_config. I don't think there is a way to detect in ARCH_CUBE if we need to use CYW... For U5  it could be detected on the other hand.
```
#define MG_ARCH MG_ARCH_CUBE
#define MG_ENABLE_DRIVER_CYW_SDIO 1
```

But that's why we have the else branch there, to warn the user to select another driver if none of the above are selected. So the above snippet would make sense.

scaprile (approved):

PR #3548, opened 2026-05-19
Fix s_link duplicate

robertc2000:


robertc2000:
Building st67w driver failed because of this

src/drivers/st67w6.c
```diff
@@ -6,7 +6,7 @@
 
 static struct mg_tcpip_if *s_ifp;
 static uint32_t s_ip, s_mask;
-static bool s_link = false, s_connecting = false;
+static bool s_wifi_link = false, s_connecting = false;
```
scaprile:
```sh
$ grep -l s_link src/drivers/*.c
src/drivers/cyw.c
src/drivers/st67w6.c
```
The same will happen later.
I would change ppp only, and use a more proper name. In the case of PPP, there's no actual physical link per se, it is LCP being up, so we can use s_lcpup, s_isup, whatever; something short enough to keep readability (not causing unnecessary line breakings)
And leave the drivers as they are. It is PPP breaking stuff, not the drivers

scaprile (changes requested):

scaprile (approved):

PR #3552, opened 2026-05-20
Fix #2694

### LINKED ISSUE #2694: MacOS unit tests fail with MbedTLS
```
FAILURE test/unit_test.c:1188: fetch(&mgr, buf, url, "GET /a.txt HTTP/1.0\n\n") == 200
```

cpq:


src/sock.c
```diff
@@ -341,7 +341,7 @@ static void read_conn(struct mg_connection *c) {
       // there can still be > 16K from last iteration, always mg_tls_recv()
       m = c->is_tls_hs ? (long) MG_IO_WAIT : mg_tls_recv(c, buf, len);
       if (n == MG_IO_ERR || n == MG_IO_RESET) {  // Windows, see #3031
-        if (c->rtls.len == 0 || m < 0) {
+        if (c->rtls.len == 0 || (m < 0 && m != MG_IO_WAIT)) {
```
scaprile, line 344: Is there a real reason with enough wait to touch this ? We've been there and trashed our stack several times; it is working fine now and I don't see why it would be different win Mac. If it is, we should carefully document why and avoid trashing the other systems

src/tls_mbed.c
```diff
@@ -122,11 +122,6 @@ void mg_tls_init(struct mg_connection *c, const struct mg_tls_opts *opts) {
   if (c->is_listening) goto fail;
   MG_DEBUG(("%lu Setting TLS", c->id));
   MG_PROF_ADD(c, "mbedtls_init_start");
-#if defined(MBEDTLS_VERSION_NUMBER) && MBEDTLS_VERSION_NUMBER >= 0x03000000 && \
-    defined(MBEDTLS_PSA_CRYPTO_C)
-  psa_crypto_init();  // https://github.com/Mbed-TLS/mbedtls/issues/9072#issuecomment-2084845711
-  // this initializes global resources and then just returns when called again
-#endif
```
scaprile, line 129: This has been added for a reason, why does it need to be removed now ? It's been working for all the other versions in all the other systems

scaprile (commented):
Too many changes with no explanation, only for Mac ?

src/tls_mbed.c
```diff
@@ -242,32 +245,46 @@ void mg_tls_flush(struct mg_connection *c) {
...
 #if defined(MBEDTLS_VERSION_NUMBER) && MBEDTLS_VERSION_NUMBER >= 0x04000000
-    if ((rc = mbedtls_ssl_ticket_setup(&ctx->tickets, PSA_ALG_GCM,
-                                       PSA_KEY_TYPE_AES, 128, 86400))
+      rc = mbedtls_ssl_ticket_setup(&ctx->tickets, PSA_ALG_GCM,
+                                    PSA_KEY_TYPE_AES, 128, 86400);
 #else
-    if ((rc = mbedtls_ssl_ticket_setup(&ctx->tickets, mg_mbed_rng, NULL,
-                                       MBEDTLS_CIPHER_AES_128_GCM, 86400))
+      rc = mbedtls_ssl_ticket_setup(&ctx->tickets, mg_mbed_rng, NULL,
+                                         MBEDTLS_CIPHER_AES_128_GCM, 86400));
```
scaprile: Linux:
```txt
src/tls_mbed.c: In function ‘mg_tls_ctx_init’:
src/tls_mbed.c:275:76: error: expected ‘;’ before ‘)’ token
src/tls_mbed.c:275:76: error: expected statement before ‘)’ token
```
```txt
src/tls_mbed.c:275:76: error: extraneous ')' before ';'
  275 |                                          MBEDTLS_CIPHER_AES_128_GCM, 86400));
      |                                                                            ^
```

src/sock.c
cpq: Linux with mbedtls 4 fails either. This branch works.

src/tls_mbed.c
cpq: Yeah, this snippet was just moved to ctx_init which is called unconditionally at mgr_init

src/sock.c
scaprile: "works" is subjective... :wink: Besides that, you are right, it makes sense. But... you didn't think of corner cases like a connection being dropped with a truncated record in the buffer. Then that connection might never close.

src/tls_mbed.c
```diff
@@ -242,32 +245,46 @@ void mg_tls_flush(struct mg_connection *c) {
...
+  psa_crypto_init();  // Initializes global PSA resources, no-op if already done
+#endif
 #ifdef MBEDTLS_SSL_SESSION_TICKETS
-    int rc;
-    mbedtls_ssl_ticket_init(&ctx->tickets);
+  {
+    struct mg_tls_ctx *ctx = (struct mg_tls_ctx *) mg_calloc(1, sizeof(*ctx));
+    if (ctx == NULL) {
+      MG_ERROR(("TLS context init OOM"));
+    } else {
```
scaprile, line 266: This was being done regardless of MBEDTLS_SSL_SESSION_TICKETS being defined or not.
Why the change ?

src/tls_mbed.c
```diff
@@ -242,32 +245,46 @@ void mg_tls_flush(struct mg_connection *c) {
...
-                                       MBEDTLS_CIPHER_AES_128_GCM, 86400))
+      rc = mbedtls_ssl_ticket_setup(&ctx->tickets, mg_mbed_rng, NULL,
+                                         MBEDTLS_CIPHER_AES_128_GCM, 86400);
 #endif
-        != 0) {
-      MG_ERROR((" mbedtls_ssl_ticket_setup %#x", -rc));
+      if (rc != 0) {
+        MG_ERROR((" mbedtls_ssl_ticket_setup %#x", -rc));
+        mbedtls_ssl_ticket_free(&ctx->tickets);
+        mg_free(ctx);
```
scaprile, line 279: This wasn't being done

src/tls_mbed.c
```diff
@@ -242,32 +245,46 @@ void mg_tls_flush(struct mg_connection *c) {
...
+      } else {
+        mgr->tls_ctx = ctx;
```
scaprile, line 281: Are you now freeing context on failure and we weren't doing that before ?

scaprile (commented):
I see now it looks you're right on the sock.c change, but...

src/tls_mbed.c
cpq: Yes exactly, the presence of context is the signal we have tickets initialised.

Before, they might have been failed to initialise, and we still tried to enable then which was one of the failure reasons. Instead of introducing a flag that tickets are OK, we're using the context presence as such flag, cause the context is not used for anything else.

PR #3556, opened 2026-05-21
Mbed2

cpq:


src/sock.c
```diff
@@ -324,42 +325,40 @@ static bool ioalloc(struct mg_connection *c, struct mg_iobuf *io) {
...
-      if (c->rtls.len < 16 * 1024 + 40) {  // TLS record, header, MAC, padding
-        if (!ioalloc(c, &c->rtls)) return;
-        n = recv_raw(c, (char *) &c->rtls.buf[c->rtls.len],
-                     c->rtls.size - c->rtls.len);
-        if (n > 0) c->rtls.len += (size_t) n;
-      }
-      // there can still be > 16K from last iteration, always mg_tls_recv()
-      m = c->is_tls_hs ? (long) MG_IO_WAIT : mg_tls_recv(c, buf, len);
-      if (n == MG_IO_ERR || n == MG_IO_RESET) {  // Windows, see #3031
-        if (c->rtls.len == 0 || m < 0) {
```
scaprile, line 344: Removing this breaks Windoze, there's an issue, Windoze "sometimes" returns -3, IO_RESET

src/sock.c
```diff
@@ -324,42 +325,40 @@ static bool ioalloc(struct mg_connection *c, struct mg_iobuf *io) {
...
     } else {
-      n = recv_raw(c, buf, len);
+      m = mg_tls_recv(c, &c->recv.buf[c->recv.len], c->recv.size - c->recv.len);
     }
-    MG_DEBUG(("%lu %ld %lu:%lu:%lu %ld err %d", c->id, c->fd, c->send.len,
-              c->recv.len, c->rtls.len, n, MG_SOCK_ERR(n)));
-    iolog(c, buf, n, true);
+    if (n == MG_IO_ERR &&
+        (m == MG_IO_ERR || (c->rtls.len == 0 && mg_tls_pending(c) == 0) ||
+         (c->is_tls_hs && c->rtls.len == rtls &&
```
scaprile, line 349: I think the first clause will catch c->is_tls_hs and break stuff.
We should start with !c->is_tls_hs && as in
```c
    (!c->is_tls_hs && (m == MG_IO_ERR || (c->rtls.len == 0 && mg_tls_pending(c) == 0)) ||
         (c->is_tls_hs && c->rtls.len == rtls &&

File: src/tls.h [hunk 3]
```diff
@@ -37,6 +37,6 @@ void mg_tls_ctx_free(struct mg_mgr *);
 #define MG_IS_DER(buf) (((uint8_t *) (buf))[0] == 0x30)  // DER begins with 0x30
 
 // Low-level IO primives used by TLS layer
-enum { MG_IO_ERR = -1, MG_IO_WAIT = -2, MG_IO_RESET = -3 };
```
scaprile, line 40: Windoze

src/tls_mbed.c
```diff
@@ -75,16 +75,16 @@ static int mg_net_send(void *ctx, const unsigned char *buf, size_t len) {
   long n = mg_io_send((struct mg_connection *) ctx, buf, len);
   MG_VERBOSE(("%lu n=%ld e=%d", ((struct mg_connection *) ctx)->id, n, errno));
   if (n == MG_IO_WAIT) return MBEDTLS_ERR_SSL_WANT_WRITE;
-  if (n == MG_IO_RESET) return MBEDTLS_ERR_NET_CONN_RESET;
+  // if (n == MG_IO_RESET) return MBEDTLS_ERR_NET_CONN_RESET;
```
scaprile, line 78: For some reason, explained in that issue, MbedTLS needs to know of conn resets. Probably only in Windoze, I don't know. Why do we need to remove that and risk regression issues ?

src/sock.c
```diff
@@ -324,42 +325,40 @@ static bool ioalloc(struct mg_connection *c, struct mg_iobuf *io) {
...
+      if (m < 0) m = MG_IO_ERR;  // but return last record data, see #3104
+    }
+    n = m;
+  } else {
+    n = recv_raw(c, &c->recv.buf[c->recv.len], c->recv.size - c->recv.len);
   }
+  MG_DEBUG(("%lu %ld %lu:%lu:%lu %ld err %d", c->id, c->fd, c->send.len,
+            c->recv.len, c->rtls.len, n, MG_SOCK_ERR(n)));
+  iolog(c, (char *) &c->recv.buf[c->recv.len], n, true);
 }
```
scaprile, line 362: This looks nice...
I can't grasp any possible corner cases, but we had them all caught. This risks many regressions, what do we have to gain ? Is it worth it ?

scaprile (changes requested):

scaprile (approved):

PR #3579, opened 2026-06-01
check DER long-form length bounds in mg_der_to_tlv

alhudz:
mg_der_to_tlv() parses ASN.1 from an untrusted TLS peer (the ECDSA signature in CertificateVerify). On a long-form length it reads der[1]-0x80 length bytes without checking they fit in dersz, so der[2..128] can be read past the buffer. Add the same guard mg_der_parse already has.


cpq:

alhudz:
Sure, will file it through the advisories page. Same untrusted-peer path applies to the other two length checks I sent (#3581, #3582), happy to roll them into the same advisory if you'd prefer one report.

scaprile:

PR #3581, opened 2026-06-01
bound certificate list and entry lengths in mg_tls_recv_cert

alhudz:
mg_tls_recv_cert uses the wire certificate-list and per-entry lengths without bounding them against recv_len, so a crafted peer Certificate message drives the reads at cert+certsz past the rtls buffer.


cpq:

alhudz:
Sure, will do. I'll file it through the advisories form with the repro details and reference this branch there. Want me to close this public PR once that's up, or leave it for you to link the fix?

scaprile:

PR #3582, opened 2026-06-02
bound session_id_len in mg_tls_server_recv_hello

alhudz:
mg_tls_server_recv_hello() reads session_id_len from the ClientHello at offset 43 and uses it to locate cipher_suites_len at rio->buf[44 + session_id_len], with only a rio->len >= 50 check. A client sending session_id_len up to 255 in a short record reads past the buffer (ASAN flags a heap-buffer-overflow ~250 bytes out). Bound it against rio->len before use, the same way the later length fields are checked.


cpq:

alhudz:
Sure, will do.

scaprile:

PR #3583, opened 2026-06-02
Added Fix for WolfSSL not working correctly as server

cywa-fvd:
Looks like there might be a bug in the TLS init and free functionality for WolfSSL. WolfSSL seams to require certificates and keys loaded into the **SSL_CTX**-object before calling **SSL_new** which the OpenSSL compatability layer doesn't provide. Thus SSL_new will always fail as the **SSL**-object pointer is NULL. This fix provides WolfSSL-specific function calls to set the certificate and key before calling **SSL_new** with the context.

The second problem is with the way OpenSSL and WolfSSL free their **SSL**-objects. WolfSSL frees the **BIO_METHOD** and the **SSL_CTX** internally via a function for resource freeing. OpenSSL does not do that and requires additional function calls for resource freeing. These function calls freeze/crash Mongoose when using WolfSSL. The fix simply checks that **MG_TLS** is not set to **MG_TLS_WOLFSSL** and only frees the resources manually on not-WolfSSL.

Fix has been manually tested on:

- Microsoft Windows 11 using WolfSSL, OpenSSL and Builtin TLS
- Linux (Ubuntu) via WSL using WolfSSL, OpenSSL and Builtin TLS
- macOS using WolfSSL, OpenSSL and Builtin TLS

Testing was only done with HTTP server and not client.

### DIFF
--- a/mongoose.c
+++ b/mongoose.c
@@ -19835,8 +19835,10 @@ void mg_tls_free(struct mg_connection *c) {
   struct mg_tls *tls = (struct mg_tls *) c->tls;
   if (tls == NULL) return;
   SSL_free(tls->ssl);
+#if MG_TLS != MG_TLS_WOLFSSL
   SSL_CTX_free(tls->ctx);
   BIO_meth_free(tls->bm);
+#endif
   mg_free(tls);
   c->tls = NULL;
 }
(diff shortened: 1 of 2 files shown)


scaprile:
"WolfSSL" is working properly and all tests pass.
Maybe "WolfSSL" for you is some specific version of it ? 
The issue template is there for a reason: to save yours and our time. Please honor the issue template and fill it as requested. Always open an issue, not a PR. Thank you.

cywa-fvd:
Then the issue is on my end.
Apologies and thank you for your time.

PR #3586, opened 2026-06-03
sanitize dash paths

scaprile:


robertc2000 (approved):

cpq:
Please don't - I'll revamp dash API a lot soon, and this codepath is not event present anymore.
I suggest to drop this PR

scaprile:
@cpq OK, just blame it on Claude that pointed this out

PR #3591, opened 2026-06-05
Fix Risc-V assembly instruction in mg_picosdk_write

will-v-pi:
I think that there is a typo here - `csrrs` is supposed to be `csrs`, as `csrrs` would need a register before `mstatus` (like the `csrrc` above)

Noticed in https://forums.raspberrypi.com/viewtopic.php?t=398792


scaprile:
Do you  actually have an issue ? Please open an issue and follow the issue template. We won't merge a PR because "you think", programs compile or not. Thank you

scaprile:
Do you  actually have an issue ?
You say "you think", programs compile or not.
Are you able to compile ?  It builds for us.
Not ? Then fill the issue template and tell us what we ask.
 Thank you

PR #3601, opened 2026-06-13
OTA rollback nits

cpq:


cpq:
Pushing TBR
@robertc2000 please test OTA on stm32 F4 and F7

robertc2000:
> Pushing TBR @robertc2000 please test OTA on stm32 F4 and F7

Tested and works!

PR #3609, opened 2026-06-22
OSS-Fuzz: Add new fuzzer targets ws approach

arthurscchan:
This PR adds a new OSS-Fuzz fuzzer targets ws approach, together with the fix of build in OSS-Fuzz (https://github.com/google/oss-fuzz/pull/15786).


scaprile:
Not interested, thank you anyway.
Please describe what you think we are not fuzzing and why that is needed. Thanks again.

PR #3618, opened 2026-07-01
Handle no pending connections in FreeRTOS TCP

robertc2000:


src/sock.c
```diff
@@ -465,6 +465,14 @@ static MG_SOCKET_TYPE raccept(MG_SOCKET_TYPE sock, union usa *usa,
   do {
     memset(usa, 0, sizeof(*usa));
     fd = accept(sock, &usa->sa, len);
+#if MG_ENABLE_FREERTOS_TCP
+    // FreeRTOS_accept() returns NULL to mean "no pending connection",
+    // therefore we avoid avoid retrying it forever on non-blocking listeners
```
scaprile: avoid avoid the avoid avoid duplication 😜

scaprile (approved):

PR #3626, opened 2026-07-06
make tls_init request boot time if it is 0

scaprile:


scaprile:
> ## Claude PR Security Review

I see, good catch Claudio...
This happens when you ask an AI in order to save some time... and don't watch how it is done in DNS.
I'm supposed to have fixed that in the second commit, with an allegoric name.

PR #3628, opened 2026-07-07
make tls_init request boot time if it is 0

scaprile:
After going for a generic solution to a non-existent problem, have decided to keep things simple(st)
The scope of this is to help in keeping simple tutorials simple, not to provide a system time base; that is best served by our Wizard scheme.
So, the major caveat in this scheme is that several simultaneous connections requiring a time base to validate a certificate, will each one generate SNTP traffic. Once a connection updates the time base, next ones will use it.
Removing such a nuisance requires a scheme like our DNS resolver scheme, which is, IMHO, too much for what is described above: system time must be managed by a system time strategy, this is a helper to keep examples simple. IMHO.

### DIFF
--- a/src/tls_builtin.c
+++ b/src/tls_builtin.c
@@ -112,6 +112,8 @@ struct tls_data {
   bool skip_verification;    // do not perform checks on server certificate
   bool cert_requested;       // client received a CertificateRequest
   bool is_twoway;            // server is configured to authenticate clients
+  bool is_sntp_pending;      // TLS handshake is waiting for wall-clock time
+  struct mg_connection *timec;  // SNTP connection for wall-clock time
   struct mg_str cert_der;    // certificate in DER format
   struct mg_str ca_der;      // current CA certificate
   struct mg_str *ca_bundle_der; // bundle of CA certificates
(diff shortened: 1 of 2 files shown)


scaprile:
Regarding Claudio's opinion: as I see it, the same can already happen with our current SNTP strategy for system time, so if you gentlemen consider this has to be addressed, I will do it on a separate PR and we'll add whatever we come up with to  Wizard code

> request/response binding validation

Yeah, like someone able to do a MITM attack will not be able to forge that too... SNTP seems as fragile as DNS. No claims while no publicly exposed code, so far, that's the reason. (and something like that requires state to be saved...)

robertc2000 (approved):
Looks like the root issue is in SNTP's lack of security. Nothing we can do about it. This should only be used for tutorials, and, like Claude says, maybe we should document it, so the users are aware to fetch a reliable timebase first.

src/tls_builtin.c
```diff
@@ -112,6 +112,8 @@ struct tls_data {
   bool skip_verification;    // do not perform checks on server certificate
   bool cert_requested;       // client received a CertificateRequest
   bool is_twoway;            // server is configured to authenticate clients
+  bool is_waiting_time;      // TLS handshake is waiting for wall-clock time
```
cpq: bad name IMO, "waiting time" sounds like literally "waiting time"
"is_pending_sntp" would be better

cpq (approved):

PR #3633, opened 2026-07-13
harden some len checks

scaprile:
DHCP code now handles PADs, that we weren't, and stops parsing on any malformed option, that we weren't. Even though we handled the message safely, we were trying to keep parsing after a malformed option, which would end in the middle of another broken option... but because of us...


scaprile:
@robertc2000 Can we run Claude again ? It found an error and I refactored the code, then deleted its comment as it was misleading
(didn't want to close the PR and open a new one...)

PR #3638, opened 2026-07-17
Allow exact-size buffers in mg_url_encode

94xhn:
`mg_url_encode()` currently reserves four bytes before every input byte and rejects exact-size output buffers. `a` with 2 bytes and `@` with 4 bytes both return 0.

Use the actual encoded width (1 or 3 bytes) while reserving one byte for the terminator. Add exact-fit and one-byte-short coverage.

Tests: Linux unit suite with ASan/UBSan (`LOCALHOST_ONLY`, 1802 tests); MinGW GCC 8 boundary harness.


scaprile:
Not interested, extra room there is there for a reason

PR #3641, opened 2026-07-20
Add ESP32 OTA rollback support

robertc2000:
* Added OTA rollback support for ARCH_ESP32
* Fixed a bug in `mg_ota_poll`: currently, we are feeding the timer when the state is `MG_OTA_TESTING`. However, this naming is a bit misleading, because we are in this state only from the moment before swapping banks until the first boot of the new firmware, so the timer will never be fed under this condition. Right after the first boot, we transition to `MG_OTA_FAILED` state, which is the actual "testing" phase in which we feed the timer.


cpq (approved):
Looks good!

PR #3656, opened 2026-07-29
Add mg_crc16(), make PPP use it

scaprile:
PPP now uses a faster smaller table-lookup algorithm based on the CRC32 engine. It is also exposed for convenience, as several protocols (HDLC-based) use it, and it can be a lighter-weight check than mg_crc32() in other places.


cpq:
LGTM

PR #3659, opened 2026-08-06
make FIN wait for TLS drain

scaprile:
Updates code (#3148) written before TCP started honoring peer rx window (#3572).
New plain send adapted fine, but built-in TLS was still operating under the assumption that we should empty and exit. Now we can be blocked by a full window, or even a zero window update (not yet), so we must not send FIN until the plain buffer is empty AND TLS has sent any ciphered data in its buffer (thing that we formerly ignored because we enforced that in our loop and there was no way to throttle us, now there is...)
We've also been ignoring errors because we were about to close anyway, now we're not.
Adds a test to check this is true.
Socket code is mirrored, is not directly tested because it requires a lot of hassle and it doesn't seem to be necessary, being it a mirror.


robertc2000 (approved):
Looks good! Please also include the updated mongoose.c, it is currently missing.

cpq (approved):

PR #3666, opened 2026-08-14
Add Ethernet controller (NETC) driver for IMXRT1180

robertc2000:


src/drivers/phy.c
```diff
@@ -75,6 +82,16 @@ void mg_phy_init(struct mg_phy *phy, uint8_t phy_addr, uint8_t config) {
   id2 = phy->read_reg(phy_addr, MG_PHY_REG_ID2);
   MG_INFO(("PHY ID: %#04x %#04x (%s)", id1, id2, mg_phy_id_to_str(id1, id2)));
 
+  if (id1 == MG_PHY_YT85x && id2 == MG_PHY_YT8531) {
+    uint16_t value;
+    phy->write_reg(phy_addr, MG_PHY_REG_BCR, MG_BIT(12)); // force auto-neg
```
scaprile: This is not good. This is userland and boardland, the hw has to set this up; users might want otherwise, boards might require otherwise.
IIRC we have this issue with other NXP boards and we set GPIOs to configure the PHY at reset
https://mongoose.ws/documentation/tutorials/hardware/
> Mongoose drivers do not configure the PHY for a specific link, they honor whatever has been defined in hardware by board designers. Most commonly, boards are designed to configure the PHY for auto-negotiation. Once the PHY connects with a switch, it reports whether it is working in 10 or 100 Mbps and in full- or half-duplex; Mongoose then configures the Ethernet controller to work accordingly, as clock frequency and data format in data pins depend on this.

The HAL must configure the board if the hardware does not.

If for some reason this is not possible, weird, as all PHYs I know are configured by pins on reset or default to auto-negotiation, then we can do this for now, but clearly comment it, and sadly later maybe add a config to choose link type.

src/drivers/netc.c
```diff
@@ -0,0 +1,469 @@
...
+  MG_DSB();
+  return true;
+}
+
+static size_t mg_tcpip_driver_netc_tx(const void *buf, size_t len,
+                                      struct mg_tcpip_if *ifp) {
+  struct mg_tcpip_driver_netc_data *d =
+      (struct mg_tcpip_driver_netc_data *) ifp->driver_data;
+  uint16_t next;
+  s_tx_clean =
```
scaprile, line 350: What is this ? A static initialized with a chip value ?
Is the chip telling us where its descriptors are in memory ?

src/drivers/netc.c
```diff
@@ -0,0 +1,469 @@
...
+  NETC_ENETC1_SI0->BDR[0].RBICR0 =
+      MG_BIT(31) | MG_BIT(0);  // ICEN, interrupt after one frame
+  NETC_ENETC1_SI0->BDR[0].RBICR1 = 0;  // no timer threshold
+  NETC_ENETC1_SI0->BDR[0].RBBAR0 = (uint32_t) rx_base;
+  NETC_ENETC1_SI0->BDR[0].RBBAR1 = (uint32_t) (rx_base >> 32);
+  NETC_ENETC1_SI0->BDR[0].RBBSR = ETH_PKT_SIZE;
+  NETC_ENETC1_SI0->BDR[0].RBPIR = 0;
+  NETC_ENETC1_SI0->BDR[0].RBCIR = 0;
+  NETC_ENETC1_SI0->BDR[0].RBLENR = ETH_DESC_CNT;
+
```
scaprile, line 318: Is all this stuff above really necessary ? Aren't any of these actually default after reset ?

scaprile (changes requested):
Check that PHY init stuff
See if you can trim this a bit... otherwise fine for me and in any case we'll see later

src/drivers/phy.c
robertc2000: For this PHY, autonegotiation bit was always zero when reading BCR after right after resetting it in mg_phy_init, the only way I could get this was to work was to manually set that bit at line 87

src/drivers/netc.c
robertc2000: Leftovers, forgot to remove them. I will check their reset values.

src/drivers/netc.c
robertc2000: That is the tail of the descriptors still owned by the DMA, it is just an index. If we reach this tail, it means we run out of descriptors.

src/drivers/phy.c
scaprile: OK, let's move on, I'll make a note on this and we'll resume September or October, I'll check the datasheet and board info and see what we can do to make things standard, or otherwise craft some cool workaround.

robertc2000: OK, I will leave a comment

scaprile (approved):

PR #3682, opened 2026-09-07
add optional TCP retransmission

scaprile:
Fixes re-opening after receiving zero window
#c857dda fixed ACKing duplicate segments, allowing peer retransmissions
Setting `ifp->enable_tcp_retransmit`adds a per-connection iobuf to hold TCP segments as they are sent. The iobuf grows as needed up to the max window size (MG_TCPIP_WIN). 
`MG_TCPIP_RTX_MS: 1s, Initial retransmission timeout, exponentially increasing up to
`MG_TCPIP_RTX_MAX_MS`: 60s, minimum permitted RTO cap
Proper ACKs reset a timer, if it expires, the first segment in the iobuf is retransmitted.
Proper ACKs delete outstanding segments in the iobuf

Adds two new tests, one for little-endians and one for big-endians, that run mip_tap_test with a 30% segment loss probability, in order to exercise this on live networks. It takes care of not losing SYNs nor FINs, but sent ACKs and payloads are randomly discarded; this exercises this very retransmission and our response to peer retransmissions (our ACKs get lost). Due to rtx timeouts, loops are now extended x10 with a macro, to avoid doing it in normal runs.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -119,10 +119,15 @@ jobs:
 
   miprtx:
     runs-on: ubuntu-latest
-    name: MIP RTX TAP
+    strategy:
+      fail-fast: false
+      matrix:
+        ssl: ["", BUILTIN]
+    name: MIP RTX TAP SSL=${{ matrix.ssl }}
     env:
       IPV6: 0
       RTX: 1
+      SSL: ${{ matrix.ssl }}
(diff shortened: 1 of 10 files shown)


scaprile:
> causing unbounded per-poll heap growth and violating the txq invariant

The queue is capped to 6K (MG_TCPIP_WIN), but I'll check this later. Please review and I'll push the fix later.

robertc2000 (approved):

cpq (approved):

PR #3692, opened 2026-09-12
health record support

cpq:


gvanem:
I'm getting this error with `clang-cl` in `test/mip_test.c` after this change:
```c
In file included from test/mip_test.c:14:
src/health.c(6,18): error: tentative definition has type 'struct mg_health' that is never completed
    6 | struct mg_health mg_health_record MG_HEALTH_RAM;
      |                  ^
src/health.c(6,8): note: forward declaration of 'struct mg_health'
    6 | struct mg_health mg_health_record MG_HEALTH_RAM;
      |        ^
1 error generated.

```

How is `src/health.h` supposed to be included in `mongoose.c`? It's not AFAICS. 
Hence the definition of `struct mg_health` is unknown.

cpq:
@gvanem weird, the header is there - https://github.com/cesanta/mongoose/blob/master/src/health.h and it gets agalgamated, https://github.com/cesanta/mongoose/blob/ed5ef09d1c300295b67771d73a91af2195a8a876/test/Makefile#L109

gvanem:
> ... and it gets agalgamated,


I did use the `mongoose.[ch]` in the root of the project. No `src/health.h` in `./mongoose.[ch]` here.

So you mean the amalgamation could be different in each `Makefile`? If so, yikes! Confusing.
Here `test/mongoose.h` is a symlink to `./mongoose.h`. Should it not be that always?

scaprile:
contents of src/health.h are in mongoose.h; and test/Makefile amalgamates src/health.h into mongoose.h
contents of src/health.c are in mongoose.c; and test/Makefile amalgamates src/*.c into mongoose.c
our tests call the `mongoose.c` and `mongoose.h` targets in src/Makefile, so they rebuild as needed
our tests pass, so check what is different for you

```c
make -C test mongoose.c
make -C test mongoose.h
```

The Makefile is in test/ because Mongoose itself is two files and does not need a Makefile.
Our non-amalgamated test is also passing.

PR #3693, opened 2026-09-13
Fix mg_json_get miscounting numbers with a long exponent

afonsojanu:
mg_atod() caps the exponent digit loop with `exp < 308`, and that same
condition also decides whether the loop keeps moving past the digit
characters. Once `exp` reaches the cap it stops consuming input even
when the number has more exponent digits left, so `numlen` comes back
shorter than the actual token.

`mg_json_get()` trusts that `numlen` to know where a number ends, so it
resumes scanning from the middle of the leftover exponent digits and
misreads the rest of the document. In practice, a perfectly valid JSON
value with a long exponent breaks lookups for every key that follows
it:

```c
const char *json = "{\"a\": 1e9999999999, \"b\": 42}";
int n = 0;
int off = mg_json_get(mg_str(json), "$.b", &n);
// off == MG_JSON_INVALID (-2), even though "b" is right there
```

The fix separates the two concerns: `i` still advances for every digit
character, but the digit only gets folded into `exp` while `exp` is
still below the cap.

Added a regression test to `test_json()` in `test/unit_test.c` with the
case above (a sibling key right after the oversized exponent), and
regenerated `mongoose.c` from `src/json.c`. Ran the full unit test
suite locally (`SUCCESS. Total tests: 1896`).


scaprile:
Thanks, we don't think we'll be handling such exponents, so we may just cap them at 99.
Unless you can provide a clear issue with a clear description of a real life problem, and translate "miscounting numbers" into readable English; I'd close this
Thank you

PR #3695, opened 2026-09-15
Guard against NULL return in mg_file_printf

eunho87:
### Problem

`mg_file_printf()` in `src/fs.c` uses the result of `mg_vmprintf()` without checking it:

```c
data = mg_vmprintf(fmt, &ap);
va_end(ap);
result = mg_file_write(fs, path, data, strlen(data));
mg_free(data);
```

`mg_vmprintf()` returns the buffer of a `mg_iobuf` that starts empty (`{0, 0, 0, 256}`). The buffer is grown by `mg_iobuf_resize()`, which uses `mg_calloc()` and, on allocation failure, leaves `io->buf` as `NULL` and returns `false`. In that case `mg_vmprintf()` returns `NULL`, and `strlen(data)` dereferences a NULL pointer.

The trigger is memory-allocation failure, so this is a robustness issue rather than a remotely exploitable one. But the rest of the codebase consistently checks allocation results (`dns.c`, `json.c`, `fs.c`, `fs_fat.c`, `iobuf.c` all null-check `mg_calloc`), and `mg_file_printf()` is a public API function (declared in `fs.h`), so a NULL from `mg_vmprintf()` should be handled the same way instead of crashing.

### Fix

Only write and free when `mg_vmprintf()` returned a buffer; otherwise return `false` (the value `result` already holds):

```c
if (data != NULL) {
  result = mg_file_write(fs, path, data, strlen(data));
  mg_free(data);
}
return result;
```

The change is applied to both `src/fs.c` and the amalgamated `mongoose.c` so both build configurations stay in sync.


cpq:
@eunho87 Thank you ! Please sign Cesanta CLA, per https://github.com/cesanta/mongoose#contributions, so we can integrate that.

eunho87:
Thanks @cpq — I've signed the Cesanta CLA at https://cesanta.com/cla.html under eunho08077@gmail.com (GitHub: eunho87). Should be all set to integrate; let me know if anything else is needed.

PR #3699, opened 2026-09-18
fix casting

scaprile:
@robertc2000 Found this (but it was a `volatile` cast too) on an unrelated work. There are 2 calls to that function, the other one is from a plain `uint32_t *`; the function signature is with a `volatile` argument, though. So:
either the casting to `volatile` is not needed, or the second call function needs to work on `volatile` stuff instead.
Please check and take over this PR. Thank you.


robertc2000:
I would say `volatile` is not needed in `flash_page_start`. We are just comparing addresses (`char *`), we are not dereferencing anything.

scaprile:
Cool, now ?

robertc2000 (approved):
Looks good now!

PR #3700, opened 2026-09-18
Reimplement BSD layer

scaprile:
Took over from original @cpq work reimplementing the core and providing non-blocking and no-RTOS functionality
- UDP _may_ appear to work, but it does not honor datagram boundaries if they arrive faster than the application calls recv()
- listen backlog enforcement test commented out due to a net_builtin issue that will be tackled later
- missing functions will be added later
- user apps wanting BSD functions _may_ include mongoose.h. However...
- third-party libraries _must not_ need mods to work, they don't need to be aware of mongoose.h. That's why this PR provides a set of BSD includes. It is assumed that the build environment will not have these; however, definitions inside mongoose.h can be disabled. The trick to properly build BSD user apps or third-party libraries is to add an include path to wherever this tree is copied. Provided examples link it as _bsd_ and add `-Ibsd` at the Makefile. This makes _our_ `sys/select.h` shadow any possible system version of it. For other headers, we use the environment ones. Tested with newlib and on Linux
  - BSD code in provided tutorials only uses standard BSD includes, no knowledge of Mongoose nor its include files

TL;DR (AI-written) on the decisions made to fix the "what if"s the refactor brought back, follows:

## why we restored BSD endpoint state from master

The newer version used the Mongoose connection itself as the only record for a
BSD socket. That was smaller, but it disappeared as soon as Mongoose closed a
connection. A task blocked in recv() could be told about the error, but there
was nothing left to represent its fd afterwards. A later close(fd) could fail,
or worse, hit a newly reused fd.

The required part of the master design is a small BSD socket object. Mongoose
owns and frees its network connection; the BSD side keeps its socket object
until the BSD user calls close(). This means an error or close can wake a
waiting reader or writer safely, and a program can use one task for reads and
another for writes on the same socket.

The new shim must remain queue-less. The BSD object supplies clear ownership,
bounded pending accepts, clean timeout handling, and predictable shutdown; it
does not require restoring the master per-socket data and accept queues.

The newer code still has a good core idea: use the same operation state machine
for both cases. The ordinary socket path stays common between FreeRTOS and
no-RTOS builds, without reintroducing a queue-backed transport.

The missing pieces are where ownership crosses that boundary: a connection
accepted before the application asks for it, a connection closing while a task
waits, and a BSD fd that must remain meaningful after Mongoose has freed its
network object. Keep the simpler common path where it helps, but restore the
persistent endpoint state from master for those cases. A no-RTOS endpoint can
use the same lifetime model.

## UDP diagnostic

The current branch is not yet BSD-compatible for UDP. sendto() changes the
endpoint's single c->rem address and sends using its current cached L2 address;
it neither resolves a new destination nor preserves the peer established by
connect(). This only works accidentally when the destination matches the last
resolved peer.

rx_udp() appends every UDP payload to flat c->recv and overwrites c->rem with
each sender. Therefore recvfrom() cannot report the source that belongs to the
front payload, and neither recvfrom() nor recv() can preserve datagram
boundaries. A no-RTOS nonblocking application may happen to read after each
packet and appear to work, but if more than one packet is pending, recv() can
join them too. A blocking FreeRTOS reader makes that coalescing case especially
likely while the network task continues receiving packets.

Proper receive handling needs BSD-owned per-datagram metadata recorded on each
MG_EV_READ: payload length and source address. c->recv can retain the bytes,
but reads must consume exactly one recorded datagram at a time; a short read
must discard that datagram's remaining bytes rather than expose them as another
packet.

Proper sendto() needs per-datagram destination and resolution state. It cannot
use the endpoint's single mutable c->rem. This likely means an outbound
datagram queue or a net_builtin send-to primitive which owns destination, ARP,
and L2 state. Also needed: connected-UDP peer filtering, correct oversized
datagram failure rather than silent truncation, implicit bind, and validation
of socket protocol and address family/length.

The apparent no-RTOS mitigation is not valid. A BSD caller can only run after
mg_mgr_poll() returns, but an interrupt/complex driver lets mg_tcpip_poll()
drain up to seven queued frames in one manager poll. Several UDP datagrams can
therefore be appended before the first recv(). The simple driver->rx path
currently processes one frame per poll, but UDP shim semantics cannot depend on
which driver implementation the application selected.

## failure to enqueue at the command queue

Queueing a command with portMAX_DELAY normally waits for room. A real send
failure means the command will never be executed, so bsd_dispatch() must log
it and return -1/EIO immediately rather than wait forever for its completion
notification.

## socket registry and pending accepts

The registry is a fixed array, not a linked list of allocated BSD socket
objects. A BSD descriptor maps directly to one array slot by subtracting the
descriptor base. This gives constant-time descriptor lookup for the common
send(), recv(), close(), fcntl(), poll(), and select() paths, stable addresses
for Mongoose fn_data, deterministic memory use, and a clear hard resource
limit. A linked list is reasonable for Mongoose manager traversal, where the
manager already visits every connection, but it would turn every BSD fd lookup
into a linear search.

The current array still linearly searches for a free slot. That cost occurs
only on socket creation and accept, so it is negligible for the intended small
limits. If it needs removal, keep the array and link only unused slots into an
embedded free list: allocation and release become constant-time without heap
allocation, while descriptor lookup remains direct.

The useful active list is per listener, not global. Each listener should own
its orphaned accepted children and their count. MG_EV_ACCEPT appends a child,
accept() pops the head, and readiness is a head-null check. This removes the
current whole-registry scans from listener readiness, accept(), and listener
close, and directly supplies the state needed to enforce listen(backlog).
When that linked pending-accept list is full, the shim rejects the already
established child and lets its normal Mongoose close path release it. This is
later than a native TCP SYN backlog decision, but is the correct limit
available at the BSD shim boundary.

The current branch creates a Mongoose connection at socket(), then bind()
stores the local address in that connection and marks a TCP socket as
is_listening. This is semantically wrong: bind() only reserves a local
endpoint; listen() creates the TCP listener. A SYN arriving between them is
currently accepted by net_builtin even though the BSD application has not
called listen().

Before BSD backlog enforcement, that incorrectly accepted child became an
orphan and waited for accept() or its timeout. With enforcement, its listener
backlog is still zero, so count >= backlog and the child is immediately
mg_error()ed after the TCP handshake. The symptom changes from a stray
pending connection to an immediate post-accept close; the root cause remains
that bind() exposed a listener.

The clean lifecycle is to keep a bound IPv4 address and a bound flag in the
BSD socket state. TCP listen() then creates the Mongoose listener, and TCP
connect() creates its client connection. UDP needs a Mongoose endpoint when
bound so incoming datagrams can be demultiplexed, and can create one lazily
for an unbound sendto()/connect(). An address costs more than one flag inside
the BSD socket struct, but it is configuration that must exist before a
transport does. It also makes getsockname(), setsockopt(), bind-related
options, and later address APIs independent of whether s->c currently exists.

Lazy transport creation makes a null s->c ambiguous: it can mean that a TCP
socket has not reached listen() or connect(), or that its Mongoose connection
has closed. A closed flag preserves that distinction. Before activation,
select()/poll() must not report EOF or HUP and recv()/recvfrom() return
ENOTCONN. MG_EV_CLOSE sets closed, restoring the normal detached-connection
EOF/error and HUP behaviour until the BSD caller closes its fd. It cannot be
inferred from err because an orderly peer close has no error.


cpq:
I am fairly sceptical about that BSD compat headers. That destroys our existing deployment logic. Say, customer wants to use Mongoose with BSD API. Suddenly we must provide a way other than copying mongoose.{c,h} to copy those headers. I don't think we should do all that. In the worst case, customer can create a phony empty headers, but breaking our integration is not the way to go. I believe that BSD compat should be dropped.

Having all BSD compat in a single src/bsd.h and implementation in a single src/bsd.c . We just need to make sure it works. Our selling point is to make it work on RTOS and bare metal. 

Please avoid AI written novels - a short summary paragraph is OK, but log reads are not.

